Issue #380, opened 2024-04-10
CT: complete specification of the CALL instruction


HerbertJordan:
The specification of the CALL instruction in Tosca CT is incomplete and needs to be extended.

PR #422, opened 2024-04-14
Add support for static-call semantic in CALL op specification

HerbertJordan:
This PR adds support for static contexts to the CALL expression specification.

This contributes to #380 

Todo:
- [x] check on all EVM implementations
- [x] assess impact on execution time of tests

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -1346,136 +1346,7 @@ func getAllRules() []Rule {
 
 	// --- CALL ---
 
-	// NOTE: this rule only covers the non-static Istanbul case in a coarse-grained
-	// way. Follow-work is required to cover other revisions and situations, as well
-	// as special cases currently covered in the effect function.
-	rules = append(rules, rulesFor(instruction{
-		op:        CALL,
-		name:      "_non_static_istanbul",
-		staticGas: 700,
-		pops:      7,
-		pushes:    1,
-		conditions: []Condition{
-			Eq(ReadOnly(), false),
-			IsRevision(R07_Istanbul),
-		},
-		parameters: []Parameter{
-			GasParameter{},
-			AddressParameter{},
-			ValueParameter{},
-			MemoryOffsetParameter{},
-			MemorySizeParameter{},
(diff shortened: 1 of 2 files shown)


go/ct/spc/specification.go
```diff
@@ -1893,21 +1764,206 @@ func rulesFor(i instruction) []Rule {
...
+		Parameter: parameters,
+		Effect:    performsCall,
+	})
+
+	res = append(res, Rule{
+		Name: "call_istanbul_static_with_value",
+		Condition: And(
+			isCallWithSufficientGasAndStack,
+			Eq(ReadOnly(), true),
+			Ne(Param(2), NewU256(0)),
```
facuMH, line 1853: should this parameter have a different value than 0 ?

regression_inputs/static_call_#XXX.json
```diff
@@ -0,0 +1,616 @@
+{
+  "Status": "running",
+  "Revision": "Istanbul",
+  "ReadOnly": true,
+  "Pc": 49,
+  "Gas": 1152921504606846976,
+  "GasRefund": 2475616058676252470,
+  "Code": "946ec9e280c51f5fb1c3b23c1424f8fde5e17503f23a34a33177323c130d7d1ee12a9fe0a7675716bcc3c7c963b9863f7df159ae9c6d9a0a309af431a16da1501d6c13306da6b198cce6c637b739285930ce53f0fbfb973fddf476f40dc2e95dc1eb36f18ba014",
+  "Stack": [
+    "94195109773ec5ba f3c9e5bc2e45b243 f5058fcff7a15984 6bfab812b045fd73",
```
facuMH, line 10: only the bottom 7 values are relevant for this call, the top of the stack can be removed.

HerbertJordan: I removed this file, since it is not really covering an issue in any of the VM implementations but in the CT adaptors.

facuMH (approved):
Looks good, thanks !! this pattern could be used for similar instructions with many differences per revision.

PR #427, opened 2024-04-17
CT: reduce make+copy pattern

### LINKED ISSUE #359: CT: replace make+copy for `*.clone`
in multiple places over the code there are two lines of the style
```
a := make([]byte, size)
copy(a,b)
```
with `[]byte`, or `slice` or `map`. This can always be replaced by 
` a := bytes.Clone(b)`.

facuMH:
This PR attempts to reduce the amount of times the pattern `make`+`copy` is called. replacing it by either `clone` or `PadSlice`.

This PR Fixes  #359

### DIFF
--- a/go/ct/gen/stack.go
+++ b/go/ct/gen/stack.go
@@ -22,7 +22,6 @@ import (
 
 	. "github.com/Fantom-foundation/Tosca/go/ct/common"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
-	"github.com/Fantom-foundation/Tosca/go/ct/utils"
 )
 
 type StackGenerator struct {
@@ -95,7 +94,9 @@ func (g *StackGenerator) BindValue(pos int, variable Variable) {
 }
 
 func (g *StackGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Stack, error) {
-	constraints := utils.RightPadSliceWithCap(g.constValues, int64(len(g.constValues)), int64(len(g.constValues)+len(g.varValues)))
+	// convert variable constraints to constant constraints
+	constraints := make([]constValueConstraint, len(g.constValues), len(g.constValues)+len(g.varValues))
+	copy(constraints, g.constValues)
 
 	for _, cur := range g.varValues {
 		value, found := assignment[cur.variable]
--- a/go/ct/gen/stack.go
+++ b/go/ct/gen/stack.go
@@ -22,6 +22,7 @@ import (
 
 	. "github.com/Fantom-foundation/Tosca/go/ct/common"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
+	"github.com/Fantom-foundation/Tosca/go/ct/utils"
 )
 
 type StackGenerator struct {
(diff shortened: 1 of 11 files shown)


go/ct/utils/adapter.go
```diff
@@ -190,3 +189,21 @@ func (c *ctRunContext) IsSlotInAccessList(addr vm.Address, key vm.Key) (addressP
...
+	res := make([]T, len, cap)
+	copy(res, source)
+	return res
+}
+
+func LeftPadSlice[T any](source []T, size int64) []T {
+	res := make([]T, size)
+	copy(res[size-int64(len(source)):], source)
+	return res
+}
```
HerbertJordan, line 209: The `adapter.go` file is not a good place for these functions, since nobody would look for them here.

Suggestion: place them into a file like "go/ct/common/slices.go"

HerbertJordan, line 209: These functions need unit tests.

HerbertJordan, line 209: Please use `int` for the sizes. Go build-in operations on slices use `int`.

go/ct/gen/stack.go
```diff
@@ -94,9 +95,8 @@ func (g *StackGenerator) BindValue(pos int, variable Variable) {
 }
 
 func (g *StackGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Stack, error) {
-	// convert variable constraints to constant constraints
-	constraints := make([]constValueConstraint, len(g.constValues), len(g.constValues)+len(g.varValues))
-	copy(constraints, g.constValues)
+	constraints := utils.RightPadSliceWithCap(g.constValues, int64(len(g.constValues)), int64(len(g.constValues)+len(g.varValues)))
+
```
HerbertJordan, line 99: This is a special case that semantically is not even a padding. I think this should remain as it used to be.

HerbertJordan (changes requested):

go/ct/st/code_test.go
```diff
@@ -105,22 +104,6 @@ func TestCode_GetData(t *testing.T) {
 	}
 }
 
-func TestCode_CopyTo(t *testing.T) {
```
simonlechner, line 108: Although it is very simple there should still be a unit test for the Code Copy function.

go/vm/test/integration_test.go
```diff
@@ -709,7 +710,7 @@ func TestNoReturnDataForCreate(t *testing.T) {
 						byte(geth.PUSH1), byte(0),
 						byte(test.returnInstruction)}
 
-					createInputBytes := rightPadBytes(createInputCode, 32)
+					createInputBytes := common.RightPadSlice(createInputCode, 32)
```
simonlechner, line 713: I am not sure if there should be dependencies in the VM implementations to the CT.

simonlechner (approved):
Just some small remarks.

go/vm/test/integration_test.go
HerbertJordan: Right, this is not great. However, right now we do have multiple CT dependencies from the VM to the CT directory, in particular for the CT adaptor. On the other hand, we have CT to VM dependencies for interfaces and basic types.

We will eventually have to fix a dependency order and clean this up in a follow-up.

HerbertJordan (approved):

go/ct/st/code_test.go
```diff
@@ -115,7 +115,7 @@ func TestCode_CopyTo(t *testing.T) {
 
 	for i := 0; i < len(src); i++ {
 		dst := make([]byte, i)
-		if got := code.CopyTo(dst); got != i || !bytes.Equal(src[0:i], dst) {
+		if got := code.CopyCodeSlice(0, i, dst); got != i || !bytes.Equal(src[0:i], dst) {
```
HerbertJordan, line 118: The test name should be updated accordingly.

go/ct/st/code_test.go
HerbertJordan: I do not think this issue was addressed as requested.

HerbertJordan (approved):

simonlechner (approved):

PR #428, opened 2024-04-17
CT: Add SELFDESTRUCT

### LINKED ISSUE #386: CT: add specification of the SELFDESTRUCT instruction
The specification of the SELFDESTRUCT instruction in Tosca CT is missing.

facuMH:
This PR adds the `SELFDESTRUCT` op and rules to the CT. 

This PR fixes #386

### DIFF
--- a/go/ct/rlz/conditions.go
+++ b/go/ct/rlz/conditions.go
@@ -660,23 +660,18 @@ func (c *isAddressCold) String() string {
 // Has Address Selfdestructed
 
 type hasSelfDestructed struct {
-	isSet bool
 }
 
 func HasSelfDestructed() Condition {
-	return &hasSelfDestructed{true}
+	return &hasSelfDestructed{}
 }
 
 func (c *hasSelfDestructed) Check(s *st.State) (bool, error) {
-	if c.isSet {
-		return s.HasSelfDestructed, nil
-	}
-	return true, nil
-
+	return s.HasSelfDestructed, nil
 }
 
 func (c *hasSelfDestructed) Restrict(generator *gen.StateGenerator) {
-	generator.SelfDestruct()
+	generator.MustBeSelfDestructed()
 }
 
 func (c *hasSelfDestructed) GetTestValues() []TestValue {
@@ -684,9 +679,9 @@ func (c *hasSelfDestructed) GetTestValues() []TestValue {
 	domain := boolDomain{}
 	restrict := func(generator *gen.StateGenerator, hasSelfDestructed bool) {
 		if hasSelfDestructed {
-			generator.SelfDestruct()
+			generator.MustBeSelfDestructed()
 		} else {
-			generator.NotSelfDestruct()
+			generator.MustNotBeSelfDestructed()
 		}
 	}
 	return []TestValue{
@@ -696,47 +691,31 @@ func (c *hasSelfDestructed) GetTestValues() []TestValue {
 }
 
 func (c *hasSelfDestructed) String() string {
-	return fmt.Sprintf("hasSelfDestructed(%v)", c.isSet)
+	return "hasSelfDestructed()"
 }
 
 ////////////////////////////////////////////////////////////
 // Has Not Address Selfdestructed
 
 type hasNotSelfDestructed struct {
-	isSet bool
 }
 
 func HasNotSelfDestructed() Condition {
-	return &hasNotSelfDestructed{true}
+	return &hasNotSelfDestructed{}
 }
 
 func (c *hasNotSelfDestructed) Check(s *st.State) (bool, error) {
-	if c.isSet {
-		return !s.HasSelfDestructed, nil
-	}
-	return true, nil
+	return !s.HasSelfDestructed, nil
 }
 
 func (c *hasNotSelfDestructed) Restrict(generator *gen.StateGenerator) {
-	generator.NotSelfDestruct()
+	generator.MustNotBeSelfDestructed()
 }
 
 func (c *hasNotSelfDestructed) GetTestValues() []TestValue {
-	property := Property(c.String())
-	domain := boolDomain{}
-	restrict := func(generator *gen.StateGenerator, hasSelfDestructed bool) {
-		if hasSelfDestructed {
-			generator.SelfDestruct()
-		} else {
-			generator.NotSelfDestruct()
-		}
-	}
-	return []TestValue{
-		NewTestValue(property, domain, true, restrict),
-		NewTestValue(property, domain, false, restrict),
-	}
+	return HasSelfDestructed().GetTestValues()
 }
 
 func (c *hasNotSelfDestructed) String() string {
-	return fmt.Sprintf("hasNotSelfDestructed(%v)", c.isSet)
+	return "hasNotSelfDestructed()"
 }
--- a/go/ct/rlz/conditions.go
+++ b/go/ct/rlz/conditions.go
@@ -20,6 +20,7 @@ import (
 	. "github.com/Fantom-foundation/Tosca/go/ct/common"
 	"github.com/Fantom-foundation/Tosca/go/ct/gen"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
+	"github.com/Fantom-foundation/Tosca/go/vm"
 )
 
 // Condition represents a state property.
@@ -660,10 +661,10 @@ func (c *isAddressCold) String() string {
 // Has Address Selfdestructed
 
 type hasSelfDestructed struct {
-	key BindableExpression[U256]
+	key BindableExpression[vm.Address]
 }
 
-func HasSelfDestructed(addr BindableExpression[U256]) Condition {
+func HasSelfDestructed(addr BindableExpression[vm.Address]) Condition {
 	return &hasSelfDestructed{addr}
 }
 
@@ -672,7 +673,7 @@ func (c *hasSelfDestructed) Check(s *st.State) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	_, ok := s.HasSelfDestructed[addr.Bytes20be()]
+	_, ok := s.HasSelfDestructed[addr]
 	return ok, nil
 }
 
@@ -702,10 +703,10 @@ func (c *hasSelfDestructed) String() string {
 // Has Not Address Selfdestructed
 
 type hasNotSelfDestructed struct {
-	key BindableExpression[U256]
+	key BindableExpression[vm.Address]
 }
 
-func HasNotSelfDestructed(key BindableExpression[U256]) Condition {
+func HasNotSelfDestructed(key BindableExpression[vm.Address]) Condition {
 	return &hasNotSelfDestructed{key}
 }
 
@@ -714,7 +715,7 @@ func (c *hasNotSelfDestructed) Check(s *st.State) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	_, ok := s.HasSelfDestructed[key.Bytes20be()]
+	_, ok := s.HasSelfDestructed[key]
 	return !ok, nil
 }
 
--- a/go/ct/rlz/conditions.go
+++ b/go/ct/rlz/conditions.go
@@ -669,11 +669,11 @@ func HasSelfDestructed(addr BindableExpression[vm.Address]) Condition {
 }
 
 func (c *hasSelfDestructed) Check(s *st.State) (bool, error) {
-	addr, err := c.key.Eval(s)
+	key, err := c.key.Eval(s)
 	if err != nil {
 		return false, err
 	}
-	_, ok := s.HasSelfDestructed[addr]
+	_, ok := s.HasSelfDestructed[key]
 	return ok, nil
 }
 
@@ -683,16 +683,22 @@ func (c *hasSelfDestructed) Restrict(generator *gen.StateGenerator) {
 	generator.SelfDestruct(key)
 }
 
-func (c *hasSelfDestructed) EnumerateTestCases(generator *gen.StateGenerator, consume func(*gen.StateGenerator) ConsumerResult) ConsumerResult {
-	positive := generator.Clone()
-	c.Restrict(positive)
-	if consume(positive) == ConsumeAbort {
-		return ConsumeAbort
+func (c *hasSelfDestructed) GetTestValues() []TestValue {
+	property := Property(c.String())
+	domain := boolDomain{}
+	restrict := func(generator *gen.StateGenerator, hasSelfDestructed bool) {
+		key := c.key.GetVariable()
+		c.key.BindTo(generator)
+		if hasSelfDestructed {
+			generator.HasSelfDestructedGen.BindHasSelfDestructed(key)
+		} else {
+			generator.HasSelfDestructedGen.BindHasNotSelfDestructed(key)
+		}
 	}
-
-	negative := generator.Clone()
-	HasSelfDestructed(c.key).Restrict(negative)
-	return consume(negative)
+	return []TestValue{
+		NewTestValue(property, domain, true, restrict),
+		NewTestValue(property, domain, false, restrict),
+	}
 }
 
 func (c *hasSelfDestructed) String() string {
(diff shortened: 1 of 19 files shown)


go/ct/rlz/expressions.go
```diff
@@ -358,3 +358,34 @@ func (c constant) GetVariable() gen.Variable {
...
+}
+
+func (contractAccount) Domain() Domain[vm.Address] { return addressDomain{} }
+
+func (contractAccount) Eval(s *st.State) (vm.Address, error) {
+	return s.CallContext.CallerAddress, nil
+}
+
+func (contractAccount) Restrict(kind RestrictionKind, addr vm.Address, generator *gen.StateGenerator) {
+	// generator.SetPc(uint16(pc.Uint64()))
```
simonlechner, line 378: Should this be code? Also applies to line 390.

go/ct/spc/specification.go
```diff
@@ -1839,6 +1956,75 @@ func logOp(n int) []Rule {
...
+	s.Status = st.Stopped
+
+	// PC should not increase, but rulesFor does it for all regular cases, so we counter it here.
+	s.Pc--
+
+	s.ReturnData = s.LastCallReturnData
+	s.GasRefund += refoundGas
+}
+
+func selfDestructEffectWithWarming(s *st.State, destinationColdCost, dynamicGas, refoundGas vm.Gas) {
```
simonlechner, line 1993:
```suggestion
func selfDestructEffectWarm(s *st.State, destinationColdCost, dynamicGas, refoundGas vm.Gas) {
```

simonlechner (approved):
Looks good, just some small remarks.

go/ct/spc/specification.go
```diff
@@ -1475,6 +1475,123 @@ func getAllRules() []Rule {
...
+			AnyKnownRevision(),
+			Eq(Status(), st.Running),
+			Eq(Op(Pc()), SELFDESTRUCT),
+			Ge(Gas(), 5000),
+		),
+		Parameter: []Parameter{
+			AddressParameter{},
+		},
+		Effect: FailEffect(),
+	})
```
HerbertJordan, line 1593: I would suspect that this leaves the the SELFDESTRUCT operation under-specified. For instance, a state that is read-only, running, with a SELFDESTRUCT operation and 4000 gas has no associated rule.

I am also surprised that this is not a case that is tested. The test-state generation should produce a state where all conditions as listed here are covered, except for the Gas.

go/ct/spc/specification.go
```diff
@@ -1839,6 +1956,75 @@ func logOp(n int) []Rule {
 	return rules
 }
 
+func selfDestructEffect(s *st.State, destinationColdCost, dynamicGas, refoundGas vm.Gas) {
```
HerbertJordan, line 1959: `refoundGas` => `refundGas`

go/ct/spc/specification.go
```diff
@@ -1839,6 +1956,75 @@ func logOp(n int) []Rule {
...
+	// transaction. The transfer of the current balance to the given account cannot fail. In particular,
+	// the destination account code (if any) is not executed, or, if the account does not exist, the
+	// balance is still added to the given address.
+
+	// account to send the current balance to
+	destinationAccount := s.Stack.Pop().Bytes20be()
+	currentAccount := s.CallContext.AccountAddress
+	CurrentBalance := s.Accounts.GetBalance(currentAccount)
+
+	// If a positive balance is sent to an empty account, the dynamic gas is 25000. Otherwise there is
```
HerbertJordan, line 1970: The documentation does not match the implementation. The 25000 are only true if the `dynamicGas` is set accordingly.

go/ct/spc/specification.go
HerbertJordan, line 1959: `dynamicGas` => `accountCreationFee`

go/ct/spc/specification.go
```diff
@@ -1839,6 +1956,75 @@ func logOp(n int) []Rule {
...
+	}
+
+	dynamicCost += destinationColdCost
+
+	if s.Gas < dynamicCost {
+		s.Status = st.Failed
+		return
+	}
+	s.Gas -= dynamicCost
+	s.Accounts.MarkWarm(destinationAccount)
```
HerbertJordan, line 2018: This function is almost identical to the function above, except that the account is warmed up here.

I would merge them, and check in the function whether the revision is high enough to cause the warm-up.

go/ct/rlz/expressions.go
HerbertJordan: Right, this seems incomplete.

HerbertJordan (changes requested):
I found a few issues that should be fixed before another round of review.

HerbertJordan (commented):
Waiting for changes discussed offline. Please re-request a review when done.

go/vm/geth/geth.go
```diff
@@ -166,12 +166,31 @@ func canTransferFunc(stateDB geth.StateDB, callerAddress common.Address, value *
 	return stateDB.GetBalance(callerAddress).Cmp(value) >= 0
 }
 
+type selfDestroyEntry struct {
```
simonlechner, line 169: The names should be consistent with the other structs and functions. This could also be simplified to just the beneficiary as we discussed offline.

go/ct/rlz/conditions.go
```diff
@@ -655,3 +655,88 @@ func (c *isAddressCold) GetTestValues() []TestValue {
...
+		if hasSelfDestructed {
+			generator.SelfDestruct()
+		} else {
+			generator.NotSelfDestruct()
+		}
+	}
+	return []TestValue{
+		NewTestValue(property, domain, true, restrict),
+		NewTestValue(property, domain, false, restrict),
+	}
```
HerbertJordan, line 737: Since it is the negation of `hasSelfDestructed` you should just call
```suggestion
	return HasSelfDestructed().GetTestValues()
```
This also solves the problem that those two conditions should target the same property.

go/ct/rlz/conditions.go
```diff
@@ -655,3 +655,88 @@ func (c *isAddressCold) GetTestValues() []TestValue {
...
+	return true, nil
+}
+
+func (c *hasNotSelfDestructed) Restrict(generator *gen.StateGenerator) {
+	generator.NotSelfDestruct()
+}
+
+func (c *hasNotSelfDestructed) GetTestValues() []TestValue {
+	property := Property(c.String())
```
HerbertJordan, line 725: This constraint should target the same property than the `HasSelfDestructed` constraint.

go/ct/rlz/conditions.go
```diff
@@ -655,3 +655,88 @@ func (c *isAddressCold) GetTestValues() []TestValue {
...
+
+func (c *hasSelfDestructed) String() string {
+	return fmt.Sprintf("hasSelfDestructed(%v)", c.isSet)
+}
+
+////////////////////////////////////////////////////////////
+// Has Not Address Selfdestructed
+
+type hasNotSelfDestructed struct {
+	isSet bool
```
HerbertJordan, line 706: This field can be removed, since it is always `true`.

go/ct/rlz/conditions_test.go
```diff
@@ -242,3 +243,48 @@ func TestCondition_String(t *testing.T) {
...
+	state := st.NewState(st.NewCode([]byte{}))
+	state.CallContext.AccountAddress = NewAddress(NewU256(0x01))
+	state.HasSelfDestructed = true
+
+	hasSelfDestructed, err := HasSelfDestructed().Check(state)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !hasSelfDestructed {
+		t.Fatal("account not set as selfdestructed, when it should be")
```
HerbertJordan, line 257: This error message does not match what was actually tested.

go/ct/rlz/conditions_test.go
```diff
@@ -242,3 +243,48 @@ func TestCondition_String(t *testing.T) {
...
+	}
+
+	state.HasSelfDestructed = false
+
+	hasNotSelfDestructed, err := HasNotSelfDestructed().Check(state)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !hasNotSelfDestructed {
+		t.Fatal("account set as selfdestructed, when it should not be")
```
HerbertJordan, line 267: Same here.

go/ct/spc/specification.go
```diff
@@ -1710,6 +1731,76 @@ func logOp(n int) []Rule {
 	return rules
 }
 
+func nonStaticSelfDestOp(revision Revision, warm bool, destinationColdCost, accountCreationFee, refundGas vm.Gas) []Rule {
```
HerbertJordan, line 1734: I do not like the name. It is not an `Op` (assuming Operation) and it also does not produce one. Please improve.

Also: no abbreviation

go/ct/spc/specification.go
```diff
@@ -1710,6 +1731,76 @@ func logOp(n int) []Rule {
...
+		warmColdString = "warm"
+		targetWarm = IsAddressWarm(Param(0))
+	} else {
+		warmColdString = "cold"
+		targetWarm = IsAddressCold(Param(0))
+	}
+
+	name := fmt.Sprintf("_%v_%v", strings.ToLower(revision.String()), warmColdString)
+
+	inst := instruction{
```
HerbertJordan, line 1748: nit: no abriviations; `inst` => `instruction`

go/ct/spc/specification.go
```diff
@@ -1347,6 +1347,27 @@ func getAllRules() []Rule {
...
+
+	rules = append(rules, rulesFor(instruction{
+		op:        SELFDESTRUCT,
+		name:      "_staticcall",
+		staticGas: 5000,
+		conditions: []Condition{
+			Eq(ReadOnly(), true),
+			AnyKnownRevision(),
+		},
+		effect: FailEffect().Apply,
```
HerbertJordan, line 1368:
```suggestion
		effect: FailEffect(),
```

go/ct/spc/specification.go
```diff
@@ -1710,6 +1731,76 @@ func logOp(n int) []Rule {
...
+	if s.Revision > R07_Istanbul {
+		s.Accounts.MarkWarm(destinationAccount)
+	}
+	// add beneficiary to list in state
+	s.HasSelfDestructed = true
+	s.SelfDestructedJournal = append(s.SelfDestructedJournal, st.NewSelfDestructEntry(s.CallContext.AccountAddress, CurrentBalance.Bytes32be()))
+	s.Status = st.Stopped
+
+	// PC should not increase, but rulesFor does it for all regular cases, so we counter it here.
+	s.Pc--
```
HerbertJordan, line 1800: nit: not really a good reason. I would remove it and only list the bare minimum of effects.

go/ct/st/state.go
```diff
@@ -46,6 +48,49 @@ const (
...
+func NewSelfDestructEntry(acc vm.Address, bal vm.Value) SelfDestructEntry {
+	return SelfDestructEntry{account: acc, oldBalance: bal}
+}
+
+func (s SelfDestructEntry) String() string {
+	return fmt.Sprintf("account: %v, oldBalance: %v", s.account, s.oldBalance)
+}
+
+func (s SelfDestructEntry) MarshalJSON() ([]byte, error) {
+	return json.Marshal(s.String())
```
HerbertJordan, line 65: nit: it kinda breaks the JSON encoding if you encode two nested elements in a string that needs to be parsed again. It might be cleaner to create a
```
type serializableEntry struct {
   Account vm.Address
   Beneficiary vm.Address
}
```
copy all data into that, and use the JSON serialization on that object. Wdyt?

HerbertJordan (changes requested):
There seem to be a number of additional simplifications and cleanups that should be applied. Please see the comments.

go/ct/spc/specification.go
facuMH: `FailEffect()` returns an `Effect`  which is not a `func(*state)`, hence I need to call the `.Apply` method, which does meet the required interface.

go/ct/rlz/conditions_test.go
```diff
@@ -242,3 +243,48 @@ func TestCondition_String(t *testing.T) {
...
+		t.Errorf("%v", err)
+	}
+
+	got, err := condition.Check(state)
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+
+	if !got {
+		t.Error("generated state does not satisfay condition")
```
HerbertJordan, line 288: nit: `satisfay` => `satisfy`

go/ct/spc/specification.go
```diff
@@ -1710,6 +1731,73 @@ func logOp(n int) []Rule {
...
+
+	instruction := instruction{
+		op:        SELFDESTRUCT,
+		name:      name,
+		staticGas: 5000,
+		pops:      1,
+		conditions: []Condition{
+			Eq(ReadOnly(), false),
+			IsRevision(revision),
+			HasNotSelfDestructed(),
```
HerbertJordan, line 1756: How about cases where the account has self-destructed before? Is there a specification rule for that?

go/ct/st/state.go
```diff
@@ -400,5 +421,26 @@ func (s *State) Diff(o *State) []string {
...
+				len(s.SelfDestructedJournal), len(o.SelfDestructedJournal)))
+		} else {
+			for index, entry1 := range s.SelfDestructedJournal {
+				account1, beneficiary1 := entry1.account, entry1.beneficiary
+				entry2 := o.SelfDestructedJournal[index]
+				account2, beneficiary2 := entry2.account, entry2.beneficiary
+				if !reflect.DeepEqual(account1, account2) || !reflect.DeepEqual(beneficiary1, beneficiary2) {
+					res = append(res, fmt.Sprintf("Different has-self-destructed journal entry:\n\t(%v, %v)\n\tvs\n\t(%v, %v)",
+						account1, beneficiary1, account2, beneficiary2))
+				}
```
HerbertJordan, line 440: The accounts can be compared with ==, and as a consequence also the entries. So this can be reduced to
```suggestion
				entry2 := o.SelfDestructedJournal[index]
				if entry1 != entry2 {
					res = append(res, fmt.Sprintf("Different has-self-destructed journal entry:\n\t(%v, %v)\n\tvs\n\t(%v, %v)",
						entry1.account, entry1.beneficiary, entry2.account, entry2.beneficiary))
				}
```

go/ct/utils/adapter.go
```diff
@@ -145,8 +144,13 @@ func (c *ctRunContext) Call(kind vm.CallKind, parameter vm.CallParameter) (vm.Ca
...
 }
 
-func (c *ctRunContext) SelfDestruct(addr vm.Address, beneficiary vm.Address) bool {
-	panic("not implemented")
+func (c *ctRunContext) SelfDestruct(address vm.Address, beneficiary vm.Address) bool {
+	if c.state.HasSelfDestructed {
+		return false
+	}
+	c.state.HasSelfDestructed = true
+	c.state.SelfDestructedJournal = append(c.state.SelfDestructedJournal, st.NewSelfDestructEntry(address, beneficiary))
```
HerbertJordan, line 152: I would move this line to the top of this function, since we want to see all self-destruct calls, even if the account as already self-destructed.

HerbertJordan (changes requested):
I think the structure is right now, but there are a few more things that should be looked at. In particular, rules for already self-destructed accounts. Please see the comments.

go/ct/spc/specification.go
```diff
@@ -1347,6 +1347,34 @@ func getAllRules() []Rule {
...
+	rules = append(rules, nonStaticSelfDestructRules(R07_Istanbul, true, vm.Gas(0), vm.Gas(0), true)...)
+	rules = append(rules, nonStaticSelfDestructRules(R07_Istanbul, true, vm.Gas(0), vm.Gas(0), false)...)
+	rules = append(rules, nonStaticSelfDestructRules(R09_Berlin, false, vm.Gas(2600), vm.Gas(25000), true)...)
+	rules = append(rules, nonStaticSelfDestructRules(R09_Berlin, false, vm.Gas(2600), vm.Gas(25000), false)...)
+	rules = append(rules, nonStaticSelfDestructRules(R09_Berlin, true, vm.Gas(0), vm.Gas(0), true)...)
+	rules = append(rules, nonStaticSelfDestructRules(R09_Berlin, true, vm.Gas(0), vm.Gas(0), false)...)
+	rules = append(rules, nonStaticSelfDestructRules(R10_London, false, vm.Gas(2600), vm.Gas(25000), true)...)
+	rules = append(rules, nonStaticSelfDestructRules(R10_London, false, vm.Gas(2600), vm.Gas(25000), false)...)
+	rules = append(rules, nonStaticSelfDestructRules(R10_London, true, vm.Gas(0), vm.Gas(0), true)...)
+	rules = append(rules, nonStaticSelfDestructRules(R10_London, true, vm.Gas(0), vm.Gas(0), false)...)
```
HerbertJordan, line 1364: With additional revisions this will become longer and longer ... at some point we should produce those cases using nested loops. But I guess we can leave this until we add the Paris revision.

HerbertJordan (dismissed):
Thanks for completing this part of the specification!

simonlechner (approved):
LGTM

PR #429, opened 2024-04-18
State enumeration cleanup

### LINKED ISSUE #370: CT: cleanup of parallel test state enumeration code
the new refactor of parallel enumeration should be removed from the driver and moved into a separate utility

simonlechner:
This PR moves the `forEachState` function to its own file and is now referred to as coordination. Unit tests for this function are also added.
To clean up the Tosca repository further, the regression inputs are moved to the ct directory.
Fixes #370.


go/ct/spc/coordination.go
HerbertJordan, line 1: nit: "enumeration.go"?

go/ct/spc/coordination_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/Fantom-foundation/Tosca/go/ct/gen"
+	"github.com/Fantom-foundation/Tosca/go/ct/rlz"
+	"github.com/Fantom-foundation/Tosca/go/ct/st"
+)
+
+// Test Condition
```
HerbertJordan, line 17: Please improve the documentation.

go/ct/spc/coordination_test.go
```diff
@@ -0,0 +1,186 @@
...
+	printFunction := func(time time.Duration, rate float64, current int64) {}
+
+	statesPerRule := 15
+
+	//var condition testCondition
+	rules := []rlz.Rule{}
+	testRule := rlz.Rule{
+		Name:      "test_rule",
+		Condition: condition(true, statesPerRule),
+		Effect:    rlz.NoEffect(),
```
HerbertJordan, line 67: nit: the `Name` and `Effect` could likely be omitted for this test since they are not relevant, right?

go/ct/spc/coordination_test.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	rules = append(rules, testRule)
```
HerbertJordan, line 69:
```suggestion
	rules := []rlz.Rule{{Condition: condition(true, statesPerRule)}}
```

go/ct/spc/coordination_test.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		t.Errorf("Unexpected error during state generation %v", err)
+	}
+
+	if counter.Load() != int64(statesPerRule) {
+		t.Errorf("unexpected number of evaluated states: %d", counter.Load())
+	}
+}
+
+func TestCoordination_AllRulesAreEnumerated(t *testing.T) {
```
HerbertJordan, line 80: I think this test name should be renamed to match what the test actually checks. Something like
`DisabledFullModeFiltersNonMatchingRules` ?

go/ct/spc/coordination_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+func TestCoordination_FilterRules(t *testing.T) {
+	filters := []string{"add", "sub", "mul", "copy", "call"}
+	for _, subString := range filters {
+		filter, err := regexp.Compile(subString)
+		if err != nil {
+			t.Error("regular expression not compilable")
+		}
+		rules := Spec.GetRules()
+		rules = FilterRules(rules, filter)
```
HerbertJordan, line 137: This test does not check that all rules that should be in the result are actually in the result. It only checks that those that are listed match the criteria.

go/ct/spc/coordination_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+func TestCoordination_RightNumberOfGoroutinesIsStarted(t *testing.T) {
+	numJobs := 4
+	seed := 0
+	fullMode := false
+	filter := regexp.MustCompile(".*")
+
+	opFunction := func(state *st.State) rlz.ConsumerResult {
+
+		// 3 goroutines (sweeper, scavenger and finalizer) are started by default
```
HerbertJordan, line 175: Maybe instead of hard-coding the number of goroutines running outside of the code to be tested, query them before the `ForEachState` run and store it in a variable. This would make the test agnostic to the use of goroutines by the test framework or the runtime system.

go/ct/spc/coordination_test.go
HerbertJordan, line 1: Since we had problems with this before, I think tests covering aborted enumerations should be added (see #368 and #371).

HerbertJordan (changes requested):
I added some minor remarks and a request for one or two additional unit tests. Please have a look.

The relocation of the regression tests is unrelated with the main change of this PR and should be covered in a distinct PR.

go/ct/spc/coordination_test.go
simonlechner: Thank you for the suggestion!

HerbertJordan (approved):

facuMH (approved):

PR #432, opened 2024-04-22
Make properties in test case enumeration explicit

HerbertJordan:
This improvement to the test case generation process fixes support for constraints where there are multiple constraints for the same property of the state. 

Before this change, when enumerating test cases for a constraint of the form

`stackSize ≥ 3 ∧ stackSize ≤ 1024`

both clauses would be independently asked for example-stack-sizes to be used in test states. Those may have been, for instance, the values `{0, 2, 3, 4, 1024}` for the first clause and `{0, 1023, 1024}` for the second clause. From those sets, due to the implicit combination of constraints yield by both constraints, the intersection was formed and tested. In the given example, this would result in the set `{0, 1024}` as values for actual test cases, ignoring the values being closer to the boundary limits.

With this change, domains are explicitly extracted from constraints and merged instead of intersected for creating the domains for the test case enumeration.

As a side-effect, constraints like `revision(Istanbul) ∧ revision(Istanbul-London)` which so far caused the generation of the 4 sample values `{Istanbul, Berlin, London, UnknownFutureRevsion}` each, leading to a total of 16 test cases, are now only producing 4 test cases. This significantly reduces the number of test case to be executed in a full CT evaluation.

Finally, the `generation-info` tool in the driver was updated to provide additional insights on the domains. The summary for a rule now looks like this:
```
----- Rule: addmod_regular -----
Conditions:
	Gas ≥ 8
	code[PC] = ADDMOD
	revision(Istanbul-London)
	stackSize ≤ 1024
	stackSize ≥ 3
	status = running
Domains:
	code[PC]: N=2, {ADDMOD, MULMOD}
	gas: N=5, {0, 7, 8, 9, 1152921504606846976}
	revision: N=4, {Istanbul, Berlin, London, UnknownNextRevision}
	stackSize: N=6, {0, 2, 3, 4, 1023, 1024}
	status: N=4, {running, stopped, reverted, failed}
Parameters:
	0: 11
	1: 11
	2: 11
Total number of cases: 1277760
Share of total rules: 1.6%
```

### DIFF
--- a/go/ct/driver/run.go
+++ b/go/ct/driver/run.go
@@ -158,7 +158,7 @@ func doRun(context *cli.Context) error {
 
 	// Summarize the result.
 	if skippedCount.Load() > 0 {
-		fmt.Printf("Number of skipped tests: %d", skippedCount.Load())
+		fmt.Printf("Number of skipped tests: %d\n", skippedCount.Load())
 	}
 
 	if len(issues) == 0 {
--- a/go/ct/rlz/conditions.go
+++ b/go/ct/rlz/conditions.go
@@ -31,9 +31,9 @@ type Condition interface {
 	// Condition holds.
 	Restrict(*gen.StateGenerator)
 
-	// EnumerateTestCases sets constraints on a copy of the given generator and
-	// invokes the given consumer function with it.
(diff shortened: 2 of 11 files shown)


simonlechner (approved):
Nice change from the recursive enumeration!

facuMH (commented):
LG! this will be very helpful

PR #434, opened 2024-04-23
Improve Stack Pool

simonlechner:
In the profile of the CT run, stack memory allocation still had a big impact on the runtime. Stacks were garbage collected and therefore had to be newly allocated. 

This PR fixes this and returns all stacks, memory allocations are no longer visible in the profile.

### DIFF
--- a/go/vm/lfvm/ct_test.go
+++ b/go/vm/lfvm/ct_test.go
@@ -356,7 +356,8 @@ func TestConvertToCt_Stack(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			for _, cur := range test {
 				want := cur.ctStack
-				got := convertLfvmStackToCtStack(cur.lfvmStack)
+				ctStack := st.NewStack()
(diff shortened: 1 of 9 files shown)


go/vm/lfvm/ct_test.go
```diff
@@ -374,6 +375,7 @@ func BenchmarkLfvmStackToCtStack(b *testing.B) {
 		stack.pushEmpty().SetUint64(uint64(i))
 	}
 	for i := 0; i < b.N; i++ {
-		convertLfvmStackToCtStack(stack)
+		ctStack := st.NewStack()
```
HerbertJordan, line 378: nit: maybe for the benchmark reusing the same stack every time is the better choice, since it is closer to the actual use case scenario.

HerbertJordan (dismissed):
Looks good, thanks for fixing!

facuMH (approved):
looks good.

HerbertJordan (approved):

PR #438, opened 2024-04-24
Unsupported Revision Error

### LINKED ISSUE #320: VM: interpreters should return a dedicated error in case of an unsupported revision
The redesign of Tosca's top-level interfaces by #311 introduced the possibility to produce errors specifically indicating the failure of an interpreter to execute some byte code -- in contrast to issues within the byte code causing the execution to produce an unsuccessful 
result. This capability should be used to specify how unsupported revisions should be handled.

Implementations of the [`Run`](https://github.com/Fantom-foundation/Tosca/blob/2b2098c7bbe0c7e1daa7576e3b00cab9fcf6b9a2/go/vm/interpreter.go#L18) method of the Interpreter interface should be required to return a dedicated `ErrUnsupportedRevision` error in cases where the requested revision is not supported by the targeted interpreter.

Task to be covered:
- [ ] the definition of an `ErrUnsupportedRevision` constant in `vm/interpreter.go`
- [ ] the update of the documentation of the `Run` function
- [ ] the addition of a unit test in `vm/test` that performs calls with an invalid revision testing that the requested error is returned
- [ ] the fixing of all interpreter implementations to pass the test
- [ ] the acceptance of this kind of error by the CT driver; the number of such cases should be reported at the end of the run

simonlechner:
This PR introduces a custom error type for unsupported revisions.
Fixes #320.
Further a newest supported revision for the CT specification is introduced. This ensures that rules that do not change with new revisions still produce test cases with new them.

### DIFF
--- a/go/ct/driver/run.go
+++ b/go/ct/driver/run.go
@@ -143,9 +143,14 @@ func doRun(context *cli.Context) error {
 			return rlz.ConsumeContinue
 		}
 
-		if err := runTest(state, evm, filter, &numUnsupportedTests); err != nil {
-			issuesCollector.AddIssue(state, err)
-			fmt.Printf("Error: %v\n", err)
+		if err := runTest(state, evm, filter); err != nil {
+			targetError := &vm.ErrUnsupportedRevision{}
+			if errors.As(err, &targetError) {
+				numUnsupportedTests.Add(1)
+				return rlz.ConsumeContinue
+			}
+
+			issuesCollector.AddIssue(state, fmt.Errorf("failed to process input state %v: %w", state, err))
 		}
 
 		return rlz.ConsumeContinue
@@ -202,7 +207,7 @@ func doRun(context *cli.Context) error {
 
 // runTest runs a single test specified by the input state on the given EVM. The
 // function returns an error in case the execution did not work as expected.
-func runTest(input *st.State, evm ct.Evm, filter *regexp.Regexp, numUnsupportedTests *atomic.Int32) error {
+func runTest(input *st.State, evm ct.Evm, filter *regexp.Regexp) error {
 	rules := spc.Spec.GetRulesFor(input)
 	if len(rules) == 0 {
 		return nil // < TODO: make this an error once the specification is complete
@@ -224,15 +229,8 @@ func runTest(input *st.State, evm ct.Evm, filter *regexp.Regexp, numUnsupportedT
 
 	result, err := evm.StepN(input.Clone(), 1)
 	defer result.Release()
-
-	targetError := &vm.ErrUnsupportedRevision{}
-	if errors.As(err, &targetError) {
-		numUnsupportedTests.Add(1)
-		return nil
-	}
-
 	if err != nil {
-		return fmt.Errorf("failed to process input state %v: %w", input, err)
+		return err
 	}
 
 	if result.Eq(expected) {
--- a/go/ct/driver/run.go
+++ b/go/ct/driver/run.go
@@ -13,6 +13,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"os"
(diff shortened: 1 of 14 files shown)


go/vm/interpreter.go
```diff
@@ -25,8 +27,10 @@ type Interpreter interface {
 	// code was correctly executed (even if the execution was aborted due do to
 	// a code-internal issue). The error is not nil if some problem within the
 	// interpreter caused the execution to fail to correctly process the provided
-	// program. In such a case the result is undefined. Interpreters are required
-	// to be thread-safe. Thus, multiple runs may be conducted in parallel.
+	// program. In such a case the result is undefined. During a call with an
+	// unsupported Revision a special unsupported Revision Error is returned.
```
HerbertJordan, line 31:
```suggestion
	// unsupported Revision an ErrUnsupportedRevision Error is returned.
```

HerbertJordan (commented):
Looks good 👍

go/vm/evmc/evmc_interpreter.go
```diff
@@ -55,6 +58,10 @@ func (e *EvmcInterpreter) Run(params vm.Parameters) (vm.Result, error) {
 		context: params.Context,
 	}
 
+	if params.Revision > NewestSupportedRevision {
+		return vm.Result{}, &vm.ErrUnsupportedRevision{}
```
HerbertJordan, line 62: You set up the `ErrUnsupportedRevision` error as a type to allow you to pass along the unsupported revision.

If this feature of the `ErrUnsupportedRevision` error is not used, it might be simpler to define a error constant like [here](https://github.com/Fantom-foundation/Tosca/blob/8b6197ed9a575a1944dcbe524c74a9b42cb59e34/go/ct/gen/error.go#L19). This makes the returning of the error simpler (not & and {}) and checking for it is easier using `errors.Is`.

go/vm/evmzero/ct.go
```diff
@@ -39,12 +40,9 @@ func NewConformanceTestingTarget() ct.Evm {
 type ctAdapter struct{}
 
 func (a ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
-	// Hack: Special handling for unknown revision, because evmzero cannot represent an invalid revision.
-	// So we mark the status as failed already.
-	// TODO: Fix this once we add full revision support to the CT and evmzero.
-	if state.Revision > common.R10_London {
+	if state.Revision > common.Revision(evmc.NewestSupportedRevision) {
```
HerbertJordan, line 43: This line implicitly assumes that the revisions defined in the `vm` package encoded the same way as the revisions in the CT's `common` package.

You could first run the `utils.ToVmParameters(state)` function below and then ask for the revision in the parameters. This revision is of the correct type, and the util function is handling the conversion for you.

The other CT adaptors have the same issue.

go/vm/lfvm/lfvm.go
```diff
@@ -36,12 +36,19 @@ type VM struct {
 	logging                 bool
 }
 
+// Defines the newest supported revision for this interpreter implementation
+const NewestSupportedRevision = vm.R10_London
```
HerbertJordan, line 40: nit: could be package private.

go/ct/driver/run.go
```diff
@@ -216,6 +224,13 @@ func runTest(input *st.State, evm ct.Evm, filter *regexp.Regexp) error {
 
 	result, err := evm.StepN(input.Clone(), 1)
 	defer result.Release()
+
+	targetError := &vm.ErrUnsupportedRevision{}
+	if errors.As(err, &targetError) {
+		numUnsupportedTests.Add(1)
+		return nil
+	}
```
HerbertJordan, line 232: nit: maybe, instead of adding an out-parameter (`numUnsupportedTests`) to this function, all errors should just be returned the caller of this function should check for the the unsupported revision error.

go/ct/driver/run.go
```diff
@@ -140,8 +143,9 @@ func doRun(context *cli.Context) error {
 			return rlz.ConsumeContinue
 		}
 
-		if err := runTest(state, evm, filter); err != nil {
+		if err := runTest(state, evm, filter, &numUnsupportedTests); err != nil {
 			issuesCollector.AddIssue(state, err)
+			fmt.Printf("Error: %v\n", err)
```
HerbertJordan, line 148: nit: forgotten debug code?

HerbertJordan (approved):

facuMH (approved):
Looks good! I did not expect to find the introduction of the `evmInstances` types, but it seems good.

facuMH (approved):

PR #440, opened 2024-04-25
Unifies handling of range constraints

HerbertJordan:
This PR introduces a new `RangeSolver` utility capable of handling range constraints of the shape
```
constraint ::= true | constraint ∧ clause
clause     ::= X op C
op         ::= < | ≤ | = | ≥ | >
```
where `X` is a numeric property to be solved for and `C` are constants.

This common range solver is integrated into the `StateGenerator` to cover `Gas`, `GasRefund`, `Revision`, and `StackSize` constraints. By doing so, also the terminology used for defining constraints on those parameters was aligned.

### DIFF
--- a/go/ct/common/revisions.go
+++ b/go/ct/common/revisions.go
@@ -27,6 +27,9 @@ const (
 	R99_UnknownNextRevision
 )
 
+const MinRevision = R07_Istanbul
+const MaxRevision = R99_UnknownNextRevision
+
 func (r Revision) String() string {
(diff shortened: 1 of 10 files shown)


facuMH (dismissed):
Looks good !!! nice unification

simonlechner (approved):
Nice to have it in the same place now for all effected fields.

simonlechner (approved):

facuMH (approved):

PR #441, opened 2024-04-30
Add Create instruction to specification

### LINKED ISSUE #384: CT: add specification of the CREATE instruction
The specification of the CREATE instruction in Tosca CT is missing.

simonlechner:
This PR adds the CREATE opcode to the CT specification.
Fixes #384


go/ct/st/call_journal.go
```diff
@@ -25,8 +25,9 @@ import (
 // of calls as well as the effect of future calls to be triggered
 // by CREATE and CALL expressions.
 type CallJournal struct {
-	Past   []PastCall
-	Future []FutureCall
+	Past        []PastCall
+	Future      []FutureCall
+	ToBeCreated vm.Address
```
HerbertJordan, line 30: The "ToBeCreated" address should be a field of the "FutureCall" struct, ideally renamed to some past-time form, like `CreatedAccount`.

go/vm/geth/ct.go
```diff
@@ -210,7 +210,30 @@ func (i *callInterceptor) StaticCall(env *geth_vm.EVM, me geth_vm.ContractRef, a
 }
 
 func (i *callInterceptor) Create(env *geth_vm.EVM, me geth_vm.ContractRef, code []byte, gas uint64, value *big.Int) ([]byte, geth_common.Address, uint64, error) {
-	panic("not implemented")
+	addressToBeCreated := geth_common.HexToAddress("0") // TODO make address random
```
HerbertJordan, line 213: This value should be taken from the call journal result (line 221 below).

go/ct/st/call_journal.go
```diff
@@ -179,5 +181,8 @@ func (c *FutureCall) Diff(other *FutureCall) []string {
 	if have, got := c.GasRefund, other.GasRefund; have != got {
 		res = append(res, fmt.Sprintf("different refund: %v vs %v", have, got))
 	}
+	if have, got := c.CreatedAccount, other.CreatedAccount; have != got {
+		res = append(res, fmt.Sprintf("different refund: %v vs %v", have, got))
```
HerbertJordan, line 185:
```suggestion
		res = append(res, fmt.Sprintf("different created account: %v vs %v", have, got))
```

go/ct/utils/adapter.go
```diff
@@ -136,10 +136,11 @@ func (c *ctRunContext) EmitLog(addr vm.Address, topics []vm.Hash, data []byte) {
...
-		Success:   res.Success,
-		Output:    res.Output,
-		GasLeft:   res.GasLeft,
-		GasRefund: res.GasRefund,
+		Success:        res.Success,
+		Output:         res.Output,
+		GasLeft:        res.GasLeft,
+		GasRefund:      res.GasRefund,
+		CreatedAddress: res.CreatedAddress,
 	}, nil
```
HerbertJordan, line 144: Can this be reduced to this?
```
return c.state.CallJournal.Call(kind, parameter), nil
```

go/ct/spc/specification.go
```diff
@@ -1378,6 +1377,66 @@ func getAllRules() []Rule {
...
 	})...)
 
+	// --- CREATE ---
+	rules = append(rules, rulesFor(instruction{
+		op:        CREATE,
+		staticGas: 32000,
+		pops:      3,
+		pushes:    1,
+		conditions: []Condition{
+			Eq(ReadOnly(), false),
```
HerbertJordan, line 1387: The completeness check should be able to detect that there is no rule for a `CREATE` call in a Read-Only context. We can leave this as a follow-up if you like, to check the ability of the completeness check to find these problems.

HerbertJordan (dismissed):
Looks good, thanks!

go/ct/spc/specification.go
simonlechner: I added rules for Read-Only mode similar to the ones for SELFDESTRUCT, we can always comment out rules to evaluate the completeness check ability.

HerbertJordan (approved):

PR #445, opened 2024-05-02
Restrict domain for value parameters in Call rules

HerbertJordan:
This PR reduces the number of test samples for value parameters of Call expressions from 11 to 3.

Number of test cases before: 2141852308
Number of test cases after: 748393108 (-65%)


facuMH (approved):
Looks good, thanks for the change !!

PR #446, opened 2024-05-02
Check balance in VM create implementations

simonlechner:
LFVM and evmzero did not handle the CREATE instruction correctly if a value higher than the accounts balance has been sent. This is fixed with this PR.
Regression tests are going to be added with the addition of the CREATE instruction to the CT in #441.

### DIFF
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -1461,11 +1461,12 @@ struct CreateImpl {
 
     ctx.return_data.clear();
 
-    if (endowment != 0 && ToUint256(ctx.host->get_balance(ctx.message->recipient)) < endowment) {
-      return {.state = RunState::kErrorCreate};
-    }
-
     auto init_code = ctx.memory.GetSpan(init_code_offset, init_code_size);
+
+    if (endowment != 0 && ToUint256(ctx.host->get_balance(ctx.message->recipient)) < endowment) {
+      top[0] = 0;
+      return {.dynamic_gas_costs = initial_gas - gas};
+    }
 
     evmc_message msg{
         .kind = (Op == op::CREATE) ? EVMC_CREATE : EVMC_CREATE2,
--- a/cpp/vm/evmzero/interpreter_test.cc
+++ b/cpp/vm/evmzero/interpreter_test.cc
@@ -4854,6 +4854,51 @@ TEST(InterpreterTest, SELFDESTRUCT_StaticCallViolation) {
 }
 
 ///////////////////////////////////////////////////////////
+// CREATE
+
+TEST(InterpreterTest, CREATE_InsuffiecientBalanceIsHandledCorrectly) {
+  MockHost host;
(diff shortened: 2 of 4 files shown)


facuMH (approved):
LGTM !! nice fix

cpp/vm/evmzero/interpreter.cc
HerbertJordan, line 1: The changes in this file fixes two different issues:
- the incorrect handling of insufficient balance; it used to be a rollback, but it should be a skipped call
- the incorrect order of the balance check and the memory expansion

For both of those, distinct regression tests should be added to #441

go/vm/lfvm/instructions.go
```diff
@@ -753,6 +754,17 @@ func opCreate(c *context) {
 		return
 	}
 
+	if !value.IsZero() {
+		balance := [32]byte(c.context.GetBalance(c.params.Recipient))
```
HerbertJordan, line 758: nit: this conversion is not needed
```suggestion
		balance := c.context.GetBalance(c.params.Recipient)
```

go/vm/lfvm/instructions.go
HerbertJordan, line 1: With this, #441 should add 3 regression tests.

cpp/vm/evmzero/interpreter.cc
HerbertJordan, line 1: Maybe those changes should also be covered in evmzero unit tests. Wdyt?
For an example on how to do that, see #354.

go/vm/lfvm/instructions.go
HerbertJordan, line 1: Maybe those changes should also be covered in LFVM unit tests. Wdyt?
For an example on how to do that, see #352.

HerbertJordan (approved):

go/vm/lfvm/instructions.go
simonlechner: The value check error in lfvm and evmzero is triggered with the same input state, therefor it is sufficient to only add two regression tests.

cpp/vm/evmzero/interpreter.cc
simonlechner: Done.

go/vm/lfvm/instructions.go
simonlechner: Done.

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -4853,6 +4853,29 @@ TEST(InterpreterTest, SELFDESTRUCT_StaticCallViolation) {
   });
 }
 
+///////////////////////////////////////////////////////////
+// CREATE
+
+TEST(InterpreterTest, CREATE_InsuffiecientBalanceIsHandledCorrectly) {
```
HerbertJordan, line 4859: There should be another test case testing that in a case where there is not enough balance the memory has grown to the size of the input data. Before this change, this used to be not the case.

HerbertJordan (approved):

HerbertJordan (approved):

cpp/vm/evmzero/interpreter_test.cc
simonlechner: Done.

PR #450, opened 2024-05-03
CR: delegate call

### LINKED ISSUE #382: CT: add specification of the DELEGATECALL instruction

facuMH:
This PR adds the DELEGATECALL op and re-enables the test `TestSpecification_NumberOfTestCasesMatchesRuleInfo` since thanks to the changes from #445  its time is reduced to acceptable ranges.

This PR fixes #382


go/ct/spc/specification.go
```diff
@@ -2024,11 +2026,15 @@ func getRulesForCall(op OpCode, revision Revision, warm, zeroValue bool, opEffec
...
 func callEffect(s *st.State, addrAccessCost vm.Gas, op OpCode) {
+	isDelegateCall := op == DELEGATECALL
+
 	gas := s.Stack.Pop()
 	target := s.Stack.Pop()
 	var value U256
 	if op == CALL {
 		value = s.Stack.Pop()
+	} else if isDelegateCall {
+		value = s.Accounts.GetBalance(s.CallContext.AccountAddress)
```
HerbertJordan, line 2037: This specification is wrong: it states that with a delegate call all the money of the calling account is to be transferred.

However, what should be passed as a value is the CallValue of the current call (`s.CallContext.Value`).

go/ct/spc/specification.go
```diff
@@ -2045,15 +2051,17 @@ func callEffect(s *st.State, addrAccessCost vm.Gas, op OpCode) {
 		memoryExpansionCost = outputMemoryExpansionCost
 	}
 
+	isValueZero := value.IsZero()
```
HerbertJordan, line 2054: Delegate calls are never transferring any value since the source and destination account are the same. So I think it would be simpler to not set the value in line 2037 to something in case of a delegate call, and instead leave it to be 0.

In the condition starting at line 2119 you can then update the value to the value taken from the call context.

go/ct/spc/specification.go
```diff
@@ -2102,19 +2110,29 @@ func callEffect(s *st.State, addrAccessCost vm.Gas, op OpCode) {
 		}
 	}
 
+	sender := s.CallContext.CallerAddress
```
HerbertJordan, line 2113: This changes the sender from the `AccountAddress` to the `CallerAddress` for regular calls and static calls. This is likely wrong and should be detected as a difference in the specification and the geth implementation.

go/vm/geth/ct.go
```diff
@@ -202,7 +202,27 @@ func (i *callInterceptor) CallCode(env *geth_vm.EVM, me geth_vm.ContractRef, add
 }
 
 func (i *callInterceptor) DelegateCall(env *geth_vm.EVM, me geth_vm.ContractRef, addr geth_common.Address, data []byte, gas uint64) ([]byte, uint64, error) {
-	panic("not implemented")
+	kind := vm.DelegateCall
+	if i.static {
+		kind = vm.StaticCall
+	}
```
HerbertJordan, line 208: Is this true? Why would DelegateCalls in static contexts be replaced by Static calls?

go/vm/geth/ct.go
```diff
@@ -202,7 +202,27 @@ func (i *callInterceptor) CallCode(env *geth_vm.EVM, me geth_vm.ContractRef, add
...
+		Input:     data,
+		Gas:       vm.Gas(gas),
+	})
+
+	i.handleGasRefund(res.GasRefund)
+	err := geth_vm.ErrExecutionReverted
+	if res.Success {
+		err = nil
+	}
+	return res.Output, uint64(res.GasLeft), err
```
HerbertJordan, line 225: This code is similar to the `Call` implementation above and the static-call below. Can this be unified somehow?

HerbertJordan (changes requested):
This specification seems to be incorrect. Please have a look.

go/vm/geth/ct.go
```diff
@@ -163,9 +163,26 @@ func convertGethStackToCtStack(state *geth_vm.GethState, stack *st.Stack) *st.St
...
 type callInterceptor struct {
-	context vm.RunContext
-	stateDb *stateDbAdapter
-	static  bool
+	parameters vm.Parameters
+	stateDb    *stateDbAdapter
+	static     bool
+}
+
+func (i *callInterceptor) makeCall(kind vm.CallKind, value vm.Value, sender, recipient vm.Address, data []byte, gas uint64) ([]byte, uint64, error) {
```
HerbertJordan, line 171: nit: the parameters `sender`, `recipient`, `value`, `data`, and `gas` are passed to create a `vm.CallParameter` struct. I think it would be nicer to pass them already as a strut, since it would give callers a sort of named-parameter syntax, documenting the parameter use.

HerbertJordan (approved):
Looks good, thanks.

go/ct/spc/specification_test.go
```diff
@@ -185,7 +185,7 @@ func TestSpecification_AtMostOneCodeAtPC(t *testing.T) {
 }
 
 // TODO: re enable this test once it's runtime is reduced
```
simonlechner, line 187: This comment should be removed since the test has been reenabled.

simonlechner (approved):

HerbertJordan (dismissed):

simonlechner (approved):

PR #454, opened 2024-05-06
CT: add CREATE2 to specification

### LINKED ISSUE #385: CT: add specification of the CREATE2 instruction
The specification of the CREATE2 instruction in Tosca CT is missing.

simonlechner:
This PR adds the CREATE2 opcode to the CT specification. 
Fixes #385.

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -1455,6 +1455,71 @@ func getAllRules() []Rule {
 		},
 	})...)
 
+	// --- CREATE2 ---
+
+	rules = append(rules, rulesFor(instruction{
+		op:        CREATE2,
+		staticGas: 32000,
+		pops:      4,
+		pushes:    1,
+		conditions: []Condition{
+			Eq(ReadOnly(), false),
+		},
+		parameters: []Parameter{
+			ValueParameter{},
+			MemoryOffsetParameter{},
+			MemorySizeParameter{},
+			NumericParameter{},
+		},
+		effect: func(s *st.State) {
+			valueU256 := s.Stack.Pop()
+			offsetU256 := s.Stack.Pop()
+			sizeU256 := s.Stack.Pop()
+			saltU256 := s.Stack.Pop()
+
+			memExpCost, offset, size := s.Memory.ExpansionCosts(offsetU256, sizeU256)
+			dynamicGas := 6*((size+31)/32) + uint64(memExpCost)
+
+			if s.Gas < vm.Gas(dynamicGas) {
+				s.Status = st.Failed
+				s.Gas = 0
+				return
+			}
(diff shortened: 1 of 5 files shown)


go/ct/spc/specification.go
```diff
@@ -1455,6 +1455,89 @@ func getAllRules() []Rule {
...
+			MemoryOffsetParameter{},
+			MemorySizeParameter{},
+			NumericParameter{},
+		},
+		effect: FailEffect().Apply,
+	})...)
+
+	rules = append(rules, rulesFor(instruction{
+		op:        CREATE2,
+		staticGas: 32000,
```
facuMH, line 1480: suggestion: `name: "_non_static"`

go/ct/spc/specification.go
```diff
@@ -1455,6 +1455,89 @@ func getAllRules() []Rule {
...
+		conditions: []Condition{
+			Eq(ReadOnly(), false),
+		},
+		parameters: []Parameter{
+			ValueParameter{},
+			MemoryOffsetParameter{},
+			MemorySizeParameter{},
+			NumericParameter{},
+		},
+		effect: func(s *st.State) {
```
facuMH, line 1492: `CREATE` and `CREATE2` effects are quite similar, do you think you could unify them ?

go/vm/lfvm/instructions.go
```diff
@@ -814,6 +814,12 @@ func opCreate2(c *context) {
 		return
 	}
 
+	// Charge for the code size
+	words := (size.Uint64() + 31) / 32
```
facuMH, line 818: if this change is a fix on the VM it should be a separate PR.

facuMH (changes requested):

go/vm/lfvm/instructions.go
simonlechner: There was no newly found bug in the VM implementation. But the changes performed in #446 to CREATE where also applied to CREATE2. This was done without testing the CREATE2 changes and the order had to be updated which is done in this PR.

go/ct/spc/specification.go
```diff
@@ -1455,6 +1455,89 @@ func getAllRules() []Rule {
 		},
 	})...)
 
+	// --- CREATE2 ---
+
+	rules = append(rules, rulesFor(instruction{
+		op:        CREATE2,
+		name:      "_staticcall",
```
HerbertJordan, line 1462: The term `staticcall` might be misleading since there is a `STATICCALL` instruction. How about just "_static"?

HerbertJordan (dismissed):
Thanks, please consider the refactoring feedback provided by @facuMH before merging.

facuMH (approved):
thanks for the fixes !!

PR #456, opened 2024-05-07
LFVM: check for balance in callcode

facuMH:
LFVM did not handle the CALLCODE instruction correctly if a value higher than the accounts balance has been sent. This is fixed with this PR.
We also add here a regression test for the CT


regression_inputs/callcode_#455.json
simonlechner, line 1: The regression inputs are named after the PR which fixes the bug, this should be #456 in this case. Several fields could be removed such as Storage, BockContext, Logs and SelfDestructed.

simonlechner (changes requested):
Some remarks on the regression input file.

simonlechner (approved):

PR #457, opened 2024-05-08
Refactor st state tests

### LINKED ISSUE #433: CT: cleanup of st.State tests
Many tests in [state_test.go](https://github.com/Fantom-foundation/Tosca/blob/main/go/ct/st/state_test.go) could benefit from a table-driven testing approach.

I noticed in several PR reviews that those tests, as they are structured now, are not effective in avoiding implementation issues in `Clone`, `Eq`, and `Diff` functions. The tests should be improved to strengthen their effectiveness on detecting such issues.

simonlechner:
Using a table driven approach to test the st state `Eq`, `Diff` and `Clone` functions.
Each field is only changed in one way, because the fields implementation of equal and diff is already tested with its own unit tests.
Fixes #433.

### DIFF
--- a/go/ct/st/serialization_test.go
+++ b/go/ct/st/serialization_test.go
@@ -23,38 +23,6 @@ import (
 )
 
 ////////////////////////////////////////////////////////////
-// Helper functions
-
-func getNewFilledState() *State {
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Thanks for cleaning this up!

facuMH (approved):
nice change!! thanks

PR #458, opened 2024-05-08
Add check ensuring the right number of opcodes

simonlechner:
Due to the huge state space finding missing opcodes requires the evaluation of millions of states. A simple check is introduced which ensures that the number of defined opcodes is consistent with evm codes website.
Has to be merged after the implementation of BLOCKHASH.

### DIFF
--- a/go/ct/common/opcodes_test.go
+++ b/go/ct/common/opcodes_test.go
@@ -65,3 +65,50 @@ func TestOpCode_CanBePrinted(t *testing.T) {
 		}
 	}
 }
+
+func TestOpCode_NumberOfOpCodes(t *testing.T) {
+	istanbulOpCodes := []OpCode{
+		STOP, ADD, MUL, SUB, DIV, SDIV, MOD, SMOD, ADDMOD, MULMOD, EXP, SIGNEXTEND,
+		LT, GT, SLT, SGT, EQ, ISZERO, AND, OR, XOR, NOT, BYTE, SHL, SHR, SAR,
+		SHA3,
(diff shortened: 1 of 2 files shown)


go/ct/common/opcodes_test.go
HerbertJordan, line 1: The PR description states that this change is checking the number of *implemented* OpCodes. However, the test only covers the number (actually the set) of defined OpCodes.

Should we add a test in the specification package that checks that all valid OpCodes have some rules?

HerbertJordan (approved):

go/ct/common/opcodes_test.go
simonlechner: I changed the PR description to "the number of defined opcodes". 

Yes, such a test could be a nice addition.

facuMH:
`BLOCKHASH` has already been merged. And now there is support for Paris and Shanghai revisions. Maybe we can rebase this PR.

go/ct/common/opcodes_test.go
```diff
@@ -65,3 +65,51 @@ func TestOpCode_CanBePrinted(t *testing.T) {
 		}
 	}
 }
+
+func TestOpCode_NumberOfOpCodes(t *testing.T) {
+	istanbulOpCodes := []OpCode{
+		0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B,
```
HerbertJordan, line 71: It would be great to have the actual OpCode constants here, instead of their values.

So, for instance `[]OpCode{STOP, ADD, MUL}`.

HerbertJordan (changes requested):

HerbertJordan (approved):
Thanks!

facuMH (approved):

PR #461, opened 2024-05-14
evmzero:  Add check for range in blockHash op

facuMH:
This PR fixes the BLOCKHASH operation by adding a check that the number of the block to be hashed is within 256 of the current block.
This PR adds a test for the case when the block to be hashed is outside of this range, in this case the returned hash is 0 and a regression test case for the CT.

### DIFF
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -864,9 +864,17 @@ struct Impl<OpCode::BLOCKHASH> {
   constexpr static OpInfo kInfo = UnaryOp(20);
 
   static OpResult Run(uint256_t* top, Context& ctx) noexcept {
+    if (top[0] > std::numeric_limits<int64_t>::max()) {
+      top[0] = 0;
+      return {};
+    }
+
     int64_t number = static_cast<int64_t>(top[0]);
     int64_t upper, lower;
     upper = int64_t(ctx.host->get_tx_context().block_number);
+    // BLOCKHASH description states that it gets the hash of a block not older
+    // than 256 in comparison with the current block. and in case the current
+    // block number is less than 256, then the lower boundary is 0.
     if (upper < 257) {
       lower = 0;
     } else {
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -865,7 +865,18 @@ struct Impl<OpCode::BLOCKHASH> {
 
   static OpResult Run(uint256_t* top, Context& ctx) noexcept {
     int64_t number = static_cast<int64_t>(top[0]);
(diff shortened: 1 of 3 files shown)


cpp/vm/evmzero/interpreter.cc
```diff
@@ -865,7 +865,18 @@ struct Impl<OpCode::BLOCKHASH> {
 
   static OpResult Run(uint256_t* top, Context& ctx) noexcept {
     int64_t number = static_cast<int64_t>(top[0]);
```
HerbertJordan, line 867: This function does not check for overflows, but I guess it should.

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -2845,12 +2848,29 @@ TEST(InterpreterTest, BLOCKHASH) {
...
+      .code = {op::BLOCKHASH},
+      .state_after = RunState::kDone,
+      .gas_before = 40,
+      .gas_after = 20,
+      .stack_before = {500},
+      .stack_after = {0},
+      .host = &host,
+  });
+}
+
```
HerbertJordan, line 2873: Please also add test cases for edge cases where the provided number is
- equal to the current block - {0, 1, 255, 256, 257}
- an overflow that would result in a valid value if truncated

HerbertJordan (changes requested):

cpp/vm/evmzero/interpreter.cc
```diff
@@ -864,8 +864,24 @@ struct Impl<OpCode::BLOCKHASH> {
...
+
     int64_t number = static_cast<int64_t>(top[0]);
-    top[0] = ToUint256(ctx.host->get_block_hash(number));
+    int64_t upper, lower;
+    upper = int64_t(ctx.host->get_tx_context().block_number);
+    if (upper < 257) {
+      lower = 0;
+    } else {
+      lower = upper - 256;
+    }
```
simonlechner, line 879: I think the reason why these checks are necessary deserves a comment.

HerbertJordan (approved):
Thanks for fixing this.

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -2845,12 +2849,87 @@ TEST(InterpreterTest, BLOCKHASH) {
...
+TEST(InterpreterTest, BLOCKHASH_overflow) {
+  // const uint64_t block_number = 1000;
+  MockHost host;
+  // EXPECT_CALL(host, get_tx_context())  //
+  //     .Times(1)
+  //     .WillOnce(Return(evmc_tx_context{.block_number = block_number}));
```
simonlechner, line 2920: Should this commented code be removed?

LuisPH3 (approved):

HerbertJordan (approved):
Looks good, thanks.

PR #462, opened 2024-05-14
CT: Add BlockHash op

### LINKED ISSUE #459: CT: add specification of the BLOCKHASH instruction
The specification of the BLOCKHASH instruction in Tosca CT is missing.

facuMH:
This PR adds the rules for BLOCKHASH operation, a new condition to enforce whether the stack top is within or out of the range of the current block number and tests for it.

This PR fixes #459

### DIFF
--- a/go/ct/gen/block_context.go
+++ b/go/ct/gen/block_context.go
@@ -66,8 +66,8 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, 
 		if assignedValue, isBound := assignment[offsetConstraint.variable]; isBound {
 
 			assignedValueI64 := int64(assignedValue.Uint64())
-			lowerBound := assignedValueI64 - offsetConstraint.upperOffset
-			upperBound := assignedValueI64 + offsetConstraint.lowerOffset - 1
+			lowerBound := assignedValueI64 - offsetConstraint.upperOffset + 2
+			upperBound := assignedValueI64 + offsetConstraint.lowerOffset
 
 			// if the range is only one number, then the assigned value must be within the range.
 			isFixedValue := offsetConstraint.upperOffset-offsetConstraint.lowerOffset == 0
@@ -75,7 +75,7 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, 
 			if isFixedValue && !isSameAsfixed {
 				return st.BlockContext{}, ErrUnsatisfiable
 			} else if isFixedValue && isSameAsfixed {
-				blockNumberRangeSolver.AddEqualityConstraint(uint64(upperBound - 1))
+				blockNumberRangeSolver.AddEqualityConstraint(uint64(upperBound))
 			}
 
 			// in case of desired out of range and bound value too small.
@@ -101,9 +101,9 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, 
 	}
 
 	// for all non bound relevante variables, assign them a value based on the constraints.
-	for i, offsetConstraint := range b.variablesOffsetConstraints {
+	for _, offsetConstraint := range b.variablesOffsetConstraints {
 		if _, isBound := assignment[offsetConstraint.variable]; !isBound {
-			variableRangeSolver := NewRangeSolver(math.MinInt64, int64(blockNumber))
+			variableRangeSolver := NewRangeSolver[int64](math.MinInt64, math.MaxInt64)
 
 			if difference := offsetConstraint.lowerOffset - offsetConstraint.upperOffset; difference == 0 {
 				// if the difference between the two offsets is 0, then we can only generate a fix value.
@@ -117,12 +117,12 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, 
 				// if blockNumber is too small, then we can ONLY generate OVER the current block number
 				if blockNumber < 256 {
 					variableRangeSolver.AddLowerBoundary(offsetConstraint.upperOffset + int64(blockNumber))
-					variableRangeSolver.AddUpperBoundary(offsetConstraint.lowerOffset)
+					variableRangeSolver.AddUpperBoundary(math.MaxInt64)
 
 				} else {
 					// if blockNumber is large enough, we can randomly generate under or over the range.
 					// note that we can only generate under the range if the block number is larger than the upper offset.
-					if rand.Intn(2) == 0 && blockNumber-uint64(offsetConstraint.upperOffset) >= blockNumber {
+					if rand.Intn(2) == 0 && blockNumber < 256 {
 						// generate under the range
 						variableRangeSolver.AddLowerBoundary(0)
 						variableRangeSolver.AddUpperBoundary(int64(blockNumber))
@@ -140,15 +140,22 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, 
 			}
 			assignment[offsetConstraint.variable] = common.NewU256(uint64(int64(blockNumber) - newValue))
 		} else {
-			// if the variable is bound, then we need to check that it does not clash with any other constraint.
-			for _, previousConstraint := range b.variablesOffsetConstraints[:i] {
-				if previousConstraint.variable == offsetConstraint.variable {
-					if previousConstraint.lowerOffset < offsetConstraint.upperOffset ||
-						previousConstraint.upperOffset > offsetConstraint.lowerOffset {
-						return st.BlockContext{}, ErrUnsatisfiable
-					}
-				}
+			// if the variable is bound, then we need to check that the current constraint holds
+
+			wantsInRange := func(c constraintPair) bool { return c.lowerOffset >= c.upperOffset }
+			wantsFixValue := func(c constraintPair) bool { return c.lowerOffset == c.upperOffset }
+			isValueInRange := func(value int64) bool { return 256 >= value && value >= 1 }
+			boundedValueOffset := int64(blockNumber) - int64(assignment[offsetConstraint.variable].Uint64())
+
+			// if we want fixed value and the fixed value is different
+			if (wantsFixValue(offsetConstraint) && offsetConstraint.lowerOffset != boundedValueOffset) ||
+				// or if we want the value to be in range but is it not
+				(wantsInRange(offsetConstraint) && !isValueInRange(boundedValueOffset)) ||
+				// or if we want it to be out of range and it is in range
+				(!wantsInRange(offsetConstraint) && isValueInRange(boundedValueOffset)) {
+				return st.BlockContext{}, ErrUnsatisfiable
 			}
+
 		}
 	}
 
--- a/go/ct/gen/block_context.go
+++ b/go/ct/gen/block_context.go
@@ -65,6 +65,10 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, 
 	for _, offsetConstraint := range b.variablesOffsetConstraints {
 		if assignedValue, isBound := assignment[offsetConstraint.variable]; isBound {
 
+			if !assignedValue.IsUint64() {
+				return st.BlockContext{}, fmt.Errorf("assigned value to variable %v is not a uint64", offsetConstraint.variable)
+			}
+
 			assignedValueI64 := int64(assignedValue.Uint64())
 			lowerBound := assignedValueI64 - offsetConstraint.upperOffset + 2
 			upperBound := assignedValueI64 + offsetConstraint.lowerOffset
@@ -145,6 +149,9 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, 
 			wantsInRange := func(c constraintPair) bool { return c.lowerOffset >= c.upperOffset }
 			wantsFixValue := func(c constraintPair) bool { return c.lowerOffset == c.upperOffset }
 			isValueInRange := func(value int64) bool { return 256 >= value && value >= 1 }
+			if !assignment[offsetConstraint.variable].IsUint64() {
+				return st.BlockContext{}, fmt.Errorf("assigned value to variable %v is not a uint64", offsetConstraint.variable)
+			}
 			boundedValueOffset := int64(blockNumber) - int64(assignment[offsetConstraint.variable].Uint64())
 
 			// if we want fixed value and the fixed value is different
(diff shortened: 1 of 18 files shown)


HerbertJordan (changes requested):
I added a list of changes I think that should be applied to this PR. Please see the comments.

However, I think the chosen approach of introducing two new conditions (`WithinBlockRange` and `OutOfBlockRange`) does not really produce the tests we would like to cover for this instruction -- at least their current implementation are not testing what is desired.

Ideally, we find a condition format that would produce test cases covering the set
```
Param(0) \in { 0, CB-257, CB-256, CB-255, CB-1, CB, CB+1, 2^32 + CB - 1, 2^256-1 }
```
where `CB` is the current block number.

Maybe you can think about something to improve this? I would have to look into this myself in more detail, but I do not have the time right now to do so.

go/ct/rlz/conditions.go
```diff
@@ -719,3 +719,98 @@ func (c *hasNotSelfDestructed) GetTestValues() []TestValue {
...
+
+func (c *withinBlockRange) Restrict(generator *gen.StateGenerator) {
+	paramVariable := c.blockNumber.GetVariable()
+	c.blockNumber.BindTo(generator)
+	generator.AddBlockNumberInRange(paramVariable)
+}
+
+func (c *withinBlockRange) GetTestValues() []TestValue {
+	property := Property(c.String())
+	domain := uint64Domain{}
```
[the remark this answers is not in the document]
facuMH: A new domain is added that samples on the edge cases for both in and out of range `[CurrentBlock-257, CurrentBlock+1]`, and this domain is used to generate the test values, hence a bool is not needed.

go/ct/rlz/conditions.go
```diff
@@ -719,3 +719,98 @@ func (c *hasNotSelfDestructed) GetTestValues() []TestValue {
...
+	testValues := []TestValue{}
+	var blockNumberDomain BlockNumberRangeDomain
+	for _, value := range blockNumberDomain.Samples(0) {
+		testValues = append(testValues, NewTestValue(property, domain, value, restrict))
+	}
+	return testValues
+}
+
+func (c *inRange256FromCurrentBlock) String() string {
+	return fmt.Sprintf("blockRangeOffset(%v)", c.blockNumber)
```
HerbertJordan, line 774: nit: "inBlockRangeOffset(%v)" or "blockNumber - 256 ≤ %v ∧ %v < blockNumber" or "%v ∈ [blockNumber - 256, blockNumber-1]" or "%v ∈ [blockNumber - 256, blockNumber)"

go/ct/rlz/conditions.go
```diff
@@ -719,3 +719,98 @@ func (c *hasNotSelfDestructed) GetTestValues() []TestValue {
...
+	c.blockNumber.BindTo(generator)
+	generator.AddBlockNumberOffset(paramVariable, -1)
+}
+
+func (c *outOfRange256FromCurrentBlock) GetTestValues() []TestValue {
+	return InRange256FromCurrentBlock(c.blockNumber).GetTestValues()
+}
+
+func (c *outOfRange256FromCurrentBlock) String() string {
+	return InRange256FromCurrentBlock(c.blockNumber).String()
```
HerbertJordan, line 813: This is wrong. It should be the opposite. Something like
```
"%v ∉ [blockNumber - 256, blockNumber)"
```

go/ct/rlz/conditions.go
```diff
@@ -719,3 +719,98 @@ func (c *hasNotSelfDestructed) GetTestValues() []TestValue {
...
+		bottom = s.BlockContext.BlockNumber - 256
+	}
+	uintParam := paramBlockNumber.Uint64()
+	return uintParam >= s.BlockContext.BlockNumber || bottom > uintParam, nil
+}
+
+func (c *outOfRange256FromCurrentBlock) Restrict(generator *gen.StateGenerator) {
+	paramVariable := c.blockNumber.GetVariable()
+	c.blockNumber.BindTo(generator)
+	generator.AddBlockNumberOffset(paramVariable, -1)
```
HerbertJordan, line 805: As in the case of the `InRang256FromCurrentBlock`, this is overly restrictive.

go/ct/rlz/conditions.go
```diff
@@ -719,3 +719,98 @@ func (c *hasNotSelfDestructed) GetTestValues() []TestValue {
...
+	paramBlockNumber, err := c.blockNumber.Eval(s)
+	if err != nil {
+		return false, err
+	}
+	bottom := uint64(0)
+	if s.BlockContext.BlockNumber > 256 {
+		bottom = s.BlockContext.BlockNumber - 256
+	}
+	uintParam := paramBlockNumber.Uint64()
+	return uintParam >= s.BlockContext.BlockNumber || bottom > uintParam, nil
```
HerbertJordan, line 799: This should be simplified by reusing `inRange256FromCurrentBlock`'s implementation of the `Check` function:
```
res, err := InRange256FromCurrentBlock(c.blockNumber).Check(s)
return !res, err
```

go/ct/rlz/conditions_test.go
```diff
@@ -288,3 +288,55 @@ func TestHasSelfDestructedCondition_HasSelfDestructRestrictsGeneratedStateToBeSe
 		t.Error("generated state does not satisfy condition")
 	}
 }
+
+func TestValidRange256_Check(t *testing.T) {
```
HerbertJordan, line 292: This is a very high-level end-to-end test for the new conditions. It would be good to test Check, Restrict, and GetTestValues independently as well.

HerbertJordan (changes requested):
I think there are a range of improvements that should still be applied before this can be merged. Please have a look at individual comments.

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,85 @@
...
 
 func NewBlockContextGenerator() *BlockContextGenerator {
-	return &BlockContextGenerator{}
+	maxBlockNumber, _ := common.GetForkBlock(common.R99_UnknownNextRevision)
+	return &BlockContextGenerator{blockNumbers: NewRangeSolver(0, maxBlockNumber),
+		variablesOffsetConstraints: []variableOffset{}}
 }
 
-func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision) (st.BlockContext, error) {
+func isWithin256Range(offset int64, min, max uint64) bool {
```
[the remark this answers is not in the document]
facuMH: mmmm I see the confusion. My interpretation was that `offset` is in the range `[min, max)`. For this case, I think `isInRange(offset, min, max)` or `isInRange(min, max, offset)` are good options.

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,85 @@
...
+			// both not in range of block number
+			if !(isWithin256Range(offsetConstraint.offset, min, blockNumber) &&
+				isWithin256Range(int64(assignment[offsetConstraint.variable].Uint64()), min, blockNumber)) ||
+				// or not both out of range from block number
+				!(!isWithin256Range(offsetConstraint.offset, min, blockNumber) &&
+					!isWithin256Range(int64(assignment[offsetConstraint.variable].Uint64()), min, blockNumber)) {
+				return st.BlockContext{}, ErrUnsatisfiable
+			}
+			// else the existing value is already where we want it.
+		}
```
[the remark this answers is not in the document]
facuMH: I don't think so because `cur` could be a different number within the desired in/out of range condition. Maybe the comments should 
```
// If the variable is already bound, check if its value and the offset are both within the range of the block number            // or both out of the range. If neither is true, return an error.
```

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,153 @@
...
+	}
+
+	// generate block number
+	blockNumber, err = blockNumberRangeSolver.Generate(rnd)
 	if err != nil {
 		return st.BlockContext{}, err
 	}
+
+	variableRangeSolver := NewRangeSolver(math.MinInt64, int64(blockNumber))
+	baseSolver := NewRangeSolver(math.MinInt64, int64(blockNumber))
```
LuisPH3, line 119: Construction of this solver is trivial, I would construct a new one on each iteration.

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,153 @@
...
+	// for all non bound relevante variables, assign them a value based on the constraints.
+	for i, offsetConstraint := range b.variablesOffsetConstraints {
+		if _, isBound := assignment[offsetConstraint.lower.variable]; !isBound {
+			variableRangeSolver.Restore(baseSolver)
+			if difference := offsetConstraint.lower.offset - offsetConstraint.upper.offset; difference == 2 || difference == -2 {
+				// if the difference between the two offsets is 2, then we can only generate a fix value.
+				variableRangeSolver.AddEqualityConstraint(offsetConstraint.lower.offset - 1)
```
LuisPH3, line 126: lower -1 ? is this out of bounds? 
should it be lower +1 or upper -1 ?
I may be wrong.....

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,153 @@
...
+			} else if offsetConstraint.lower.offset > 1 && offsetConstraint.upper.offset < 256 {
+				// if lower offset is greater than 1 and upper is less than 256 we generate in range.
+				variableRangeSolver.AddLowerBoundary(offsetConstraint.upper.offset)
+				variableRangeSolver.AddUpperBoundary(offsetConstraint.lower.offset)
+			} else {
```
LuisPH3, line 131: There is something fishy in this code. The member `kind` of the `variableInequality` is never accessed.
I am not sure if this does the right thing for all scenarios, since the value is never used.  It is quite complicated as well. 

I presume this code could be simplified with the help of some member function in the `variableInequality`, and/or in the `constraintPair`, where get the bounded range.
Such a function would be much easier to unit test as well.  And  then the block in the loop turns into:
```
if _, isBound := assignment[offsetConstraint.lower.variable]; !isBound {
     newValue, err := offsetConstraint.Generate(rnd)
```

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,153 @@
...
+			}
+			newValue, err := variableRangeSolver.Generate(rnd)
+			if err != nil {
+				return st.BlockContext{}, err
+			}
+			assignment[offsetConstraint.lower.variable] = common.NewU256(uint64(int64(blockNumber) - newValue))
+		} else {
+			// if the variable is bound, then we need to check that it does not clash with any other constraint.
+			for _, previousConstraint := range b.variablesOffsetConstraints[:i] {
+				if previousConstraint.lower.variable == offsetConstraint.lower.variable {
```
LuisPH3, line 151: Same applies here, a member function in the `constraintPair` would make this test easier.

LuisPH3 (changes requested):
I saw something I did not complete understood:  Long story short, we have hard-coded some logic that should be computed based on the `kind` member of `variableInequality`:

- Your tests seem to be testing the relevant value generation
- Nevertheless, the `kind` member is never accessed, 

This makes me think that the arithmetic is hard-coded into the generator, assigning semantics to lower/upper members, this is not bad, but if the code is complete and working we could remove the `variableInequality` or use it instead:

I think the problem comes from the pair lower, and upper bounds, and assigning semantics to them. If instead you have an array of `inequalityType` and a boolean op, the list could define the range (or out of range)  we are looking for: 

- `[ (x > 0)  ,  (x  < 266)] +  AND=> in range`
- `[ (x < 1) , (x  > 265)]  + OR=> or range`

Anyway, I am pretty sure that @HerbertJordan has a better overview about how this system should be used.

go/ct/gen/block_context.go
facuMH: short answer is `lower.offset` is not the same as "lower bound". But let's discuss this offline and make it clear. Maybe the sintax or naming should be changed.

go/ct/gen/block_context.go
LuisPH3: As discussed, here we are dealing with the tradeoff between implementing a generator that can deal with inequalities vs the ad-hoc implementation for this routine. 
because of time constraints, we leave the ad-hoc implementation

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,85 @@
...
+		variablesOffsetConstraints: []variableOffset{}}
 }
 
-func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision) (st.BlockContext, error) {
+func isWithin256Range(offset int64, min, max uint64) bool {
+	val := max - uint64(offset)
+	return min <= val && val < max
+}
+
+func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand, revision common.Revision) (st.BlockContext, error) {
```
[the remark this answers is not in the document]
facuMH: after a discussion offline, we decided to first check for any bounded variable before the call to the block context generator, and if there are any, constraint the generation of the `BlockNumber` to that pre-bounded variable (as well as the current revision). And it only becomes an unsatisfiable condition if `a+256` is still less than the first block number of the current revision, or `a-1` is greater than the last block number of the current revision.

LuisPH3 (approved):

go/ct/gen/block_context.go
```diff
@@ -13,43 +13,164 @@
...
+			newValue, err := variableRangeSolver.Generate(rnd)
+			if err != nil {
+				return st.BlockContext{}, err
+			}
+			assignment[offsetConstraint.variable] = common.NewU256(uint64(int64(blockNumber) - newValue))
+		} else {
+			// if the variable is bound, then we need to check that it does not clash with any other constraint.
+			for _, previousConstraint := range b.variablesOffsetConstraints[:i] {
+				if previousConstraint.variable == offsetConstraint.variable {
+
```
LuisPH3, line 146: Can you explain to me why is it not enough with testing that the assigned value is in or out of the constraint range? 
as I understand, both B and the offset variable ($a) have been assigned at this point, and the only reason why it would not be satisfiable is if $a is not in the constraint, right?

LuisPH3 (approved):

go/ct/gen/block_context.go
facuMH: discusses offline. We agreed on a simpler way to check this.

HerbertJordan (commented):
I have reviewed everything but the generator by now. Since it is already taking me a long time, I am submitting the current comments I have so fare. I still need to look into the generator.

facuMH:
This PR got too big and mixed files and issues. It will be split into several smaller PRs, to facilitate review and understanding.

PR #469, opened 2024-05-21
LFVM:  revision enum

### LINKED ISSUE #396: lfvm: improve internal handling of revisions
The LFVM implementation inherited its support for multiple revisions from geth. As such, it utilizes a set of boolean flags named `isLondon` or `isBerlin` to indicate the active revision. This, however, is not well scalable.

To improve the handling of revisions, a revision enum should be integrated -- ideally by reusing the one provided by Tosca's `vm` package.

facuMH:
This PR changes the revision bools for vm.Revision enums in lfvm

This PR fixes #396

### DIFF
--- a/go/vm/lfvm/ct.go
+++ b/go/vm/lfvm/ct.go
@@ -77,8 +77,7 @@ func (a ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 		memory:      memory,
 		status:      RUNNING,
 		code:        converted,
-		isBerlin:    params.Revision >= vm.R09_Berlin,
-		isLondon:    params.Revision >= vm.R10_London,
+		revision:    params.Revision,
 		return_data: state.LastCallReturnData.ToBytes(),
 	}
 
--- a/go/vm/lfvm/gas.go
+++ b/go/vm/lfvm/gas.go
@@ -333,7 +333,7 @@ func gasSStoreEIP2200(c *context) (vm.Gas, error) {
 func gasSStoreEIP2929(c *context) (vm.Gas, error) {
 
 	clearingRefund := vm.Gas(params.SstoreClearsScheduleRefundEIP2200)
-	if c.revision == vm.R10_London {
+	if c.isLondon() {
 		clearingRefund = vm.Gas(params.SstoreClearsScheduleRefundEIP3529)
 	}
 
@@ -393,7 +393,7 @@ func gasSStoreEIP2929(c *context) (vm.Gas, error) {
 }
 
 func gasEip2929AccountCheck(c *context, address vm.Address) error {
-	if c.revision >= vm.R09_Berlin {
+	if c.isBerlin() {
 		// Charge extra for cold locations.
 		if !c.context.IsAddressInAccessList(address) {
 			if !c.UseGas(vm.Gas(params.ColdAccountAccessCostEIP2929 - params.WarmStorageReadCostEIP2929)) {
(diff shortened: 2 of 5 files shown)


simonlechner (dismissed):
LGTM

go/vm/lfvm/instructions.go
```diff
@@ -617,7 +617,7 @@ func opSelfbalance(c *context) {
 }
 
 func opBaseFee(c *context) {
-	if c.isLondon {
+	if c.revision == vm.R10_London {
```
HerbertJordan, line 620: The flag `isXXX` was set if the revision was `XXX` or higher. So I guess this needs to be 
```suggestion
	if c.revision >= vm.R10_London {
```

go/vm/lfvm/gas.go
```diff
@@ -333,7 +333,7 @@ func gasSStoreEIP2200(c *context) (vm.Gas, error) {
 func gasSStoreEIP2929(c *context) (vm.Gas, error) {
 
 	clearingRefund := vm.Gas(params.SstoreClearsScheduleRefundEIP2200)
-	if c.isLondon {
+	if c.revision == vm.R10_London {
```
HerbertJordan, line 336:
```suggestion
	if c.revision >= vm.R10_London {
```

HerbertJordan (dismissed):

go/vm/lfvm/instructions.go
LuisPH3: Just a question:
Changes in the revision are assumed to be valid in further revisions? 
rephrasing: Is the OpCode revision system always incremental?

go/vm/lfvm/ct.go
```diff
@@ -77,8 +77,7 @@ func (a ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 		memory:      memory,
 		status:      RUNNING,
 		code:        converted,
-		isBerlin:    params.Revision >= vm.R09_Berlin,
-		isLondon:    params.Revision >= vm.R10_London,
+		revision:    params.Revision,
 		return_data: state.LastCallReturnData.ToBytes(),
 	}
```
LuisPH3, line 82: I see that the new enums are more robust, but the comparison <= or ==  is used quite often later on.
Instead of comparing the enum every time, potentially writting it wrong. Would it make sense to make a function (either member or freestanding) that returns bool whenever the context is Berlin or London capable? This way the comparison is written once. But the option to write a switch somewhere down the line is still there.

LuisPH3 (approved):
Once fixed comparisons, it seems good to me. 
my suggestion is just a nitpick.

HerbertJordan (approved):

go/vm/lfvm/instructions.go
HerbertJordan: Those changes may be backward incompatible.

go/vm/lfvm/interpreter.go
```diff
@@ -87,6 +86,14 @@ func (c *context) SignalError(err error) {
 	c.status = ERROR
 }
 
+func (c *context) isBerlin() bool {
```
HerbertJordan, line 89: Just a though: how about calling this `isBerlinOrLater()`?

HerbertJordan: In some way, the previous version that used conditions like `c.revision >= vm.R09_Berlin` captured the right thought and it is quite readable. Maybe the functions should be removed again? Wdyt?

PR #477, opened 2024-05-28
Replace unbounded code conversion cache with limited LRU cache

### LINKED ISSUE #314: LFVM: the code converte is accumulating an unbound number of converted codes
When converting EVM bytecode into LFVM code, LFVM is retaining converted code in a [cache](https://github.com/Fantom-foundation/Tosca/blob/723021f7de4654743f80395dbb623fe4918cba75/go/vm/lfvm/converter.go#L21). This cache is currently unbound in size, and can grow to several GB when processing a large range of blocks. Also, it represents a security vulnerability for applications using LFVM.

A limit on the number of cached converted codes should be introduced.

HerbertJordan:
This PR replaces the former unbound cache for converted codes in the LFVM interpreter with a fixed-sized LRU cache. Additionally, some unnecessary debug code was removed.

This fixes #314.

The PR also adds benchmarks for the cache to identify the performance impact:
```
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/vm/lfvm
                              │   old.txt   │               new.txt               │
                              │   sec/op    │   sec/op     vs base                │
ConversionCacheLookupSpeed-12   31.61n ± 4%   51.20n ± 4%  +61.97% (p=0.000 n=10)
ConversionCacheUpdateSpeed-12   573.1n ± 6%   427.7n ± 6%  -25.38% (p=0.000 n=10)
geomean                         134.6n        148.0n        +9.94%

                              │   old.txt    │               new.txt                │
                              │     B/op     │    B/op     vs base                  │
ConversionCacheLookupSpeed-12   0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
ConversionCacheUpdateSpeed-12   318.0 ± 8%     112.0 ± 0%  -64.78% (p=0.000 n=10)
geomean                                    ²               -40.65%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                              │   old.txt    │            new.txt             │
                              │  allocs/op   │ allocs/op   vs base            │
ConversionCacheLookupSpeed-12   0.000 ± 0%     0.000 ± 0%  ~ (p=1.000 n=10) ¹
ConversionCacheUpdateSpeed-12   0.000 ± 0%     1.000 ± 0%  ? (p=0.000 n=10)
geomean                                    ²               ?                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

By changing to the utilization of an LRU instead of the previous map, an ~62% increase in the lookup time is observed. This is likely due to the additional need to track the least-recently-used order. However, as shown in the following result, this does not have any significant impact on the interpreter performance when running codes.
```
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/vm/test
               │   old.txt   │              new.txt               │
Inc/1/lfvm-12    3.373µ ± 5%   3.327µ ± 9%       ~ (p=0.393 n=10)
Inc/10/lfvm-12   3.263µ ± 7%   3.389µ ± 4%       ~ (p=0.123 n=10)
geomean          3.317µ        3.357µ       +1.21%
```
The `Inc` stress-tests the interpreter startup and shutdown time by running a simple code merely increasing a given value by a fixed offset and returning the result.

Remaining tasks:
- [x] check end-to-end impact using micro benchmarks
- [x] test potential impact on CT tests
- [x] do a test-run on historic data

### DIFF
--- a/go/vm/lfvm/converter.go
+++ b/go/vm/lfvm/converter.go
@@ -28,10 +28,12 @@ import (
 	evm "github.com/ethereum/go-ethereum/core/vm"
 )
 
+const codeCacheCapacity = 50_000 // up to ~ 1 GB for 22 KiB max code size
+
 var cache *lru.Cache[vm.Hash, Code]
 
 func init() {
-	res, err := lru.New[vm.Hash, Code](50_000) // up to ~ 1 GB for 22 KiB max code size
+	res, err := lru.New[vm.Hash, Code](codeCacheCapacity)
 	if err != nil {
 		panic(fmt.Errorf("failed to create cache: %v", err))
 	}
@@ -42,11 +44,11 @@ func clearConversionCache() {
 	cache.Purge()
 }
 
-func Convert(code []byte, with_super_instructions bool, create bool, noCodeCache bool, codeHash vm.Hash) (Code, error) {
+func Convert(code []byte, withSuperInstructions bool, isInitCode bool, noCodeCache bool, codeHash vm.Hash) (Code, error) {
 	// Do not cache use-once code in create calls.
 	// In those cases the codeHash is also invalid.
-	if create || noCodeCache {
-		return convert(code, with_super_instructions)
+	if isInitCode || noCodeCache {
+		return convert(code, withSuperInstructions)
 	}
 
 	res, exists := cache.Get(codeHash)
@@ -54,11 +56,11 @@ func Convert(code []byte, with_super_instructions bool, create bool, noCodeCache
 		return res, nil
 	}
 
-	res, error := convert(code, with_super_instructions)
+	res, error := convert(code, withSuperInstructions)
 	if error != nil {
 		return nil, error
 	}
-	if !create {
+	if !isInitCode {
 		cache.Add(codeHash, res)
 	}
 	return res, nil
--- a/go/vm/lfvm/converter.go
+++ b/go/vm/lfvm/converter.go
@@ -13,16 +13,14 @@
 package lfvm
 
 import (
-	"bytes"
 	"errors"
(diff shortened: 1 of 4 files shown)


go/vm/lfvm/converter.go
```diff
@@ -13,74 +13,56 @@
...
+}
 
 func clearConversionCache() {
-	mu.Lock()
-	defer mu.Unlock()
-	cache = map[vm.Hash]cache_val{}
+	cache.Purge()
 }
 
 func Convert(code []byte, with_super_instructions bool, create bool, noCodeCache bool, codeHash vm.Hash) (Code, error) {
```
simonlechner, line 46: I am a little confused by the name `create`, at first sight I thought that the cache entry is created when it is true. But the opposite is the case.

facuMH: perhaps `isCreateCall` would be a better name  accompanied by the comment below `// Do not cache use-once code in create calls.`

facuMH (dismissed):
lgtm

LuisPH3:
The PR replaces unbound cache with bound cache. This is used during the convert phase of the lfvm which runs once before code execution. Test with repeated initializations stresses this convert phase. Nevertheless, the instruction set in the program may not stress this cache entirely.

go/vm/lfvm/converter.go
HerbertJordan: Rename to `isInitCode` and fixed the name of the `with_super_instructions` parameter as well.

LuisPH3 (approved):

facuMH (approved):

PR #478, opened 2024-05-28
Update evmzero unit tests default Revision to Shanghai

LuisPH3:
The interpreter unit test presents a series of issues:

- It assumes instruction behavior is Istanbul unless other is specified. This assumption has leaked some untested behaviors, namely the different semantics for SELFDESTRUCT across Istanbul, Berlin, & London.
- It is 5000+ plus lines, turns difficult to maintain. And it takes a long time to compile as well. 

A replacement should, at least:
- Test the normal operation for each instruction on each Revision. 
- Use fixtures/parametrized tests to avoid repetition. This could be tricky with Mocks.
- Multiple files

From @HerbertJordan feedback: 
it needs to be a structured approach; just having test defined for each pair of operation and revision is not sufficient; [value-parameterized](https://github.com/google/googletest/blob/main/docs/advanced.md#value-parameterized-tests) tests may be an option
    we need to schedule this explicitly as a task for a release; this can take some time, and should be done in a clean way, potentially broken up into multiple steps;

### DIFF
--- a/cpp/vm/evmzero/interpreter_test.cc
+++ b/cpp/vm/evmzero/interpreter_test.cc
@@ -17,6 +17,7 @@
 
 #include <evmc/evmc.hpp>
 
+#include "evmc/evmc.h"
 #include "vm/evmzero/opcodes.h"
 
 namespace tosca::evmzero {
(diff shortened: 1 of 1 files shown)


HerbertJordan (commented):
It is very difficult to judge from the diff whether this PR is changing something substantial on the code coverage. It would help if you could provide answers to these questions:
- why do we need to update the default revision?
- in what cases did you duplicate tests and create a new PreBerlin version? What was your method to get to those diffs?
- how did you get to the new values for gas prices and refunds?

LuisPH3:
This PR is overseen by #485 #481, and #486

LuisPH3:
This will be done in  #494

Issue #465, opened 2024-05-21
Implement PUSH0 instruction


HerbertJordan:
As part of the Shanghai revision, the PUSH0 instruction was added ([EIP-3855](https://eips.ethereum.org/EIPS/eip-3855)).

To support this instruction in Tosca, the following steps are required:
- update geth dependency to a version that covers Shanghai (see #394)
- add PUSH0 instruction support to LFVM + unit tests
- add PUSH0 instruction support to the evmzero + unit tests
- add PUSH0 specification to the CT spec

PR #481, opened 2024-05-29
Add evmzero support for PUSH0

LuisPH3:
Part of #465


facuMH (approved):
Looks good !! maybe add a PR description even if we had an offline discussion about it

cpp/vm/evmzero/interpreter.cc
```diff
@@ -966,6 +967,21 @@ struct Impl<OpCode::POP> {
...
+template <>
+struct Impl<OpCode::PUSH0> {
+  constexpr static OpInfo kInfo = OpInfo{
+      .pops = 0,
+      .pushes = 1,
+      .static_gas = 2,
+      .introduced_in = EVMC_SHANGHAI,
+  };
+
+  static OpResult Run(uint256_t* top, Context&) noexcept {
```
HerbertJordan, line 979: The context can be skipped.
```suggestion
  static OpResult Run(uint256_t* top) noexcept {
```

cpp/vm/evmzero/interpreter.h
```diff
@@ -54,7 +55,7 @@ struct InterpreterArgs {
   const evmc_message* message = nullptr;
   const evmc_host_interface* host_interface = nullptr;
   evmc_host_context* host_context = nullptr;
-  evmc_revision revision = EVMC_ISTANBUL;
+  evmc_revision revision = EVMC_SHANGHAI;
```
HerbertJordan, line 58: This seems unrelated to the intent of this PR. What is the motivation for this?

cpp/vm/evmzero/interpreter.h
```diff
@@ -111,7 +112,7 @@ struct Context {
 
   evmc::HostInterface* host = nullptr;
 
-  evmc_revision revision = EVMC_ISTANBUL;
+  evmc_revision revision = EVMC_SHANGHAI;
```
HerbertJordan, line 115: Same here.

HerbertJordan (changes requested):
Looks good, just a few minor comments.

cpp/vm/evmzero/interpreter.h
LuisPH3: I see a number of issues with this member, I will revert it to Istanbul until I come out with a better method to provide better revision testing for evmzero (other than CT) here: #478

Some notes about the issue: 

- The default value for revision in the context struct is always overwritten, either by the effective InterpreterArgs, or by the unit test Args: This violates the single point of truth. 
- evmzero lacks of the definition for the range of supported revisions. (an equivalent to `newestSupportedRevision` in lfvm), but instead provides a baseline revision by default.  This works in conjunction with unit test running under the assumption that instructions follow Istanbul semantics unless another revision is prescribed. This unit test fails to model semantic changes that happen for some instructions between revisions. IMO it would be beneficial to explicitly test every revision support for every instruction in the unit test level as well.

cpp/vm/evmzero/interpreter.h
LuisPH3: Answered in the previous comment.

cpp/vm/evmzero/interpreter.h
LuisPH3: Reverted, we will come back to this issue some other time.

simonlechner (approved):
LGTM

facuMH (approved):

HerbertJordan (approved):

PR #483, opened 2024-05-29
CT: define boundaries on the revision to be used in explicit test cases

facuMH:
This test was relying on the random number generator using always the same seed and being used the same number of times. As elements were added to the state generator, this test stopped working. With this fix we can always ensure that the test runs with states generated with supported revision by the CT and the implementations.

### DIFF
--- a/go/ct/evm_test.go
+++ b/go/ct/evm_test.go
@@ -37,6 +37,7 @@ var evms = map[string]ct.Evm{
 func TestCt_ExplicitCases(t *testing.T) {
 	tests := map[string]Condition{
 		"jump_to_2^32": And(
+			RevisionBounds(R07_Istanbul, NewestSupportedRevision),
 			Eq(Status(), st.Running),
 			Eq(Op(Pc()), JUMP),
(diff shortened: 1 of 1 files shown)


go/ct/evm_test.go
```diff
@@ -37,13 +37,15 @@ var evms = map[string]ct.Evm{
 func TestCt_ExplicitCases(t *testing.T) {
 	tests := map[string]Condition{
 		"jump_to_2^32": And(
+			RevisionBounds(R07_Istanbul, R10_London), // TODO: update to newer revisions as suppor is added
```
HerbertJordan, line 40: Please use the [NewestSupportedRevision](https://github.com/Fantom-foundation/Tosca/blob/bc85fbcf8f913ead471a408e81f38295f5986ac3/go/ct/common/revisions.go#L34) constant here and below.

HerbertJordan (changes requested):

HerbertJordan (approved):

LuisPH3 (approved):
What are explicit cases?

simonlechner (approved):

PR #484, opened 2024-05-29
CT: Add recent block hashes to state, and generator

facuMH:
This PR adds a list of the last 256 blocks' hashes to the CT's state and the opcode for `BLOCKHASH`. 
This list of hashes will be needed for the implementation of the rule, but since it can be in its own PR for ease of read and understanding. 

This PR contributes to #459

### DIFF
--- a/go/ct/gen/state_test.go
+++ b/go/ct/gen/state_test.go
@@ -434,3 +434,19 @@ func TestStateGenerator_ReturnDataShouldBeEmpty(t *testing.T) {
 		t.Errorf("unexpected length of generated return data, wanted %d, got %d", want, got)
 	}
 }
+
+// //////////////////////////////////////////////////////////
+// Block number hashes
+func TestStateGenerator_BlockNumberHashes(t *testing.T) {
+	newHashes := []vm.Hash{}
+	state := genRandomState(t)
+	for i := 0; i < 256; i++ {
+		if state.RecentBlockHashes[i] == (vm.Hash{}) {
+			t.Error("unexpected hash value, should not be zero")
+		}
+		if slices.Contains(newHashes, state.RecentBlockHashes[i]) {
+			t.Errorf("unexpected hash value, should be unique %v", state.RecentBlockHashes[i])
+		}
+		newHashes = append(newHashes, state.RecentBlockHashes[i])
+	}
+}
--- a/go/ct/gen/state_test.go
+++ b/go/ct/gen/state_test.go
@@ -441,9 +441,6 @@ func TestStateGenerator_BlockNumberHashes(t *testing.T) {
 	newHashes := []vm.Hash{}
 	state := genRandomState(t)
 	for i := 0; i < 256; i++ {
-		if state.RecentBlockHashes[i] == (vm.Hash{}) {
-			t.Error("unexpected hash value, should not be zero")
-		}
 		if slices.Contains(newHashes, state.RecentBlockHashes[i]) {
 			t.Errorf("unexpected hash value, should be unique %v", state.RecentBlockHashes[i])
 		}
--- a/go/ct/st/state.go
+++ b/go/ct/st/state.go
@@ -154,7 +154,6 @@ func NewState(code *Code) *State {
 		CallData:              Bytes{},
 		LastCallReturnData:    Bytes{},
 		SelfDestructedJournal: []SelfDestructEntry{},
-		RecentBlockHashes:     [256]vm.Hash{},
 	}
 }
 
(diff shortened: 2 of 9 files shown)


HerbertJordan:
Please update the name of this PR to reflect its content.

go/ct/st/state.go
```diff
@@ -341,6 +345,15 @@ func (s *State) String() string {
 	write("\tHasSelfDestructed: %v\n", s.HasSelfDestructed)
 	write("\tSelfDestructedJournal: %v\n", s.SelfDestructedJournal)
 
+	// only print if next instruction is blockhash and the top of the stack is a valid uint64
+	if s.Code != nil && s.Code.Length() > int(s.Pc) && s.Stack != nil && s.Stack.Size() > 0 {
+		offset := s.Stack.stack[s.Stack.Size()-1]
+		if s.Code.IsCode(int(s.Pc)) && OpCode(s.Code.code[s.Pc]) == BLOCKHASH &&
+			offset.IsUint64() {
+			write("\tHash of block %d-%d: %v\n", s.BlockContext.BlockNumber, offset, s.RecentBlockHashes[offset.Uint64()])
```
HerbertJordan, line 353: I do not think the `offset` can be printed using the `%d` verb. This should have been discovered by a test.

HerbertJordan (changes requested):

go/ct/st/state.go
```diff
@@ -153,6 +154,7 @@ func NewState(code *Code) *State {
 		CallData:              Bytes{},
 		LastCallReturnData:    Bytes{},
 		SelfDestructedJournal: []SelfDestructEntry{},
+		RecentBlockHashes:     [256]vm.Hash{},
```
HerbertJordan, line 157: nit: This should be unnecessary since it is equivalent to the default.

HerbertJordan (changes requested):

go/ct/gen/state_test.go
```diff
@@ -434,3 +434,19 @@ func TestStateGenerator_ReturnDataShouldBeEmpty(t *testing.T) {
...
+// //////////////////////////////////////////////////////////
+// Block number hashes
+func TestStateGenerator_BlockNumberHashes(t *testing.T) {
+	newHashes := []vm.Hash{}
+	state := genRandomState(t)
+	for i := 0; i < 256; i++ {
+		if state.RecentBlockHashes[i] == (vm.Hash{}) {
+			t.Error("unexpected hash value, should not be zero")
+		}
+		if slices.Contains(newHashes, state.RecentBlockHashes[i]) {
```
LuisPH3, line 447: I think we don't need this test, even if this test pass, this does not prevent the random number generator from making the same sequence in runtime,

LuisPH3: ?

LuisPH3 (approved):
Please comment before closing comments.

go/ct/gen/state_test.go
facuMH: sorry. I forgot to reply. Yes, you are right, the generator could potentially generate a hash that is all zeros, and that would not be an error. hence the test was removed.

go/ct/st/state.go
```diff
@@ -439,5 +451,11 @@ func (s *State) Diff(o *State) []string {
 		}
 	}
 
+	for i := 0; i < 256; i++ {
+		if s.RecentBlockHashes[i] != o.RecentBlockHashes[i] {
+			res = append(res, fmt.Sprintf("Different block number hash at index %d: %x vs %x", i, s.RecentBlockHashes[i], o.RecentBlockHashes[i]))
```
HerbertJordan, line 456:
```suggestion
	for i, want := range s.RecentBlockHashes {
		if want != o.RecentBlockHashes[i] {
			res = append(res, fmt.Sprintf("Different block number hash at index %d: %x vs %x", i, want, o.RecentBlockHashes[i]))
```

go/ct/utils/adapter.go
```diff
@@ -121,7 +121,16 @@ func (c *ctRunContext) GetTransactionContext() vm.TransactionContext {
 }
 
 func (c *ctRunContext) GetBlockHash(number int64) vm.Hash {
-	panic("not implemented")
+	min := int64(0)
+	max := int64(c.state.BlockContext.BlockNumber)
+	if c.state.BlockContext.BlockNumber > 256 {
```
HerbertJordan, line 126: nit:
```suggestion
	if max > 256 {
```

HerbertJordan (changes requested):

HerbertJordan (approved):

PR #486, opened 2024-05-29
Add CT support for PUSH0

LuisPH3:
Part of #465

### DIFF
--- a/go/ct/rlz/conditions.go
+++ b/go/ct/rlz/conditions.go
@@ -319,7 +319,8 @@ type revisionBounds struct{ min, max Revision }
 
 func RevisionBounds(min, max Revision) Condition {
 	if min > max {
-		min, max = max, min
+		// This is an specification error, and should not be silently ignored
+		panic(fmt.Sprintf("Invalid revision bounds: %v > %v", min, max))
 	}
 	return &revisionBounds{min, max}
 }
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -1414,6 +1414,7 @@ func getAllRules() []Rule {
 		pushes:    1,
 		conditions: []Condition{
 			Eq(ReadOnly(), true),
+			RevisionBounds(MinRevision, R11_Paris),
 		},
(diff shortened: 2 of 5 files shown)


facuMH (approved):
lgtm

go/ct/rlz/conditions.go
```diff
@@ -319,7 +319,8 @@ type revisionBounds struct{ min, max Revision }
 
 func RevisionBounds(min, max Revision) Condition {
 	if min > max {
-		min, max = max, min
+		// This is an specification error, and should not be silently ignored
+		panic(fmt.Sprintf("Invalid revision bounds: %v > %v", min, max))
```
HerbertJordan, line 323: Thanks for spotting this!

go/ct/spc/specification.go
```diff
@@ -1387,6 +1414,7 @@ func getAllRules() []Rule {
 		pushes:    1,
 		conditions: []Condition{
 			Eq(ReadOnly(), true),
+			RevisionBounds(MinRevision, R11_Paris),
```
HerbertJordan, line 1417: This is a temporary change, right? Until they are updated to match Shanghai?
If so, maybe add a TODO indicating that this needs to be cleaned up.

HerbertJordan (approved):

facuMH (approved):

go/ct/spc/specification.go
LuisPH3: Right, we really need to be sure that this code goes away whoever merges first.

Catch 22 scenario: 
- if I commit without bumping up newest supported: my code panics with the new check or it did something wrong without.
- If I do upgrade, the newest supported rev:  Code Size change is not yet there and ct starts failing for create. 

We flirted with the idea of adding a RevisionBounds mechanism so that if the revision is future to the currently supported one, the ct would ignore it. Nevertheless, this would require some refactoring between conditions and generators, and we discussed offline that it does not worth the effort.

simonlechner (approved):

PR #487, opened 2024-05-31
Ignore vscode settings

LuisPH3:
Although it might be useful to share settings, these often include hard-coded paths to local installed tools.
Do not commit one's personal preferences on how to configure the IDE. 

Please feel free to add to reviewers whoever may be interested in this discussion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@
 # Dependency directories (remove the comment below to include it)
 # vendor/
 build/
+
+# IDE files
+.vscode


facuMH (dismissed):

facuMH:
I agree that custom config files should not be committed. However it should be properly documented how to set them up to facilitate testing/formatting/etc.

HerbertJordan (dismissed):

.gitignore
```diff
@@ -13,4 +13,6 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
-build/
```
simonlechner, line 16: Why is the build directory no longer ignored? I think it should still be included if somebody builds Tosca using the command line.

facuMH (approved):

PR #488, opened 2024-05-31
CT: Add interval solver

facuMH:
This PR adds an interval solver. This tool can be used to describe a set of intervals and get a random number uniformly from all the possible values considering all the intervals. 

This PR contributes to #459

### DIFF
--- a/go/ct/gen/interval_solver.go
+++ b/go/ct/gen/interval_solver.go
@@ -2,16 +2,26 @@ package gen
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
 	"golang.org/x/exp/constraints"
 	"pgregory.net/rand"
 )
 
+// IntervalSolver is a generic utility to solve constraints of the form
+//
+//	X ∈ [a1..b1] ∪ [a2..b2] ∪ ... ∪ [an..bn]
+//
+// The solver maintains a list of disjoint intervals [a..b] that represent the
+// domain of the variable x. The solver can exclude ranges from the domain and
+// check if the domain is empty, making the constraints unsatisfiable.
 type IntervalSolver[T constraints.Integer] struct {
 	intervals []interval[T]
 }
 
+// NewIntervalSolver creates a new IntervalSolver with the domain [min..max].
+// If min > max, the solver is initialized with an empty domain.
 func NewIntervalSolver[T constraints.Integer](min, max T) *IntervalSolver[T] {
 	initialInterval := interval[T]{low: min, high: max}
 	if initialInterval.isEmpty() {
@@ -22,7 +32,10 @@ func NewIntervalSolver[T constraints.Integer](min, max T) *IntervalSolver[T] {
 	}
 }
 
+// Exclude removes the range [min..max] from the domain of the solver.
+// If min > max, the interval would be empty and so is ignored.
 func (s *IntervalSolver[T]) Exclude(min, max T) {
+	// this insures an empty interval will never be added.
 	if max < min {
 		return
 	}
@@ -60,8 +73,8 @@ func (s *IntervalSolver[T]) Contains(value T) bool {
 }
 
 func (s *IntervalSolver[T]) AddEqualityConstraint(value T) {
-	s.AddLowerBoundary(value)
 	s.AddUpperBoundary(value)
+	s.AddLowerBoundary(value)
 }
 
 func (s *IntervalSolver[T]) AddLowerBoundary(value T) {
@@ -69,6 +82,9 @@ func (s *IntervalSolver[T]) AddLowerBoundary(value T) {
 		return
 	}
 	min := s.intervals[0].low
+	if value <= min {
+		return
+	}
 	s.Exclude(min, value-1)
 }
 
@@ -77,6 +93,9 @@ func (s *IntervalSolver[T]) AddUpperBoundary(value T) {
 		return
 	}
 	max := s.intervals[len(s.intervals)-1].high
+	if value >= max {
+		return
+	}
 	s.Exclude(value+1, max)
 }
 
@@ -84,16 +103,14 @@ func (s *IntervalSolver[T]) IsSatisfiable() bool {
 	return len(s.intervals) > 0
 }
 
+// Generate returns a random value from the domain of the solver.
 func (s *IntervalSolver[T]) Generate(rnd *rand.Rand) (T, error) {
-	if s.intervals == nil || len(s.intervals) == 0 {
+	if len(s.intervals) == 0 {
 		return 0, ErrUnsatisfiable
 	}
 
 	domainSize := uint64(0)
 	for _, interval := range s.intervals {
-		if interval.isEmpty() {
-			return 0, ErrUnsatisfiable
-		}
 		domainSize += uint64(interval.high - interval.low + 1)
 	}
 
@@ -123,24 +140,14 @@ func (s *IntervalSolver[T]) String() string {
 }
 
 func (s *IntervalSolver[T]) Clone() *IntervalSolver[T] {
-	return &IntervalSolver[T]{intervals: append([]interval[T]{}, s.intervals...)}
+	return &IntervalSolver[T]{intervals: slices.Clone(s.intervals)}
 }
 
 func (s *IntervalSolver[T]) Equals(other *IntervalSolver[T]) bool {
-	if len(s.intervals) != len(other.intervals) {
-		return false
-	}
-	for i, interval := range s.intervals {
-		if interval != other.intervals[i] {
-			return false
-		}
-	}
-	return true
+	return slices.Equal(s.intervals, other.intervals)
 }
 
-// [a..b] ... pick one element in the range
-// [a..b][c..d][e..f] ... pick one element in any of those ranges
-
+// interval represents an interval [a..b] of type T.
 type interval[T constraints.Integer] struct {
 	low, high T
 }
--- a/go/ct/gen/interval_solver.go
+++ b/go/ct/gen/interval_solver.go
@@ -13,7 +13,7 @@ import (
 //
 //	X ∈ [a1..b1] ∪ [a2..b2] ∪ ... ∪ [an..bn]
 //
-// The solver maintains a list of disjoint intervals [a..b] that represent the
+// The solver maintains a list of inclusive intervals [a..b] that represent the
 // domain of the variable x. The solver can exclude ranges from the domain and
 // check if the domain is empty, making the constraints unsatisfiable.
 type IntervalSolver[T constraints.Integer] struct {
@@ -166,7 +166,7 @@ const (
 	isEqual             relation = iota // [1..3] isEqual [1..3]
 	isBefore                            // [1..3] is before [4..6]
 	isAfter                             // [4..6] is after [1..3]
-	isSubset                            // [1..6] isContained in [2..5]
+	isSubset                            // [1..6] contains [2..5]
 	isSuperset                          // [2..5] isEnclosed in [1..6]
 	intersectsWithStart                 // [1..3] intersectsWithStart [2..6]
 	intersectsWithEnd                   // [2..6] intersectsWithEnd [1..5]
--- a/go/ct/gen/interval_solver.go
+++ b/go/ct/gen/interval_solver.go
@@ -1,0 +1,177 @@
+package gen
+
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/exp/constraints"
+	"pgregory.net/rand"
+)
+
+type IntervalSolver[T constraints.Integer] struct {
+	intervals []interval[T]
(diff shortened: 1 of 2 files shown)


go/ct/gen/interval_solver.go
```diff
@@ -0,0 +1,177 @@
...
+func (s *IntervalSolver[T]) AddUpperBoundary(value T) {
+	if len(s.intervals) == 0 {
+		return
+	}
+	max := s.intervals[len(s.intervals)-1].high
+	s.Exclude(value+1, max)
+}
+
+func (s *IntervalSolver[T]) IsSatisfiable() bool {
+	return len(s.intervals) > 0
```
LuisPH3, line 80: Because the empty interval exists, the solver may not be satisfiable if initialized with only empty intervals.

go/ct/gen/interval_solver.go
```diff
@@ -0,0 +1,177 @@
...
+	if len(s.intervals) == 0 {
+		return 0, ErrUnsatisfiable
+	}
+
+	domainSize := uint64(0)
+	for _, interval := range s.intervals {
+		domainSize += uint64(interval.high - interval.low + 1)
+	}
+
+	if domainSize == 0 { // the domain is the full uint64 domain
```
LuisPH3, line 93: If domain size is less than 0, because of invalid/empty intervals, there should be an error

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	"pgregory.net/rand"
+)
+
+func TestIntervalSolver_CanFormulateConstraints(t *testing.T) {
+	tests := map[string]struct {
+		setup func(*IntervalSolver[int32])
+		want  string
+	}{
+		"empty": {
```
LuisPH3, line 14: missleading, rename to default.

LuisPH3 (changes requested):
It is really ingenuous. 
Please solve the issue when solver is initialized to the empty interval.

go/ct/gen/interval_solver.go
facuMH: great observations !! tests and checks added for empty intervals aka `[a..b]` where `a>b`

go/ct/gen/interval_solver.go
facuMH: because `domainSize` is a `uint64` it can not be negative, it would be an overflow, which could easily be confused with a big interval, for this a check was added to verify that the interval is not empty, aka `interval.low<interval.high`

LuisPH3 (approved):

go/ct/gen/interval_solver.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/exp/constraints"
+	"pgregory.net/rand"
+)
+
+type IntervalSolver[T constraints.Integer] struct {
```
HerbertJordan, line 11: Need documentation!

HerbertJordan (changes requested):

go/ct/gen/interval_solver.go
```diff
@@ -0,0 +1,188 @@
...
+func (s *IntervalSolver[T]) Generate(rnd *rand.Rand) (T, error) {
+	if s.intervals == nil || len(s.intervals) == 0 {
+		return 0, ErrUnsatisfiable
+	}
+
+	domainSize := uint64(0)
+	for _, interval := range s.intervals {
+		if interval.isEmpty() {
+			return 0, ErrUnsatisfiable
+		}
```
HerbertJordan, line 96: This should not happen. There should be an invariant that makes sure that there is no empty interval in the interval list.

go/ct/gen/interval_solver.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	clauses := []string{}
+	for _, interval := range s.intervals {
+		clauses = append(clauses, fmt.Sprintf("[%d..%d]", interval.low, interval.high))
+	}
+	return fmt.Sprintf("X ∈ %s", strings.Join(clauses, " ∪ "))
+}
+
+func (s *IntervalSolver[T]) Clone() *IntervalSolver[T] {
+	return &IntervalSolver[T]{intervals: append([]interval[T]{}, s.intervals...)}
```
HerbertJordan, line 126: nit: use `slices.Clone(s.intervals)`

go/ct/gen/interval_solver.go
```diff
@@ -0,0 +1,188 @@
...
+func (s *IntervalSolver[T]) Equals(other *IntervalSolver[T]) bool {
+	if len(s.intervals) != len(other.intervals) {
+		return false
+	}
+	for i, interval := range s.intervals {
+		if interval != other.intervals[i] {
+			return false
+		}
+	}
+	return true
```
HerbertJordan, line 138: Use `slices.Equal()`

HerbertJordan, line 138: This function is only true if the following invariants are guaranteed:
- the intervals are sorted
- there are no overlaps; in particular, there are no empty intervals
The last one should be the case, but in the print function you are checking for this empty intervals. So this should be ensured.

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,248 @@
...
+	const N = 20
+	// remove interval [c..d] from [a..b]
+	for a := 0; a < N; a++ {
+		for b := 0; b < N; b++ {
+			for c := 0; c < N; c++ {
+				for d := 0; d < N; d++ {
+					if a > b || c > d {
+						// we skip empty intervals
+						continue
+					}
```
HerbertJordan, line 124: Remove this => the implementation must be able to handle empty intervals.

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+		}
+	}
+}
+
+func TestIntervalSolver_EmptyInterval(t *testing.T) {
+	solver := NewIntervalSolver[int32](200, 100)
+	_, err := solver.Generate(rand.New())
+	if err != ErrUnsatisfiable {
+		t.Errorf("failed to generate with empty interval: %v", err)
```
HerbertJordan, line 146:
```suggestion
		t.Errorf("failed to generate point in empty interval: %v", err)
```

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+}
+
+func TestIntervalSolver_AddEmptyIntervalHasNoEffect(t *testing.T) {
```
HerbertJordan, line 150:
```suggestion
func TestIntervalSolver_ExcludeEmptyIntervalHasNoEffect(t *testing.T) {
```

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	if solver1.Equals(solver2) {
+		t.Errorf("equals fails to report different instances")
+	}
+}
+
+func TestInterval_isEmpty(t *testing.T) {
+	tests := map[string]struct {
+		interval interval[int32]
+		want     bool
```
HerbertJordan, line 234: nit: would be more readable if you use the fields
```
low int32
high int32
empty bool
```

go/ct/gen/interval_solver_test.go
HerbertJordan, line 1: Please add tests for additional (edge) cases, like:
- intervals covering the full 64-bit range of uint64
- intervals on unsigned types and 0 as a lower end
- intervals on signed types with negative values

HerbertJordan (changes requested):

go/ct/gen/interval_solver.go
LuisPH3: The way I see it, if you have a constructor function with 2 integers, and they get misplaced, the object will be constructed in an insatisfacible state. Would it?

LuisPH3 (changes requested):
The empty interval issue is now solved!. 
I am concerned about over/under flow when moving boundaries. :(

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,326 @@
...
+			}
+		}
+	}
+}
+
+func TestIntervalSolver_EmptyInterval(t *testing.T) {
+	solver := NewIntervalSolver[int32](200, 100)
+	_, err := solver.Generate(rand.New())
+	if err != ErrUnsatisfiable {
+		t.Errorf("failed to generate point in empty interval: %v", err)
```
HerbertJordan, line 142: This error message is wrong.

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+}
+
+func TestIntervalSolver_ExcludeEmptyIntervalHasNoEffect(t *testing.T) {
+	solver := NewIntervalSolver[int32](200, 400)
+	clone := solver.Clone()
+	clone.Exclude(300, 250)
+	if !solver.Equals(clone) {
+		t.Fatalf("empty interval should not be modified by adding empty interval")
```
HerbertJordan, line 151: this error message is wrong

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,326 @@
...
+	solver := NewIntervalSolver[int32](200, 400)
+	solver.Exclude(250, 300)
+
+	rnd := rand.New()
+	for i := 0; i < 100; i++ {
+		res, err := solver.Generate(rnd)
+		if err != nil {
+			t.Fatalf("error solving intervals: %v", err)
+		}
+		if res < 200 || res > 400 || (res >= 250 && res <= 300) {
```
HerbertJordan, line 165: There should be a test case making sure that the boundaries are hit by the generation process.

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,326 @@
...
+		"lower":    func(s *IntervalSolver[int32]) { s.AddLowerBoundary(300) },
+		"upper":    func(s *IntervalSolver[int32]) { s.AddUpperBoundary(300) },
+		"equality": func(s *IntervalSolver[int32]) { s.AddEqualityConstraint(300) },
+	}
+
+	for name, setup := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			solver := NewIntervalSolver[int32](200, 400)
+			solver.intervals = nil
```
HerbertJordan, line 194: I would prefer creating the empty interval through public API.
```suggestion
			solver := NewIntervalSolver[int32](400, 200)
```

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,326 @@
...
+			check: func(v int64) bool { return v != math.MaxInt64 }},
+		"remove-max": {
+			setup: func(s *IntervalSolver[int64]) { s.Exclude(math.MaxInt64, math.MaxInt64) },
+			check: func(v int64) bool { return v != math.MaxInt64 }},
+		"remove-all-positive": {
+			setup: func(s *IntervalSolver[int64]) { s.Exclude(0, math.MaxInt64) },
+			check: func(v int64) bool { return v < 0 }},
+		"remove-all-negative": {
+			setup: func(s *IntervalSolver[int64]) { s.Exclude(math.MinInt64, 0) },
+			check: func(v int64) bool { return v >= 0 }},
```
HerbertJordan, line 303: It should actually fail if 0 is generated.

HerbertJordan (changes requested):

LuisPH3 (dismissed):
:+1:

LuisPH3 (approved):
:+1:

go/ct/gen/interval_solver_test.go
```diff
@@ -0,0 +1,498 @@
...
+func TestIntervalSolver_ExcludeEmptyIntervalHasNoEffect(t *testing.T) {
+	solver := NewIntervalSolver[int32](200, 400)
+	clone := solver.Clone()
+	clone.Exclude(300, 250)
+	if !solver.Equals(clone) {
+		t.Errorf("excluding empty interval should have no effect")
+	}
+}
+
+func TestIntervalSolver_TestIntervals(t *testing.T) {
```
HerbertJordan, line 181:
```suggestion
func TestIntervalSolver_GenerateProducesValuesFromWithinTheIntervals(t *testing.T) {
```

HerbertJordan (approved):
Thanks for cleaning this up!

go/ct/gen/interval_solver.go
simonlechner, line 1: I think the general comment should also include whether the intervals are inclusive or exclusive.

simonlechner (approved):
Just a minor comment, looks good.

LuisPH3 (approved):

PR #490, opened 2024-06-03
CT:  Add target block number and revision generation into block number generator.

facuMH:
This PR uses the newly introduced interval solver to generate a value and assign it to a variable in `assignment`. It also moves the generation of revision, and deduces it from the block number generated, which is now constrained when the revision conditions are called. 

Part of #459

### DIFF
--- a/go/ct/gen/block_context.go
+++ b/go/ct/gen/block_context.go
@@ -28,22 +28,25 @@ import (
 // BlockContextGenerator is a generator for block contexts.
 // It can take constraints on the block number, which come from
 // restricting the revision, and it can also take constraints on
-// the on different variables to generta in respect to the block number.
+// on different variables in respect to the block number to be solved for.
 // The constraints can be added in the form of fixed-value constraints
 // on variables, and constraints on the range of values that variables
 // can take.
-// The generator can be marked as unsatisfiable if the constraints
-// added to it are conflicting.
+// if conflicting constraints are added the generator will turn
+// unsatisfiable, and return an error when trying to generate a block context.
 type BlockContextGenerator struct {
 	blockNumberSolver *RangeSolver[uint64]
 
 	// This map is used to keep track of range constraints of variables.
-	// variables limited to recent block constraints are mapped to
-	// true, variables limited to be not in recent block constraints
+	// Variables limited to recent block number range are mapped to
+	// true, variables required to be not in the recent block number range
 	// are mapped to false.
 	rangeConstraints map[Variable]bool
 
 	// This map is used to keep track of fixed-value constraints of variables.
+	// the values we keep here are the offset from the block number, not the
+	// actual value, because of this, the value can be negative as well,
+	// meaaning we could want to generate a value bigger than the block number.
 	valueConstraint map[Variable]int64
 
 	// This flag is set to true if the contained set of constraints
@@ -55,37 +58,16 @@ func NewBlockContextGenerator() *BlockContextGenerator {
 	return &BlockContextGenerator{}
 }
 
-// addOffset adds an offset to a U256 value, checking for overflow.
-func addOffset(value U256, offset int64) (result U256, hasOverflown bool) {
-	if offset < 0 {
-		result := value.Sub(NewU256(uint64(-offset)))
-		return result, result.Gt(value)
-	}
-	result = value.Add(NewU256(uint64(offset)))
-	return result, result.Lt(value)
-}
-
-// addWithOverflowCheck adds two uint64 values, checking for overflow.
-func addWithOverflowCheck(a, b uint64) (result uint64, hasOverflown bool) {
-	result = a + b
-	return result, result < a
-}
-
 // generateResultingBlockNumber generates a block number based on the constraints
-// added to the generator, this could be regarding revision constraints,
-// fixed-value constraints, range constraints, or previously assigned variables
-// ( the latter only if they are also target variable for any relevant constraint )
+// added to the generator. These could be revision constraints,
+// fixed-value constraints, range constraints, or previously bound variables
+// in the given assignment.
 func (b *BlockContextGenerator) generateResultingBlockNumber(assignment Assignment, rnd *rand.Rand) (uint64, error) {
 
 	blockNumberSolver := NewIntervalSolver[uint64](0, math.MaxUint64)
 	if b.blockNumberSolver != nil {
-		// apply constraints on block number solver derived from revision constraints
-		if b.blockNumberSolver.min > 0 {
-			blockNumberSolver.Exclude(0, b.blockNumberSolver.min-1)
-		}
-		if b.blockNumberSolver.max < math.MaxUint64 {
-			blockNumberSolver.Exclude(b.blockNumberSolver.max+1, math.MaxUint64)
-		}
+		blockNumberSolver.AddLowerBoundary(b.blockNumberSolver.min)
+		blockNumberSolver.AddUpperBoundary(b.blockNumberSolver.max)
 	}
 
 	// apply constraints on block number solver derived from predefined assignments
@@ -117,11 +99,16 @@ func (b *BlockContextGenerator) generateResultingBlockNumber(assignment Assignme
 					blockNumberSolver.AddUpperBoundary(upper)
 				}
 			} else {
-				if !assignedValue.Sub(NewU256(256)).IsUint64() {
-					return 0, ErrUnsatisfiable
-				}
+				min := uint64(math.MaxUint64)
+				max := uint64(math.MaxUint64)
 				value := assignedValue.Uint64()
-				blockNumberSolver.Exclude(value+1, value+256)
+				if assignedValue.IsUint64() && value < max {
+					min = assignedValue.Uint64() + 1
+					if assignedValue.Uint64() < (max - 256) {
+						max = assignedValue.Uint64() + 256
+					}
+				}
+				blockNumberSolver.Exclude(min, max)
 			}
 		}
 	}
@@ -129,8 +116,8 @@ func (b *BlockContextGenerator) generateResultingBlockNumber(assignment Assignme
 	return blockNumberSolver.Generate(rnd)
 }
 
-// processValueConstraint processes the fixed-value constraints on variables.
-func (b *BlockContextGenerator) processValueConstraint(assignment Assignment, resultingBlockNumber uint64) error {
+// bindVariablesInValueConstraints processes the fixed-value constraints on variables.
+func (b *BlockContextGenerator) bindVariablesInValueConstraints(assignment Assignment, resultingBlockNumber uint64) error {
 	for variable, offset := range b.valueConstraint {
 		requiredValue, underflow := addOffset(NewU256(resultingBlockNumber), -offset)
 		if underflow {
@@ -149,8 +136,8 @@ func (b *BlockContextGenerator) processValueConstraint(assignment Assignment, re
 	return nil
 }
 
-// processRangeConstraint processes the range constraints on variables.
-func (b *BlockContextGenerator) processRangeConstraint(resultingBlockNumber uint64, assignment Assignment, rnd *rand.Rand) error {
+// bindVariablesInRangeConstraints processes the range constraints on variables.
+func (b *BlockContextGenerator) bindVariablesInRangeConstraints(resultingBlockNumber uint64, assignment Assignment, rnd *rand.Rand) error {
 	lowerBound := uint64(0)
 	if resultingBlockNumber > 256 {
 		lowerBound = resultingBlockNumber - 256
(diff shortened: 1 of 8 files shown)


go/ct/gen/block_context.go
```diff
@@ -68,13 +238,191 @@ func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision)
...
+		if inRange {
+			b.markUnsatisfiable()
+		}
+	} else {
+		b.rangeConstraints[variable] = false
+	}
+}
+
+// SetBlockNumberOffsetValue adds a constraint on the variable so that it is
+// generated with a value that is offset from the block number.
```
HerbertJordan, line 370: A variable is not "generated". The Generator process assigns a value to it. Please fix the comment.

go/ct/gen/block_context.go
```diff
@@ -68,13 +238,191 @@ func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision)
...
+// If the generator is already marked as unsatisfiable or if the variable
+// is already constrained to a different offset, the generator remains
+// unsatisfiable.
```
HerbertJordan, line 373:
```suggestion
```

go/ct/gen/block_context.go
```diff
@@ -68,13 +238,191 @@ func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision)
...
+	if b.unsatisfiable {
+		return
+	}
+	if lower > upper || lower < 0 || upper < 0 {
+		b.markUnsatisfiable()
+		return
+	}
+	min, _ := GetForkBlock(lower)
+	max, _ := GetForkBlock(upper)
+	len, _ := GetBlockRangeLengthFor(upper)
```
HerbertJordan, line 406: Misses error handling.

go/ct/gen/block_context.go
```diff
@@ -13,52 +13,222 @@
...
+				}
+				if upper, overflow := addWithOverflowCheck(value, 256); !overflow {
+					blockNumberSolver.AddUpperBoundary(upper)
+				}
+			} else {
+				if !assignedValue.Sub(NewU256(256)).IsUint64() {
+					return 0, ErrUnsatisfiable
+				}
+				value := assignedValue.Uint64()
+				blockNumberSolver.Exclude(value+1, value+256)
```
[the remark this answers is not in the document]
facuMH: overflow check will be added. But I used `Sub` because we are in the "out of range" case, so we want to restrict `BlockNumberSolver` to smaller numbers than `assignedValue`

go/ct/gen/block_context.go
```diff
@@ -13,52 +13,222 @@
...
+}
+
+// processRangeConstraint processes the range constraints on variables.
+func (b *BlockContextGenerator) processRangeConstraint(resultingBlockNumber uint64, assignment Assignment, rnd *rand.Rand) error {
+	lowerBound := uint64(0)
+	if resultingBlockNumber > 256 {
+		lowerBound = resultingBlockNumber - 256
+	}
+	upperBound := resultingBlockNumber
+	blockNumberRangeGenerator := NewRangeSolver(lowerBound, upperBound-1)
```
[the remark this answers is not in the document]
facuMH: if `resultingBlockNumber` is `0`, then we are the first node and there should not be any viable "previous" block to check for, in that case there should not be any condition or the condition is unfulfillable, either way, `blockNumberRangeGenerator` should be unsatisfiable, no ?

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+		"conflicting-revisions-with-in-range-and-predefined-assignment-max-uint64": func(b *BlockContextGenerator, a Assignment) {
+			b.SetRevision(common.R07_Istanbul)
+			b.RestrictVariableToOneOfTheLast256Blocks("a")
+			a["a"] = common.NewU256(0xffffffffffffffff)
+		},
+		"coflicting-assigned-value-and-fixed-offset": func(b *BlockContextGenerator, a Assignment) {
+			b.SetBlockNumberOffsetValue("a", 440)
+			b.SetBlockNumberOffsetValue("a", 40)
+		},
+	}
```
HerbertJordan, line 391: This test should be result in an unsatisfiable error, but it does not:
```
                "no-valid-block-range": func(b *BlockContextGenerator, a Assignment) {
			b.blockNumberSolver = NewRangeSolver[uint64](0, 0)
			b.RestrictVariableToOneOfTheLast256Blocks("a")
		},
```

This test case forces the resulting block number to be `0`, which is a valid block number if the constraint would be to be part of the Istanbul revision (range 0 to 1000). The random number generator picking the revision in this range may chose to pick 0.

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+			b.SetRevision(common.R07_Istanbul)
+			b.RestrictVariableToOneOfTheLast256Blocks("a")
+			a["a"] = common.NewU256(1, 1)
+		},
+		"conflicting-revisions-with-in-range-and-predefined-assignment-max-uint64": func(b *BlockContextGenerator, a Assignment) {
+			b.SetRevision(common.R07_Istanbul)
+			b.RestrictVariableToOneOfTheLast256Blocks("a")
+			a["a"] = common.NewU256(0xffffffffffffffff)
+		},
+		"coflicting-assigned-value-and-fixed-offset": func(b *BlockContextGenerator, a Assignment) {
```
HerbertJordan, line 387:
```suggestion
		"conflicting-assigned-value-and-fixed-offset": func(b *BlockContextGenerator, a Assignment) {
```

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+				b.SetRevision(common.R07_Istanbul)
+				b.RestrictVariableToNoneOfTheLast256Blocks("a")
+				a["a"] = common.NewU256(8000)
+			},
+			check: func(t *testing.T, res st.BlockContext, a Assignment) {
+				if res.BlockNumber >= 1000 {
+					t.Errorf("produced block number is not a valid Istanbul block, got %d", res.BlockNumber)
+				}
+			},
+		},
```
HerbertJordan, line 319: This test fails, though I think it should not:
```suggestion
		},
		"tight-block-range-with-single-solution": {
			setup: func(b *BlockContextGenerator, a Assignment) {
				b.blockNumberSolver = NewRangeSolver[uint64](0, 1)
				b.RestrictVariableToOneOfTheLast256Blocks("a")
			},
			check: func(t *testing.T, res st.BlockContext, a Assignment) {
				if res.BlockNumber != 1 {
					t.Errorf("expected block number to be 1, got %d", res.BlockNumber)
				}
				if value := a["a"]; value != common.NewU256(0) {
					t.Errorf("expected variable 'a' to be assigned to 0, got %v", value)
				}
			},
		},
```

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+			overflow: false,
+		},
+		"overflow": {
+			base: common.NewU256(0xffffffffffffffff, 0xffffffffffffffff,
+				0xffffffffffffffff, 0xffffffffffffffff),
+			offset:   1,
+			want:     common.NewU256(0),
+			overflow: true,
+		},
+		"underdflow": {
```
HerbertJordan, line 446:
```suggestion
		"underflow": {
```

go/ct/gen/block_context.go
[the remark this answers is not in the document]
HerbertJordan: I added two test cases which, I think, demonstrate the problem here.

go/ct/gen/block_context.go
```diff
@@ -13,52 +13,216 @@
...
+				if !isInRange(lowerBound, currentValue, upperBound) {
+					return ErrUnsatisfiable
+				}
+			} else {
+				if currentValue.IsUint64() {
+					if lowerBound <= value && value < upperBound {
+						return ErrUnsatisfiable
+					}
+				}
+			}
```
HerbertJordan, line 168: Is this the same?
```suggestion
			if inRange != isInRange(lowerBound, currentValue, upperBound) {
				return ErrUnsatisfiable
			}
```

go/ct/gen/block_context.go
```diff
@@ -68,13 +232,211 @@ func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision)
...
+	if b.blockNumberSolver == nil && b.rangeConstraints == nil && b.valueConstraint == nil {
+		return "true"
+	}
+
+	clauses := []string{}
+	if b.blockNumberSolver != nil {
+		if b.blockNumberSolver.IsSatisfiable() {
+			clauses = append(clauses, b.blockNumberSolver.Print("BlockNumber"))
+			// fmt.Sprintf("BlockNumber ∈ [%d..%d]",b.blockNumberSolver.min,b.blockNumberSolver.max,)
+
```
HerbertJordan, line 273: please remove

go/ct/gen/block_context.go
```diff
@@ -68,13 +232,211 @@ func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision)
...
+	}
+	len, err := GetBlockRangeLengthFor(upper)
+	if err != nil {
+		b.markUnsatisfiable()
+		return
+	}
+	if len == math.MaxUint64 {
+		max = math.MaxUint64
+	} else {
+		max += len - 1
```
HerbertJordan, line 403: nit: the overflow here may happen also for values of `len` smaller than `math.MaxUint64`

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+			want: "$a ∉ [BlockNumber-256..BlockNumber-1]",
+		},
+		"two-variables-in-range": {
+			setup: func(b *BlockContextGenerator) {
+				b.RestrictVariableToOneOfTheLast256Blocks("a")
+				b.RestrictVariableToOneOfTheLast256Blocks("b")
+			},
+			want: "$a ∈ [BlockNumber-256..BlockNumber-1] Λ $b ∈ [BlockNumber-256..BlockNumber-1]",
+		},
+		"one-variable-with-fixed-value": {
```
HerbertJordan, line 114: nit: (also for other places)
```suggestion
		"one-variable-with-fixed-offset": {
```

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+			b.SetBlockNumberOffsetValue("a", 44)
+			b.RestrictVariableToNoneOfTheLast256Blocks("a")
+		},
+		"conflicting-fixed-values-with-in-range": func(b *BlockContextGenerator, _ Assignment) {
+			b.SetBlockNumberOffsetValue("a", 400)
+			b.RestrictVariableToOneOfTheLast256Blocks("a")
+		},
+		"block-number-overflow": func(b *BlockContextGenerator, a Assignment) {
+			b.SetBlockNumberOffsetValue("a", 400)
+			a["a"] = common.NewU256(1, 500) // 2^64+400
```
HerbertJordan, line 366:
```suggestion
			a["a"] = common.NewU256(1, 500) // 2^64+500
```

go/ct/gen/block_context_test.go
HerbertJordan, line 387: seems to be the same as `conflicting-fixed-values` above

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+			test(generator, assignement)
+			res, err := generator.Generate(assignement, rand.New())
+			if err != ErrUnsatisfiable {
+				t.Errorf("expected unsatisfiable error, got %v with block number %d and assignment %v", err, res.BlockNumber, assignement)
+			}
+		})
 	}
+}
+
+// addOffset
```
HerbertJordan, line 406: nit: remove

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
 	}
 }
 
-func TestBlockContextGen_BlockNumberError(t *testing.T) {
-	rnd := rand.New(0)
-	blockContextGenerator := NewBlockContextGenerator()
-	_, err := blockContextGenerator.Generate(rnd, common.R99_UnknownNextRevision+1)
-	if err == nil {
-		t.Errorf("Failed to produce error with invalid revision.")
+func TestBlockContextGen_addWithOverfloCheck(t *testing.T) {
```
HerbertJordan, line 468:
```suggestion
func TestBlockContextGen_addWithOverflowCheck(t *testing.T) {
```

go/ct/gen/block_context_test.go
```diff
@@ -54,56 +56,476 @@ func TestBlockContextGen_Generate(t *testing.T) {
...
+
+	tests := map[string]struct {
+		base     uint64
+		offset   uint64
+		want     uint64
+		overflow bool
+	}{
+		"small-positive": {base: 44, offset: 44, want: 88, overflow: false},
+		"overflow":       {base: 0xffffffffffffffff, offset: 1, want: 0, overflow: true},
+	}
```
HerbertJordan, line 478:
```suggestion
		"max":              {base: 0xffffffffffffffff, offset: 0xffffffffffffffff, want: 0xfffffffffffffffe, overflow: true},
	}
```

HerbertJordan (changes requested):
I added two additional test cases that should point to problems referenced in the comments.

LuisPH3 (commented):
plz check if we need to generate blocks out of range, which are the current block or even future ids.

HerbertJordan (approved):

LuisPH3 (approved):

PR #491, opened 2024-06-03
Add new method to prime the Geth VM regarding revision

LuisPH3:
After Paris, revision is decided by the availability of Verkle timestamps for the subsequent revisions and the existence of a base hash value.

This procedure can be seen here: 
https://github.com/Fantom-foundation/tosca-go-ethereum/blob/master/params/config.go#L938

To influence the revisions in use by the CT, we prime the VM with artificial values.

This PR is needed for PUSH0, MCOPY and any other instruction with Revision >= Shanghai 

This PR was tested by including the commit into the PUSH0 branch and validating the existence of the operation in the jumptable.


go/vm/geth/geth.go
```diff
@@ -101,9 +101,18 @@ func createGethInterpreterContext(parameters vm.Parameters) (*geth.EVM, *geth.Co
 	// Set hard forks for chainconfig
 	chainConfig := *params.AllEthashProtocolChanges
 	chainConfig.ChainID = new(big.Int).SetBytes(context.ChainID[:])
-	chainConfig.IstanbulBlock = big.NewInt(int64(vm.R07_Istanbul) * 10)
-	chainConfig.BerlinBlock = big.NewInt(int64(vm.R09_Berlin) * 10)
-	chainConfig.LondonBlock = big.NewInt(int64(vm.R10_London) * 10)
+	chainConfig.IstanbulBlock = big.NewInt(int64(vm.R07_Istanbul) * 1000)
```
HerbertJordan, line 104: Can you update this to use [`GetForkBlock`](https://github.com/Fantom-foundation/Tosca/blob/84cbb23429c9275a1d890a1e3cd8615d2f448cdc/go/ct/common/revisions.go#L105) instead of hard-coded constants?

go/ct/rlz/conditions.go
```diff
@@ -328,8 +328,9 @@ func IsRevision(revision Revision) Condition {
 	return RevisionBounds(revision, revision)
 }
 
+// Any revision covered by the CT, excludes future revisions.
```
HerbertJordan, line 331: nit: comments in Go by convention always start with the function or type documented. In this case
```suggestion
// AnyKnownRevision restricts the revision to any revision covered by the CT specification.
```

go/vm/geth/geth.go
```diff
@@ -101,9 +101,18 @@ func createGethInterpreterContext(parameters vm.Parameters) (*geth.EVM, *geth.Co
...
+	chainConfig.BerlinBlock = big.NewInt(int64(vm.R09_Berlin) * 1000)
+	chainConfig.LondonBlock = big.NewInt(int64(vm.R10_London) * 1000)
+	if parameters.Revision >= vm.R11_Paris {
+		chainConfig.MergeNetsplitBlock = big.NewInt(int64(vm.R11_Paris) * 1000)
+	}
+	if parameters.Revision >= vm.R12_Shanghai {
+		chainConfig.ShanghaiTime = verkleTime(vm.R12_Shanghai)
```
HerbertJordan, line 111: Like for the fork block, we should define a `GetForkTime(revision Revision)` in the CT common directory and then use this function to determine the time to be set here.

go/vm/geth/geth.go
```diff
@@ -117,7 +126,9 @@ func createGethInterpreterContext(parameters vm.Parameters) (*geth.EVM, *geth.Co
 
 	// Create empty block context based on block number
 	blockCtx := geth.BlockContext{
-		BlockNumber: big.NewInt(int64(parameters.Revision)*10 + 2),
+		// Note: after Paris,the block number no longer indicates the revision
+		// but the existence of the random field in the block context.
```
HerbertJordan, line 130: I think this comment can be removed. The `BlockNumber` in the context serves its independent purpose (to provide the input for the `NUMBER` op-code), and I do not think the expectation is that this field is influencing the revision as this is handled by other code in this function.

go/vm/geth/geth.go
```diff
@@ -126,6 +137,11 @@ func createGethInterpreterContext(parameters vm.Parameters) (*geth.EVM, *geth.Co
 		Transfer:    transferFunc,
 		CanTransfer: canTransferFunc,
 	}
+
+	if parameters.Revision >= vm.R11_Paris {
+		blockCtx.Random = &common.Hash{}
```
HerbertJordan, line 142: This should be commented with something like:
```suggestion
                // Setting the random signals to geth that a post-merge (Paris) revision should be utilized.
		blockCtx.Random = &common.Hash{}
```

HerbertJordan, line 142: This field may be the input for the `PREVRANDO` operation in geth. If so, `PREVRANDO` should fail in Paris.

To fix, set the value to the same value as the `Difficulty` in line 158.

HerbertJordan (changes requested):

go/vm/geth/geth.go
LuisPH3: I checked the vm, there is a new function for PREVRANDAO which fetches the value from this new context field. 
I update them to use the same value.

go/vm/geth/geth.go
LuisPH3: So far this was the only place forwarding any information mapping our revision enum into geth.
(only use of parameters. Revision in the function).
Nevertheless, the new comment reflects the intent better.

go/vm/geth/geth.go
LuisPH3: This was a little too much work and is is still not 100% how I would like it.
A clean implementation should get rid first of one the vm.Revision and ct.Revision types

The proposed solution factors this code out and reuses an already existing function that was meat to do this. (although it was never used before) 
Please let me know if we should keep these changes or we revert to the hardcoded values.

go/ct/common/revisions.go
```diff
@@ -122,6 +122,23 @@ func GetForkBlock(revision Revision) (uint64, error) {
 	return 0, fmt.Errorf("unknown revision: %v", revision)
 }
 
+func GetVerkleTime(revision Revision) (uint64, error) {
```
HerbertJordan, line 125: Please add documentation. What is it intended for and how to use it.

HerbertJordan, line 125: Rename to `GetForkTime`

go/vm/geth/geth.go
```diff
@@ -95,15 +96,92 @@ func (m *gethVm) Run(parameters vm.Parameters) (vm.Result, error) {
...
+	case vm.R12_Shanghai:
+		return ct.R12_Shanghai
+	case vm.R13_Cancun:
+		return ct.R13_Cancun
+
+	}
+	panic(fmt.Sprintf("Unknown revision: %v", revision))
+}
+
+// GetChainConfig returns a chain config for the given chain ID and target revision.
```
HerbertJordan, line 119: nit: the function name in the comment does not match the function name

go/vm/geth/geth.go
```diff
@@ -95,15 +96,92 @@ func (m *gethVm) Run(parameters vm.Parameters) (vm.Result, error) {
...
+	}
+	londonBlock, err := ct.GetForkBlock(ct.R10_London)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get London fork block: %v", err))
+	}
+	parisBlock, err := ct.GetForkBlock(ct.R11_Paris)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get Paris fork block: %v", err))
+	}
+	shanghaiTime, err := ct.GetForkBlock(ct.R12_Shanghai)
```
HerbertJordan, line 140: Should this be `GetForkTime` instead of `GetForkBlock`?

HerbertJordan (dismissed):

go/vm/geth/geth.go
```diff
@@ -95,15 +96,92 @@ func (m *gethVm) Run(parameters vm.Parameters) (vm.Result, error) {
...
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get Shanghai verkle time: %v", err))
```
HerbertJordan, line 142:
```suggestion
		panic(fmt.Sprintf("Failed to get Shanghai fork time: %v", err))
```

go/vm/geth/geth.go
```diff
@@ -95,15 +96,92 @@ func (m *gethVm) Run(parameters vm.Parameters) (vm.Result, error) {
...
+	}
+	cancunTime, err := ct.GetForkBlock(ct.R13_Cancun)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get Cancun verkle time: %v", err))
```
HerbertJordan, line 146:
```suggestion
		panic(fmt.Sprintf("Failed to get Cancun fork time: %v", err))
```

simonlechner (dismissed):
LGTM

go/ct/common/revisions.go
```diff
@@ -122,6 +122,23 @@ func GetForkBlock(revision Revision) (uint64, error) {
...
 
+func GetForkTimestamp(revision Revision) (uint64, error) {
+	switch revision {
+	case R07_Istanbul:
+	case R09_Berlin:
+	case R10_London:
+	case R11_Paris:
+		return 0, fmt.Errorf("revision: %v is too low for timestamps", revision)
+	case R12_Shanghai:
+		return 4000, nil
```
facuMH, line 133: should we just for the sake of consistency use the same numbers as for `GetForkBlock` ?

go/vm/geth/geth.go
```diff
@@ -95,15 +96,92 @@ func (m *gethVm) Run(parameters vm.Parameters) (vm.Result, error) {
...
+		return ct.R12_Shanghai
+	case vm.R13_Cancun:
+		return ct.R13_Cancun
+
+	}
+	panic(fmt.Sprintf("Unknown revision: %v", revision))
+}
+
+// GetChainConfig returns a chain config for the given chain ID and target revision.
+// The baseline config is used as a starting point, so that any prefilled configuration from go-ethereum params/config.go can be used.
```
facuMH, line 120:
```suggestion
// The baseline config is used as a starting point, so that any prefilled configuration from go-ethereum:params/config.go can be used.
```

go/vm/geth/geth.go
```diff
@@ -95,15 +96,92 @@ func (m *gethVm) Run(parameters vm.Parameters) (vm.Result, error) {
...
+}
+
+// makeChainConfig returns a chain config for the given chain ID and target revision.
+// The baseline config is used as a starting point, so that any prefilled configuration from go-ethereum params/config.go can be used.
+// chainId needs to be prefilled as it may be accessed with the opcode CHAINID.
+// the fork-blocks and the fork-times are set to the respective values for the given revision.
+func makeChainConfig(baseline params.ChainConfig, chainId *big.Int, targetRevision ct.Revision) params.ChainConfig {
+	istanbulBlock, err := ct.GetForkBlock(ct.R07_Istanbul)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get Istanbul fork block: %v", err))
```
facuMH, line 126: nit: maybe a lambda func could be used to reproduce the same error msg just changing the revision.

facuMH (approved):
just a couple of nit/suggestions, feel free to disregard them, looks good otheriwse.

go/ct/common/revisions.go
LuisPH3: There are the same values as ForkBlock, I still do not know what are the implications here. 
It may be that we need to tune the timestamps later on. For now, I'll leave them as they are.

go/vm/geth/geth.go
LuisPH3: Yes, I agree, but I would rather leave this code for now, and clean it up when we can remove the errors from the getForkBlock and the two enums for revisions. 
Then, it is going to look much better.

facuMH (approved):

go/ct/common/revisions.go
```diff
@@ -122,6 +122,23 @@ func GetForkBlock(revision Revision) (uint64, error) {
 	return 0, fmt.Errorf("unknown revision: %v", revision)
 }
 
+func GetForkTime(revision Revision) (uint64, error) {
```
HerbertJordan, line 125: Please add documentation.

HerbertJordan (approved):

simonlechner (approved):

LuisPH3:
Unfortunately, while writing documentation, I had second thoughts about the timestamp generation:

evm revision is decided based on:
- Block number (for revisions < Shanghai
- existence of Random + BlockCtx.time 

The arbitrary timestamps generated before work because the random generator may create numbers much greater than the few thousands hard-coded. Nevertheless, this needs to be accurate, so that the CT can pinpoint revisions precisely. 

Sorry about this, I overlooked before, A new commit is coming, it is self-contained.

- [ ] run CT
- [ ] run CT in rebased push0 branch to verify latter revisions.

facuMH (approved):

go/ct/common/revisions.go
```diff
@@ -122,21 +122,28 @@ func GetForkBlock(revision Revision) (uint64, error) {
 	return 0, fmt.Errorf("unknown revision: %v", revision)
 }
 
-func GetForkTime(revision Revision) (uint64, error) {
+// GetForkTime returns the revison fork timestamp.
+// It is intended to provide input for test state generators to produce consistent
+// fork timestamps and code revisions, as well as for adapters between the CT framework
+// and EVM interpreters to support the state conversion.
+// This function will never fail, as it may be required to generate a timestamp for an guture revision.
```
HerbertJordan, line 129:
```suggestion
// This function will never fail, as it may be required to generate a timestamp for an future revision.
```

go/ct/gen/block_context.go
```diff
@@ -54,7 +54,10 @@ func (*BlockContextGenerator) Generate(rnd *rand.Rand, revision common.Revision)
 	gasPrice := common.RandU256(rnd)
 
 	prevRandao := common.RandU256(rnd)
-	timestamp := rnd.Uint64()
+
+	time := common.GetForkTime(revision)
+	nextTime := common.GetForkTime(revision + 1)
+	timestamp := rnd.Uint64n(nextTime-time) + time
```
HerbertJordan, line 60: Nice catch!

go/ct/common/revisions.go
```diff
@@ -122,21 +122,28 @@ func GetForkBlock(revision Revision) (uint64, error) {
...
+func GetForkTime(revision Revision) uint64 {
```
HerbertJordan, line 130: Nice to see that you managed to eliminate the error.

HerbertJordan (approved):
These are nice improvements, thanks.

facuMH (approved):

simonlechner (approved):

Issue #489, opened 2024-06-03
Limit init code size


simonlechner:
Starting with Shanghai a CREATE/CREATE2 call fails if the initcode size is greater than the chain's maximum initcode size.

PR #498, opened 2024-06-05
LFVM: Add init code size check

simonlechner:
Starting with Shanghai, the init code of a create call can no longer have arbitrary size. These changes are added to the LFVM create/create2 implementations.

Contributes to #489.

### DIFF
--- a/go/vm/lfvm/instructions.go
+++ b/go/vm/lfvm/instructions.go
@@ -800,13 +800,28 @@ func opExtcodehash(c *context) {
 	}
 }
 
-// Constants for CREATE / CREATE2 opcodes
-const (
-	MaxCodeSize     = 24576           // Maximum bytecode to permit for a contract
-	MaxInitCodeSize = 2 * MaxCodeSize // Maximum initcode to permit in a creation transaction and create instructions
+func checkInitCodeSize(c *context, size *uint256.Int) bool {
+	const (
+		MaxCodeSize     = 24576           // Maximum bytecode to permit for a contract
+		MaxInitCodeSize = 2 * MaxCodeSize // Maximum initcode to permit in a creation transaction and create instructions
+		InitCodeWordGas = 2               // Once per word of the init code when creating a contract.
+	)
 
-	InitCodeWordGas = 2 // Once per word of the init code when creating a contract.
-)
+	if !c.isShanghai() {
+		return true
+	}
+	if !size.IsUint64() || size.Uint64() > MaxInitCodeSize {
+		c.UseGas(c.gas)
+		c.status = MAX_INIT_CODE_SIZE_EXCEEDED
+		return false
+	}
+	if !c.UseGas(vm.Gas(InitCodeWordGas * ((size.Uint64() + 31) / 32))) {
+		c.status = OUT_OF_GAS
+		return false
+	}
+
+	return true
+}
 
 func opCreate(c *context) {
 	var (
@@ -823,16 +838,8 @@ func opCreate(c *context) {
 		return
 	}
 
-	if c.revision >= vm.R12_Shanghai {
-		if !size.IsUint64() || size.Uint64() > MaxInitCodeSize {
-			c.UseGas(c.gas)
-			c.status = MAX_INIT_CODE_SIZE_EXCEEDED
-			return
-		}
-		if !c.UseGas(vm.Gas(InitCodeWordGas * ((size.Uint64() + 31) / 32))) {
-			c.status = OUT_OF_GAS
-			return
-		}
+	if !checkInitCodeSize(c, size) {
+		return
 	}
 
 	if !value.IsZero() {
@@ -895,16 +902,8 @@ func opCreate2(c *context) {
 		return
 	}
 
-	if c.revision >= vm.R12_Shanghai {
-		if !size.IsUint64() || size.Uint64() > MaxInitCodeSize {
-			c.UseGas(c.gas)
-			c.status = MAX_INIT_CODE_SIZE_EXCEEDED
-			return
-		}
-		if !c.UseGas(vm.Gas(InitCodeWordGas * ((size.Uint64() + 31) / 32))) {
-			c.status = OUT_OF_GAS
-			return
-		}
+	if !checkInitCodeSize(c, size) {
+		return
 	}
 
 	// Charge for the code size
--- a/go/vm/lfvm/instructions.go
+++ b/go/vm/lfvm/instructions.go
@@ -800,6 +800,14 @@ func opExtcodehash(c *context) {
 	}
 }
 
+// Constants for CREATE / CREATE2 opcodes
+const (
+	MaxCodeSize     = 24576           // Maximum bytecode to permit for a contract
+	MaxInitCodeSize = 2 * MaxCodeSize // Maximum initcode to permit in a creation transaction and create instructions
+
+	InitCodeWordGas = 2 // Once per word of the init code when creating a contract.
+)
+
 func opCreate(c *context) {
 	var (
 		value  = c.stack.pop()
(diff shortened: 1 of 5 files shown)


go/vm/lfvm/instructions.go
```diff
@@ -764,6 +772,18 @@ func opCreate(c *context) {
 		return
 	}
 
+	if c.revision >= vm.R12_Shanghai {
```
facuMH, line 775: why not use `c.isShanghai()` ?

go/vm/lfvm/instructions.go
```diff
@@ -824,6 +844,18 @@ func opCreate2(c *context) {
...
```
facuMH, line 847: why not use `c.isShanghai()` ?

go/vm/lfvm/instructions.go
facuMH, line 775: maybe you could factor out the check into a helper function and then just call the check in the 2 ops, instead of having the same check twice ?

facuMH (changes requested):

go/vm/lfvm/instructions.go
```diff
@@ -764,6 +772,18 @@ func opCreate(c *context) {
...
+		if !size.IsUint64() || size.Uint64() > MaxInitCodeSize {
+			c.UseGas(c.gas)
+			c.status = MAX_INIT_CODE_SIZE_EXCEEDED
+			return
+		}
+		if !c.UseGas(vm.Gas(InitCodeWordGas * ((size.Uint64() + 31) / 32))) {
```
HerbertJordan, line 781: The gas computation could overflow. This should be checked for.

HerbertJordan (changes requested):
Although it is not an established practice in the LFVM directory, we should add unit tests for changes. It should not be too complicated. An example is given by #352 introducing [this](https://github.com/Fantom-foundation/Tosca/blob/f8497d0604fc8c8a8c604760e12d8291ef590902/go/vm/lfvm/instructions_test.go#L209) test also for the call expression.

go/vm/lfvm/instructions.go
LuisPH3: +1

LuisPH3 (approved):

go/vm/lfvm/instructions.go
```diff
@@ -749,6 +749,28 @@ func opExtcodehash(c *context) {
 	}
 }
 
+func createShanghaiInitCodeSizeCheck(c *context, size *uint256.Int) bool {
+	const (
+		MaxCodeSize     = 24576           // Maximum bytecode to permit for a contract
+		MaxInitCodeSize = 2 * MaxCodeSize // Maximum initcode to permit in a creation transaction and create instructions
+
```
facuMH, line 756: unnecessary empty line ?

go/vm/lfvm/instructions_test.go
```diff
@@ -285,3 +285,59 @@ func TestCreateChecksBalance(t *testing.T) {
...
+		context:  runContext,
+		stack:    NewStack(),
+		memory:   NewMemory(),
+		gas:      50000,
+		revision: vm.R12_Shanghai,
+	}
+
+	// Prepare stack arguments.
+	ctxt.stack.stack_ptr = 3
+	ctxt.stack.data[0].Set(uint256.NewInt(49153)) // < size of the init code
```
facuMH, line 308: maybe add a comment explaining where does this number comes from ? or use `MaxInitCodeSize+1`

facuMH (approved):

LuisPH3 (approved):

go/vm/lfvm/instructions.go
```diff
@@ -749,6 +749,28 @@ func opExtcodehash(c *context) {
 	}
 }
 
+func createShanghaiInitCodeSizeCheck(c *context, size *uint256.Int) bool {
```
HerbertJordan, line 752: This reads like the function would create a check ... rename to `checkInitCodeSize`?

go/vm/lfvm/instructions.go
```diff
@@ -749,6 +749,28 @@ func opExtcodehash(c *context) {
...
+	const (
+		MaxCodeSize     = 24576           // Maximum bytecode to permit for a contract
+		MaxInitCodeSize = 2 * MaxCodeSize // Maximum initcode to permit in a creation transaction and create instructions
+
+		InitCodeWordGas = 2 // Once per word of the init code when creating a contract.
+	)
+
+	if c.isShanghai() {
```
HerbertJordan, line 760: nit: as it is in a function now, an early exit reduces the nesting levels.

HerbertJordan (approved):

go/vm/lfvm/instructions_test.go
HerbertJordan, line 1: As stand-alone unit test for the LFVM those tests are a bit limited. 

They only check a single situation -- the one where the length is 1 byte above the expected limit. They do not check positive cases or the actual boundary.

One can claim that the CT will test those boundaries thoroughly, but I would argue that we should not depend to much on the CT. Unit tests should be the first line of defense against errors.

So maybe you can parameterize those tests and check different init code size lengths. Some that are accepted, some that are not. Also, check that this limit is only enforced by Shanghai and later.

simonlechner: I refactored the unit tests to check different revisions and input sizes.

go/vm/lfvm/instructions_test.go
```diff
@@ -465,3 +466,116 @@ func TestMCopy(t *testing.T) {
...
 	}
 }
+
+func TestCreateShanghaiInitCodeSize(t *testing.T) {
+	maxInitCodeSize := uint64(49152)
+	tests := []struct {
+		revision       vm.Revision
+		init_code_size uint64
+		expected       Status
+	}{
```
LuisPH3, line 476: If the change is introduced with Shanghai, I would have expected a test pre-shanghai, to make sure that the behavior has not been altered for older revisions.

simonlechner: There are several test cases with Paris revision checking the difference between old revisions and Shanghai which introduced changes. Is this not sufficient?

go/vm/lfvm/instructions_test.go
```diff
@@ -465,3 +466,116 @@ func TestMCopy(t *testing.T) {
...
+		{vm.R11_Paris, 0, RUNNING},
+		{vm.R11_Paris, 1, RUNNING},
+		{vm.R11_Paris, 2000, RUNNING},
+		{vm.R11_Paris, maxInitCodeSize - 1, RUNNING},
+		{vm.R11_Paris, maxInitCodeSize, RUNNING},
+		{vm.R11_Paris, maxInitCodeSize + 1, RUNNING},
+		{vm.R11_Paris, 100000, RUNNING},
+		{vm.R11_Paris, math.MaxUint64, RUNNING},
```
HerbertJordan, line 484: It is surprising that this test is not failing -- either with an out-of-gas error or a bad allocation crash.

I looked into it, and it looks like `CREATE` is not growing the memory to fit its input requirements. Is this something that is missing? If so, why did this not show up in the CT?

go/vm/lfvm/instructions_test.go
```diff
@@ -465,3 +466,116 @@ func TestMCopy(t *testing.T) {
...
+		// gas cost from evm.codes
+		{vm.R11_Paris, 0, 32000 - static_gas},
+		{vm.R11_Paris, 1, 32003 - static_gas},
+		{vm.R11_Paris, 2000, 32196 - static_gas},
+		{vm.R11_Paris, 49152, 41216 - static_gas},
+
+		{vm.R12_Shanghai, 0, 32000 - static_gas},
+		{vm.R12_Shanghai, 1, 32005 - static_gas},
+		{vm.R12_Shanghai, 2000, 32322 - static_gas},
+		{vm.R12_Shanghai, 49152, 44288 - static_gas},
```
HerbertJordan, line 545: nit: instead of hard-coded gas prices it would be nice to have a function for those to see how those values are derived.

HerbertJordan (dismissed):
The PR looks good, but I think it may point to an unrelated implementation problem in the create operators. Please see the comments.

go/vm/lfvm/instructions_test.go
LuisPH3: my bad, I confused the revision order again.

LuisPH3 (approved):

LuisPH3 (approved):

go/vm/lfvm/instructions_test.go
simonlechner: This should be investigated and solved in #524.
I added this as a comment in the code since it should not be part of this PR.

HerbertJordan (approved):

facuMH (approved):

PR #499, opened 2024-06-05
evmzero: Limit init code size

simonlechner:
Starting with Shanghai, the init code of a create call can no longer have arbitrary size. These changes are added to the evmzero create/create2 implementations.

Contributes to #489.

### DIFF
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -1519,17 +1519,19 @@ struct CreateImpl {
       return {.dynamic_gas_costs = initial_gas - gas};
 
     if (ctx.revision >= EVMC_SHANGHAI) {
-      const int64_t max_code_size = 24576;
-      const int64_t max_init_code_size = 2 * max_code_size;
-      const int64_t init_code_word_cost = 2;
+      const int64_t kMaxCodeSize = 24576;  // introduced with EIP-3860
+      const int64_t kMaxInitCodeSize = 2 * kMaxCodeSize;
+      const int64_t kInitCodeWordCost = 2;
 
-      if (init_code_size > max_init_code_size) {
+      if (init_code_size > kMaxInitCodeSize) {
         return {
             .state = RunState::kErrorInitCodeSizeExceeded,
             .dynamic_gas_costs = initial_gas - gas,
         };
       }
-      if (gas -= init_code_word_cost * static_cast<int64_t>(((init_code_size + 31) / 32)), gas < 0) {
+
+      const int64_t init_code_cost = kInitCodeWordCost * static_cast<int64_t>(((init_code_size + 31) / 32));
+      if (gas -= init_code_cost, gas < 0) {
         return {.dynamic_gas_costs = initial_gas - gas};
       }
     }
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -56,6 +56,8 @@ const char* ToString(RunState state) {
       return "ErrorCreate";
     case RunState::kErrorStaticCall:
       return "ErrorStaticCall";
+    case RunState::kErrorInitCodeSizeExceeded:
+      return "ErrorInitCodeSizeExceeded";
   }
   return "UNKNOWN_STATE";
 }
@@ -1515,6 +1517,22 @@ struct CreateImpl {
         ctx.MemoryExpansionCost(init_code_offset_u256, init_code_size_u256);
     if (gas -= mem_cost; gas < 0) [[unlikely]]
       return {.dynamic_gas_costs = initial_gas - gas};
+
+    if (ctx.revision >= EVMC_SHANGHAI) {
+      const int64_t max_code_size = 24576;
+      const int64_t max_init_code_size = 2 * max_code_size;
+      const int64_t init_code_word_cost = 2;
+
+      if (init_code_size > max_init_code_size) {
+        return {
+            .state = RunState::kErrorInitCodeSizeExceeded,
+            .dynamic_gas_costs = initial_gas - gas,
+        };
+      }
+      if (gas -= init_code_word_cost * static_cast<int64_t>(((init_code_size + 31) / 32)), gas < 0) {
+        return {.dynamic_gas_costs = initial_gas - gas};
+      }
+    }
 
     if constexpr (Op == op::CREATE2) {
       const int64_t minimum_word_size = static_cast<int64_t>((init_code_size + 31) / 32);
--- a/cpp/vm/evmzero/interpreter_test.cc
+++ b/cpp/vm/evmzero/interpreter_test.cc
@@ -5010,6 +5010,72 @@ TEST(InterpreterTest, CREATE_MemoryIsGrownAlsoWithInsufficientBalance) {
   });
 }
 
+TEST(InterpreterTest, CREATE_ShanghaiLimitInitCodeSize) {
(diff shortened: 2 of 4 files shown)


facuMH (dismissed):

cpp/vm/evmzero/interpreter.cc
```diff
@@ -1487,6 +1489,22 @@ struct CreateImpl {
...
+      const int64_t max_init_code_size = 2 * max_code_size;
+      const int64_t init_code_word_cost = 2;
+
+      if (init_code_size > max_init_code_size) {
+        return {
+            .state = RunState::kErrorInitCodeSizeExceeded,
+            .dynamic_gas_costs = initial_gas - gas,
+        };
+      }
+      if (gas -= init_code_word_cost * static_cast<int64_t>(((init_code_size + 31) / 32)), gas < 0) {
```
HerbertJordan, line 1503: nit: i think it would help readability to decompose this into two steps:
```suggestion
      const int64_t init_code_costs = init_code_word_cost * static_cast<int64_t>(((init_code_size + 31) / 32));
      if (gas -= init_code_costs, gas < 0) {
```

cpp/vm/evmzero/interpreter.cc
```diff
@@ -1487,6 +1489,22 @@ struct CreateImpl {
     if (gas -= mem_cost; gas < 0) [[unlikely]]
       return {.dynamic_gas_costs = initial_gas - gas};
 
+    if (ctx.revision >= EVMC_SHANGHAI) {
+      const int64_t max_code_size = 24576;
+      const int64_t max_init_code_size = 2 * max_code_size;
+      const int64_t init_code_word_cost = 2;
```
HerbertJordan, line 1495: When starting the `evmzero` the agreement was to follow Google's C++ style guide. Not sure to what extend that was done, but one part of it would ask constants to be prefixed with a `k` followed by mixed-case (see [guidlines](https://google.github.io/styleguide/cppguide.html#Constant_Names)). So those constants should be named `kMaxCodeSize`, `kMaxInitCodeSize`, and `kInitCodeWordCost`.

HerbertJordan (changes requested):
Please add unit tests.

cpp/vm/evmzero/interpreter.cc
LuisPH3: Why doesn't clang-format deal with this ?
IMO, we should not need to manually deal with these details. 
I offer to write a patch about that.

LuisPH3 (approved):

LuisPH3 (approved):

cpp/vm/evmzero/interpreter.cc
```diff
@@ -1487,6 +1489,24 @@ struct CreateImpl {
     if (gas -= mem_cost; gas < 0) [[unlikely]]
       return {.dynamic_gas_costs = initial_gas - gas};
 
+    if (ctx.revision >= EVMC_SHANGHAI) {
+      const int64_t kMaxCodeSize = 24576;
```
facuMH, line 1493: maybe add a comment explaining where does this number comes from ?

cpp/vm/evmzero/interpreter.cc
```diff
@@ -1487,6 +1489,24 @@ struct CreateImpl {
...
+      const int64_t kMaxInitCodeSize = 2 * kMaxCodeSize;
+      const int64_t kInitCodeWordCost = 2;
+
+      if (init_code_size > kMaxInitCodeSize) {
+        return {
+            .state = RunState::kErrorInitCodeSizeExceeded,
+            .dynamic_gas_costs = initial_gas - gas,
```
facuMH, line 1500: should the cost still be updated on a failed case ?

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -5006,6 +5006,36 @@ TEST(InterpreterTest, CREATE_MemoryIsGrownAlsoWithInsufficientBalance) {
   });
 }
 
+TEST(InterpreterTest, CREATE_ShanghaiLimitInitCodeSize) {
```
facuMH, line 5009: does it make sense to add a test for previous revisions ?

facuMH (approved):

cpp/vm/evmzero/interpreter.cc
HerbertJordan: @LuisPH3 do you think clang format is powerful enough to identify such cases?

HerbertJordan (approved):

cpp/vm/evmzero/interpreter.cc
LuisPH3: I was thinking something in this direction:
https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html
(sorry, this is part of clang-tidy, not format)

cpp/vm/evmzero/interpreter.cc
simonlechner: I don't think it makes sense to write a huge comment why something has been introduced and I also don't want to write a comment for a constant definition that it is from the ethereum standard.

cpp/vm/evmzero/interpreter.cc
simonlechner: This is also performed in all the other operations, all gas has to be consumed in a failed case.

cpp/vm/evmzero/interpreter.cc
LuisPH3: I think a reference to the source of this value would be in order. How do I find the definition of this value? 
just the eip number would be enough.

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -5006,6 +5006,36 @@ TEST(InterpreterTest, CREATE_MemoryIsGrownAlsoWithInsufficientBalance) {
...
+          },
+      .revision = EVMC_SHANGHAI,
+  });
+}
+
+TEST(InterpreterTest, CREATE_ShanghaiCodeDepositCost) {
+  RunInterpreterTest({
+      .code = {op::CREATE},
+      .state_after = RunState::kErrorGas,
+      .gas_before = 42000,  // sufficient amount of gas before Shanghai
```
LuisPH3, line 5028: Test is called shanghai, is this before cancun? 
previous test tests shanghai as well. Is this correct? I am not familiar which exact version did introduce this change.

LuisPH3 (changes requested):
Sorry to ask for comments again, but I saw something fishy in the unit testing, please check that it is correct.

Both tests test Shanghai only, should it be one new behavior and another one old behavior?  
Comment is a little misleading as well.

cpp/vm/evmzero/interpreter_test.cc
simonlechner: This change has been introduced in Shanghai which if before Cancun.

One test is checking that the create call fails due to exceeded init code size and the other one due to insufficient gas. With this change the dynamic cost of create has been increased. The comment documents that this has been a sufficient amount of gas for this call before Shanghai. But I will rework the tests since they don't seem to be easily understandable.

cpp/vm/evmzero/interpreter.cc
simonlechner: I added the EIP number as a comment.

HerbertJordan (approved):

facuMH (approved):

LuisPH3 (approved):

PR #500, opened 2024-06-05
CT: Add create init code limit

### LINKED ISSUE #489: Limit init code size

simonlechner:
Starting with Shanghai, the init code of a create call can no longer have arbitrary size. These changes are added to the create/create2 effect implementation.

Fixes #489, can only be merged after #498 and #499 to ensure correct CT execution.

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -1584,6 +1584,21 @@ func createEffect(s *st.State, callKind vm.CallKind) {
 	memExpCost, offset, size := s.Memory.ExpansionCosts(offsetU256, sizeU256)
 	dynamicGas := memExpCost
 
+	if s.Revision >= R12_Shanghai {
+		const (
+			MaxCodeSize     = 24576           // Maximum bytecode to permit for a contract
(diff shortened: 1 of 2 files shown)


facuMH (dismissed):

HerbertJordan (approved):

LuisPH3 (changes requested):
please rebase with main and remove revision guards in the Create instructions

HerbertJordan (approved):

go/ct/spc/specification.go
```diff
@@ -1557,7 +1554,6 @@ func getAllRules() []Rule {
 		pushes:    1,
 		conditions: []Condition{
 			Eq(ReadOnly(), false),
-			RevisionBounds(MinRevision, R11_Paris),
```
facuMH, line 1560: why remove them instead of making them `AnyKnownRevision()` ?

facuMH (approved):

go/ct/spc/specification.go
simonlechner: This constrained is already added by the `rulesFor` function.

LuisPH3 (approved):

Issue #467, opened 2024-05-21
Implement MCOPY instruction


HerbertJordan:
As part of the Cancun revision, the MCOPY instruction was added ([EIP-5656](https://eips.ethereum.org/EIPS/eip-5656)).

To support this instruction in Tosca, the following steps are required:
- update geth dependency to a version that covers Cancun (see #394)
- add MCOPY instruction support to LFVM + unit tests
- add MCOPY instruction support to the evmzero + unit tests
- add MCOPY specification to the CT spec

PR #501, opened 2024-06-05
MCOPY for evmzero

LuisPH3:
This is tested with a preliminary CT branch that can run both geth and evmzero mcopy.

Part of #467


facuMH (approved):

cpp/vm/evmzero/memory.h
```diff
@@ -70,6 +70,14 @@ class Memory {
     std::copy_n(memory_.data() + memory_offset, buffer.size(), buffer.data());
   }
 
+  // Reads from memory, writes into memory, grows memory if necessary.
+  void MemCopy(uint64_t src_offset, uint64_t dest_offset, uint64_t size) {
+    Grow(src_offset, size);
+    Grow(dest_offset, size);
+    // copy with overlap guarantees
+    std::memmove(memory_.data() + dest_offset, memory_.data() + src_offset, size);
```
HerbertJordan, line 78: Can you add unit test for this function into [memory_test.cc](https://github.com/Fantom-foundation/Tosca/blob/main/cpp/vm/evmzero/memory_test.cc)? 
In particular it should check that overlapping memory regions are handled correctly.

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -5198,6 +5198,123 @@ TEST(InterpreterTest, Step_Single_Memory) {
       1);
 }
 
+///////////////////////////////////////////////////////////
+// Stepping interpreter - JUMP
```
HerbertJordan, line 5202: This comment looks incorrect.

HerbertJordan (dismissed):

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -5198,6 +5198,124 @@ TEST(InterpreterTest, Step_Single_Memory) {
...
+
+TEST(InterpreterTest, MCOPY) {
+  RunInterpreterStepTest(
+      {
+          .code = {op::MCOPY},
+          .state_after = RunState::kRunning,
+          .pc_before = 0,
+          .pc_after = 1,
+          .gas_before = 1000,
+          .gas_after = 1000 - 6,
```
facuMH, line 5220: it seems error prone to use the same literals in all tests. maybe we could add a `const` in a scope just for these tests that is used for `pre_existing_gas`, `static_costg` and `dynamic_cost` (for the latter tests) ?

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -5198,6 +5198,124 @@ TEST(InterpreterTest, Step_Single_Memory) {
...
+
+TEST(InterpreterTest, MCOPY_PreCancun) {
+  RunInterpreterStepTest({
+      .code = {op::MCOPY},
+      .state_after = RunState::kErrorOpcode,
+      .revision = EVMC_SHANGHAI,
+  });
+}
+
+TEST(InterpreterTest, MCOPY) {
```
facuMH, line 5212: name suggestion `MCOPY_Regular`  ?

facuMH (dismissed):
looks good, couple of minor minor suggestions

cpp/vm/evmzero/interpreter_test.cc
LuisPH3: This is not the pattern in this file. 
Nevertheless, this file is going to suffer severe changes.

cpp/vm/evmzero/interpreter_test.cc
LuisPH3: In my experience, tests need to be explicit and global constants are vulnerable to be changed in the future, breaking tests or triggering way more work than needed.
The beauty of this code is that you do not need to scroll anywhere to see what it is supposed to be doing. A global constant defined in a 5000+ lines requires unnecessary scrolling.

cpp/vm/evmzero/memory_test.cc
```diff
@@ -214,5 +214,73 @@ TEST(MemoryTest, Equality) {
...
+    }
+  };
+
+  // Check pattern before copy in source range
+  check_pattern(memory.GetSpan(start_offset, size));
+
+  memory.MemCopy(0, start_offset, size);
+
+  // Check pattern after copy in dst range
+  check_pattern(memory.GetSpan(0, size));
```
HerbertJordan, line 282: The parameter `dest_offset` is ignored.

cpp/vm/evmzero/memory_test.cc
```diff
@@ -214,5 +214,73 @@ TEST(MemoryTest, Equality) {
...
+    int count = 0ul;
+    for (auto i = 0ul; i < effective_size; ++i) {
+      EXPECT_EQ(span[i], count++ % 37) << "at index " << i;
+    }
+  };
+
+  // Check pattern before copy in source range
+  check_pattern(memory.GetSpan(start_offset, size));
+
+  memory.MemCopy(0, start_offset, size);
```
HerbertJordan, line 279: This will potentially grow the memory, which should be checked.
Also, if so, the `effective_size` will be to short (just 1 byte long for some tests) and the `check_pattern` will not check all the copied data, right?

HerbertJordan (changes requested):

cpp/vm/evmzero/memory_test.cc
LuisPH3: Right, the remainder of the copied data should be zeroes (assuming zero initialized memory expansions) 
if we did an out of bounds copy, I will check for zeroes.

HerbertJordan (approved):

PR #502, opened 2024-06-06
CT: fix jump/i explicit test and evmzero adapter

facuMH:
This PR fixes the `ExplicitCases` test for `JUMP` and `JUMPI` instructions by forcing it to run with all the supported revisions, and updates the evmzero adapter to support Shangai as well.

This problem was hidden by the random number generator only generating revisions older than Shangai


LuisPH3:
Sorry about that, good catch.

LuisPH3 (approved):

simonlechner (approved):
Nice catch.

Issue #468, opened 2024-05-21
Implement BLOBHASH and BLOBBASEFEE instructions


HerbertJordan:
As part of the Cancun revision, the BLOB instructions BLOBHASH ([EIP-4844](https://eips.ethereum.org/EIPS/eip-4844)) and BLOBBASEFEE ([EIP-7516](https://eips.ethereum.org/EIPS/eip-7516)) have been added.

To support this instruction in Tosca, the following steps are required:
- update geth dependency to a version that covers Cancun (see #394)
- add BLOBHASH and BLOBBASEFEE instruction support to LFVM + unit tests
- add BLOBHASH and BLOBBASEFEE instruction support to the evmzero + unit tests
- add BLOBHASH and BLOBBASEFEE specification to the CT spec

facuMH:
@HerbertJordan please confirm the following, just to make sure I interpreted correctly the 2 new instructions.
- BLOBBASEFEE: simply a pushes on the stack a u256 to be interpreted as the already existing `BlobBaseFee`, analogous to the already existing `BaseFee` in the `BlockContext`.

- BLOBHASH: reads head of the stack as an `index` and check `index < len(tx.blob_versioned_hashes)` , and then push on the stack `tx.blob_versioned_hashes` which could be implemented as an array of hashes (I believe this should also be in the `BlockContext` since here is where we keep most of the transaction context info).

simonlechner:
This issue has been fixed in #532, #531, #528, 527, #523, #504 and #503.

PR #504, opened 2024-06-06
evmzero: BLOBBASEFEE

facuMH:
This PR adds the implementation of `BLOBBASEFEE` instruction in evmzero, three basic tests for it and updating the newest supported revision in the evmzero ct adaptor. 

This PR contributes to #468 

This PR needs the version of `evmc` to be upgraded to v11.


LuisPH3 (approved):

LuisPH3 (dismissed):

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -3198,6 +3198,47 @@ TEST(InterpreterTest, BASEFEE_PreRevision) {
...
+TEST(InterpreterTest, BLOBBASEFEE_OutOfGas) {
+  RunInterpreterTest({
+      .code = {op::BLOBBASEFEE},
+      .state_after = RunState::kErrorGas,
+      .gas_before = 1,
+      .revision = EVMC_CANCUN,
+  });
+}
+
+TEST(InterpreterTest, DISABLED_BLOBBASEFEE_StackOverflow) {}
```
HerbertJordan, line 3231: Why is this test `DISABLED_`?

HerbertJordan (approved):

simonlechner (dismissed):

cpp/vm/evmzero/interpreter_test.cc
LuisPH3: We discussed this yesterday: all StackOverflow are disabled and/or empty. #520

LuisPH3 (dismissed):

cpp/vm/evmzero/interpreter_test.cc
facuMH: yes. But we suspect this was done as "remainder" to implement them later on. I think with the helper function I introduce, and someone can take task #520  later on and implement all the missing tests.

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -156,6 +156,14 @@ void RunInterpreterStepTest(const InterpreterTestDescription& desc, int steps =
   EXPECT_EQ(ctx.return_data, desc.return_data);
 }
 
+Stack& fullStack() {
```
HerbertJordan, line 159: I would prefer this function to return a `Stack` instead of a mutable `Stack&` reference. In this case you also do not have to use a `static` instance (=global) in the implementation, which makes this function non-thread-safe.

HerbertJordan (changes requested):

cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -156,6 +156,14 @@ void RunInterpreterStepTest(const InterpreterTestDescription& desc, int steps =
...
+  static auto stack = Stack();
```
LuisPH3, line 160: Is this an optimization? to avoid generating new ones?
We will flash it clean with zeroes anyway. I guess a new instance would be safer.

cpp/vm/evmzero/interpreter_test.cc
facuMH: The stack overflow tests will be implemented in another PR since it requires a little more work as discussed with @LuisPH3

cpp/vm/evmzero/interpreter_test.cc
facuMH: The stack overflow tests will be implemented in another PR since it requires a little more work as discussed with @LuisPH3

HerbertJordan (approved):

PR #510, opened 2024-06-07
Fix Flaky unittest, allocate enough space for all needed opCodes

### LINKED ISSUE #507: Code generation introduces non-deterministic failure in ct/evm_test.go
The test has experienced intermittent failures in CI when compiling unrelated changes.
Initial investigation shows that the code generator attempts to introduce 2 op codes in a code segment of length 1, this
is a scenario that the algorithm should be able to solve. 

example of failure with unrelated code changes: https://github.com/Fantom-foundation/Tosca/actions/runs/9413810813/job/25931362205?pr=506

LuisPH3:
What does this solve?  
The unit test would have failed because of creating a code buffer of size 1 to store 2 instructions.

What does this not solve: 
It is still not possible to additionally set the Pc to a constant direction, this would require to generate the Pc at a different point in time and account for it. This is not needed right now.

fixes #507

### DIFF
--- a/go/ct/gen/code.go
+++ b/go/ct/gen/code.go
@@ -31,6 +31,9 @@ type CodeGenerator struct {
 	varOps               []varOpConstraint
 	varIsCodeConstraints []varIsCodeConstraint
 	varIsDataConstraints []varIsDataConstraint
+
+	// testing only
+	codeSize *int
 }
 
 type constOpConstraint struct {
@@ -131,12 +134,17 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 		minSize = 1
 	}
 
-	// We use an exponential distribution for the code size here since long codes
-	// extend the runtime but are expected to reveal limited extra code coverage.
-	const expectedSize float64 = 200
-	size := int(rnd.ExpFloat64()/(1/expectedSize)) + minSize
-	if size > st.MaxCodeSize {
-		size = st.MaxCodeSize
+	var size int
+	if g.codeSize != nil {
+		size = *g.codeSize + minSize
+	} else {
+		// We use an exponential distribution for the code size here since long codes
+		// extend the runtime but are expected to reveal limited extra code coverage.
+		const expectedSize float64 = 200
+		size = int(rnd.ExpFloat64()/(1/expectedSize)) + minSize
+		if size > st.MaxCodeSize {
+			size = st.MaxCodeSize
+		}
 	}
 
 	// Solve variable constraints. constOpConstraints are generated, the
--- a/go/ct/gen/code.go
+++ b/go/ct/gen/code.go
@@ -136,7 +136,10 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 
 	var size int
 	if g.codeSize != nil {
-		size = *g.codeSize + minSize
+		if *g.codeSize < minSize {
+			return nil, fmt.Errorf("%w, fixed code size %d is too small for constraints", ErrUnsatisfiable, *g.codeSize)
+		}
+		size = *g.codeSize
 	} else {
 		// We use an exponential distribution for the code size here since long codes
 		// extend the runtime but are expected to reveal limited extra code coverage.
--- a/go/ct/gen/code.go
+++ b/go/ct/gen/code.go
@@ -129,7 +129,7 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 	minSize = max(minSize, len(opCount))
 
 	// If there are any variables that need to be bound to code, there must be at
-	// least on instruction in the resulting code.
+	// least one instruction in the resulting code.
 	if minSize == 0 && len(g.varIsCodeConstraints) > 0 {
 		minSize = 1
 	}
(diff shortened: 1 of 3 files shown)


facuMH (dismissed):
thanks for making the test more stable. nice catch on edge case

go/ct/gen/code.go
```diff
@@ -114,6 +115,22 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
...
+	for _, constraint := range varOps {
+		opCount[constraint.op] = true
+	}
+	for _, constraint := range ops {
+		opCount[constraint.op] = true
+	}
+	minSize = max(minSize, len(opCount))
+
+	// If there are any variables that need to be bound to code, there must be at
+	// least on instruction in the resulting code.
```
HerbertJordan, line 129:
```suggestion
	// least one instruction in the resulting code.
```

go/ct/gen/code.go
```diff
@@ -114,6 +115,22 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
...
 
+	// Make enough space to host all the different opCodes used in condition.
+	opCount := make(map[OpCode]bool)
+	for _, constraint := range varOps {
+		opCount[constraint.op] = true
+	}
+	for _, constraint := range ops {
+		opCount[constraint.op] = true
+	}
+	minSize = max(minSize, len(opCount))
```
HerbertJordan, line 126: Please create a test case in `code_test.go` that fails without this modification and passes with this change.

go/ct/gen/code.go
```diff
@@ -114,6 +115,22 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
...
+
+	// If there are any variables that need to be bound to code, there must be at
+	// least on instruction in the resulting code.
+	if minSize == 0 && len(g.varIsCodeConstraints) > 0 {
+		minSize = 1
+	}
```
HerbertJordan, line 132: Please create a test case in `code_test.go` that fails without this modification and passes with this change.

HerbertJordan (changes requested):

go/ct/gen/code.go
LuisPH3: It is going to be pretty difficult to make deterministic, after all, the code size is a function of rand.

LuisPH3: image
I need some ideas about how to write a test that can call this function and deterministically test the corner cases.
I can only think about the injection of a random generator object mock, which I could influence.

HerbertJordan: It might help adding a upper bound for the code size as an additional constraint to the generator, narrowing down on the `st.MaxCodeSize`. This way, you can force the `size` to be 0, 1, or 2 in tests. Wdyt?

HerbertJordan: You might even have a fixed code-size constraint added just for testing. Something like
```
codeSize *int
```
as a simple field, that if set is used, and if `nil` is ignored.

go/ct/gen/code_test.go
```diff
@@ -354,6 +354,142 @@ func TestCodeGenerator_ClonesCanBeUsedToResetGenerator(t *testing.T) {
 	}
 }
 
+func TestCodeGenerator_CodeIsLargeEnoughForAllConditionOps(t *testing.T) {
+
+	type constOp struct {
+		p  int
```
HerbertJordan, line 360: Please do not use abbreviation for identifiers. It makes reading the code unnecessarily hard.

go/ct/gen/code_test.go
```diff
@@ -354,6 +354,142 @@ func TestCodeGenerator_ClonesCanBeUsedToResetGenerator(t *testing.T) {
...
+		op OpCode
+	}
+
+	type varOp struct {
+		v  string
+		op OpCode
+	}
+
+	tests := map[string]struct {
+		offset       int
```
HerbertJordan, line 370: The name `offset` does not convey any information about the purpose of this field.

Looking at the implementation below, it looks like it is the forced code length. So why not name it `codeLength` or `codeSize`?

go/ct/gen/code.go
```diff
@@ -114,12 +118,33 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
...
+
+	// If there are any variables that need to be bound to code, there must be at
+	// least one instruction in the resulting code.
+	if minSize == 0 && len(g.varIsCodeConstraints) > 0 {
+		minSize = 1
+	}
+
+	var size int
+	if g.codeSize != nil {
+		size = *g.codeSize + minSize
```
HerbertJordan, line 139: This is surprising ... the field is named `codeSize` but it does not seem to put a constraint on the code size. Instead, it adds some extra "boost" to the minimum code size?

I would have thought to make it simply a strict constraint on the code size. If `codeSize` is set and it has a value, that is the size of the produced code. If the value is smaller than the `minSize`, the generator produces an `Unsatisfiable` error.

This way, the parameter has a simpler interpretation. Wdyt?

HerbertJordan (changes requested):

go/ct/gen/code.go
LuisPH3: Right, that would work if the code size would be generated using a range. but instead the generated size is using this ad-hoc code:
image
I am not sure what takes precedence here, the feature of returning small code sizes seems reasonable to me. 
how would you proceed, would you add an exponential generator to the range generator tool?

HerbertJordan (approved):

facuMH (approved):

simonlechner (approved):

Issue #512, opened 2024-06-07
Count words with overflow check.


LuisPH3:
regex `\((\w+) \+ 31\) / 32 ` produce many uses where there is no overflow check.  6 times in the spec
make sure this is considered.

HerbertJordan:
Related to #493

PR #513, opened 2024-06-07
MCOPY ct

LuisPH3:
Add specification for MCOPY


go/ct/st/memory.go
```diff
@@ -88,9 +88,16 @@ func (m *Memory) Grow(offset, size uint64) {
 // ExpansionCosts calculates the expansion costs for the given offset and size.
 // It does not grow memory. It also returns offset and size converted to uint64.
 func (m *Memory) ExpansionCosts(offset_u256, size_u256 U256) (memCost vm.Gas, offset, size uint64) {
-	if !size_u256.IsUint64() || (!offset_u256.IsUint64() && !size_u256.IsZero()) {
+	if !size_u256.IsUint64() {
```
facuMH, line 91: maybe we should define our coding style for this kind of cases.

go/ct/spc/specification.go
```diff
@@ -722,6 +722,58 @@ func getAllRules() []Rule {
...
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+		},
+		effect: func(s *st.State) {
+			destOffsetU256 := s.Stack.Pop()
+			offsetU256 := s.Stack.Pop()
+			sizeU256 := s.Stack.Pop()
+
+			srcCost, srcOffset, size := s.Memory.ExpansionCosts(offsetU256, sizeU256)
+			destCost, destOffset, _ := s.Memory.ExpansionCosts(destOffsetU256, sizeU256)
+			wordCountCost := vm.Gas(3 * ((size + 31) / 32))
```
HerbertJordan, line 747: The expression `size+31` may overflow.

HerbertJordan (approved):

go/ct/st/memory.go
LuisPH3: It is not only style, the logic is different, geth will ignore overshooting offsets if the size is 0, generating no expansion. Previous code did generate expansion for size 0.

go/ct/spc/specification.go
LuisPH3: Follow up #512

PR #515, opened 2024-06-07
CT: Add Probe command

facuMH:
This PR adds probe command, which generates unconstrained (except for `IsCode(Pc()),RevisionBounds(common.R07_Istanbul, common.R11_Paris))`)  random states and tries to find a rule that can process it. 
 ```
 [t=  30:00] - Processing ~116k tests per second, total 221064464`
All tests passed successfully!
Issues found: 0
```
It also move the issue tracking from `run.go` to `cli/issues.go` since it is used by multiple commands.

This PR depends on #518 so that the `probe` command runs successfully

### CODE REVIEW

File: go/ct/driver/run.go [hunk 1]
```diff
@@ -163,15 +162,15 @@ func doRun(context *cli.Context) error {
 		if err != nil {
 			return fmt.Errorf("failed to create output directory for %d issues", len(issues))
 		}
-		for i, issue := range issuesCollector.issues {
+		for i, issue := range issuesCollector.GetIssues() {
```
HerbertJordan, line 165: Use `issueCollector.ExportIssues()` here.

go/ct/driver/cli/utils.go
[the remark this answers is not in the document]
LuisPH3: :+1:

go/ct/gen/code.go
```diff
@@ -114,6 +114,12 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 		minSize += size
 	}
 
+	// If there are any variables that need to be bound to code, there must be at
+	// least on instruction in the resulting code.
+	if minSize == 0 && len(g.varIsCodeConstraints) > 0 {
```
LuisPH3, line 119: We will merge this in my pr soon, where I have the test for it. 
if you get the chance, remove it. otherwise we will rebase it.

LuisPH3 (approved):

LuisPH3 (approved):

HerbertJordan (changes requested):
I will review this once it is marked as ready for review and the title has been improved.

LuisPH3 (approved):

LuisPH3 (approved):

simonlechner (dismissed):

LuisPH3 (approved):

go/ct/driver/probe.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/Fantom-foundation/Tosca/go/ct/gen"
+	"github.com/Fantom-foundation/Tosca/go/ct/rlz"
+	"github.com/Fantom-foundation/Tosca/go/ct/spc"
+	"github.com/Fantom-foundation/Tosca/go/ct/st"
+	"github.com/dsnet/golib/unitconv"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/exp/maps"
+)
+
+var ProbeCmd = cli.Command{
```
HerbertJordan, line 36: Can you wrap the command into the `cliUtils.AddCommonFlags(..)` to add general flag support?

go/ct/driver/probe.go
```diff
@@ -0,0 +1,198 @@
...
+	seed := cliUtils.SeedFlag.Fetch(context)
+
+	timeout := time.Duration(context.Int("timeout")) * time.Minute
+	hasTimeouted := false
+
+	// The constraints to be placed on generated states.
+	condition := rlz.And(
+		rlz.IsCode(rlz.Pc()),
+		// update to cancun when all necessary changed have been implemented
+		rlz.RevisionBounds(common.MinRevision, common.R12_Shanghai),
```
HerbertJordan, line 77: I doubt that we will remember updating this. 

Can we replace `common.R12_Shanghai` with `common.NewestSupportedRevision`?
If not, maybe we can introduce a `common.NewestFullySpecifiedRevision` for these kind of use cases?

go/ct/driver/probe.go
```diff
@@ -0,0 +1,198 @@
...
+	progressGroup.Add(1)
+	go func() {
+		defer progressGroup.Done()
+		start := time.Now()
+		last := uint64(0)
+		for {
+			select {
+			case <-stopProgressPrinter:
+				return
+			case <-time.After(5 * time.Second):
```
HerbertJordan, line 95: Consider using a [Ticker](https://gobyexample.com/tickers) for this. Tickers are made for this use case.

go/ct/driver/probe.go
```diff
@@ -0,0 +1,198 @@
...
+				last = current
+				rate := float64(diff) / 5
+				fmt.Printf(
+					"[t=%4d:%02d] - Processing ~%s tests per second, total %d\n",
+					int(relativeTime.Seconds())/60, int(relativeTime.Seconds())%60,
+					unitconv.FormatPrefix(rate, unitconv.SI, 0), current,
+				)
+				if relativeTime > timeout {
+					fmt.Printf("Timeout reached, stopping the probing\n")
+					hasTimeouted = true
```
HerbertJordan, line 108: There is a race-condition on `hasTimeouted`. You should make it an [atomic.Bool](https://pkg.go.dev/sync/atomic#Bool).

HerbertJordan (changes requested):

go/ct/driver/probe.go
facuMH: `common.NewestSupportedRevision` is already Cancun. 

not sure introducing another concept that we also need to eventually update is better than leaving  a `TODO`.

LuisPH3: I would use newest every time. today we are Cancun, but this will certainly change and then we would have to revisit this code. Your rule is defined for the foreseeable future, that is newest supported.

facuMH: the problem with using `NewestSupportedRevision` is that we don't currently support all instructions in that set, so this command would fail. I'm ok with halting this PR until we have full support for `Cancun`

LuisPH3 (approved):

HerbertJordan (approved):

simonlechner (approved):

PR #517, opened 2024-06-10
Add offset aware routine to count words in size operations.

LuisPH3:
This PR affects the specification

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -249,7 +249,7 @@ func getAllRules() []Rule {
 			}
 			s.Gas -= memExpCost
 
-			wordCost := vm.Gas(6 * ((size + 31) / 32))
+			wordCost := vm.Gas(6 * SizeInWords(size))
 			if s.Gas < wordCost {
 				s.Status = st.Failed
 				s.Gas = 0
@@ -1237,8 +1237,8 @@ func getAllRules() []Rule {
 			sizeU256 := s.Stack.Pop()
 
 			cost, destOffset, size := s.Memory.ExpansionCosts(destOffsetU256, sizeU256)
-			minimumWordSize := (size + 31) / 32
-			cost += vm.Gas(3 * minimumWordSize)
+
+			cost += vm.Gas(3 * SizeInWords(size))
 			if s.Gas < cost {
 				s.Status = st.Failed
 				s.Gas = 0
@@ -1315,8 +1315,7 @@ func getAllRules() []Rule {
 			sizeU256 := s.Stack.Pop()
 
 			expansionCost, destOffset, size := s.Memory.ExpansionCosts(destOffsetU256, sizeU256)
-			words := (size + 31) / 32
-			expansionCost += vm.Gas(3 * words)
+			expansionCost += vm.Gas(3 * SizeInWords(size))
 			if s.Gas < expansionCost {
 				s.Status = st.Failed
 				s.Gas = 0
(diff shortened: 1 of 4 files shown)


facuMH (dismissed):
nice refactor

go/ct/spc/specification.go
```diff
@@ -249,7 +249,13 @@ func getAllRules() []Rule {
...
 			s.Gas -= memExpCost
 
-			wordCost := vm.Gas(6 * ((size + 31) / 32))
+			words, overflow := wordCountWithOverflow(size)
+			if overflow {
+				s.Status = st.Failed
+				s.Gas = 0
+				return
+			}
+			wordCost := vm.Gas(6 * words)
```
HerbertJordan, line 258: As part of the `ExpansionCosts` function the gas costs for the expansion is computed. If this cost [overflows](https://github.com/Fantom-foundation/Tosca/blob/7d737ec9f5f83e4e53b7b8c73cf292fdba1cecde/go/ct/st/memory.go#L108), `MaxUint64` is returned for the expansion cost, which always exceeds the amount of available gas (since the static costs of this operation are not 0).

Thus, technically this `wordCost` computation was guarded against overflows, but it was not obvious. I agree, this change makes this more apparent, but I would suspect that the `overflow` branch is dead code.

go/ct/spc/specification.go
```diff
@@ -2362,3 +2397,11 @@ func sumWithOverflow(values ...vm.Gas) (vm.Gas, bool) {
...
 	return res, false
 }
+
+func wordCountWithOverflow(size uint64) (vm.Gas, bool) {
+	offsetted := size + 31
+	if offsetted < size {
+		return 0, true
+	}
+	return vm.Gas(offsetted / 32), false
+}
```
HerbertJordan, line 2407: There is a subtle semantic hidden in this code -- what to do with a `size ∈ [MaxUint64-30,MaxUint64]`.

For those values, you could compute a valid result by returning `MaxUint64/32`. However, this implementation chooses to signal an overflow. Is such an overflow actually an error? After all, it only happens for an intermediate value.

Maybe a cleaner implementation may be:
```suggestion
func getNumWords(size uint64) uint64 {
        if size > math.MaxUint64-31 {
		return math.MaxUint64/32 + 1
	}
	return (size + 31) / 32
}
```

go/ct/spc/specification.go
```diff
@@ -2362,3 +2397,11 @@ func sumWithOverflow(values ...vm.Gas) (vm.Gas, bool) {
 	}
 	return res, false
 }
+
+func wordCountWithOverflow(size uint64) (vm.Gas, bool) {
```
HerbertJordan, line 2401: The result should not be of the type `vm.Gas` since it is a number of words.

go/ct/spc/specification.go
HerbertJordan: I looked it up in go-ethereum, and they actually define it like this: [here](https://github.com/Fantom-foundation/go-ethereum-sonic/blob/2400937cc3b1c370ea8735d7af86eac621e4eb70/core/state_transition.go#L120-L127)

`evmone` ignores the overflows: [here](https://github.com/ethereum/evmone/blob/705409fdcd06e3ffbcfc8955a3ad5acfd7e360f2/lib/evmone/instructions.hpp#L61-L64)

HerbertJordan (changes requested):
I think the overflow needs to be handled differently. It should not cause a failure but it should perform the correct rounding.

go/ct/spc/specification.go
LuisPH3: correct

facuMH (approved):

HerbertJordan (approved):
Nice fix, thanks!

PR #522, opened 2024-06-10
CT: Specification Soundness test

facuMH:
This PR adds:
- a test to check for specification soundness
- a rule that checks that when the PC is pointing at data, nothing gets executed
- a rule for each of the 256 available codes in the 2 byte spectrum, and checking the invalid codes are treated properly.

### DIFF
--- a/go/ct/spc/specification_map.go
+++ b/go/ct/spc/specification_map.go
@@ -49,24 +49,26 @@ func NewSpecificationMap(rules ...Rule) Specification {
 
 func (s *specificationMap) GetRulesFor(state *st.State) []Rule {
 	result := []Rule{}
-	var opString string
+	getRules := func(opString string) {
+		for _, rule := range s.rules[opString] {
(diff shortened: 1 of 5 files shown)


LuisPH3 (dismissed):

go/ct/spc/specification_map.go
```diff
@@ -49,24 +49,27 @@ func NewSpecificationMap(rules ...Rule) Specification {
 
 func (s *specificationMap) GetRulesFor(state *st.State) []Rule {
 	result := []Rule{}
-	var opString string
+	getRules := func(opString string) {
+		opRules := s.rules[opString]
+		for _, rule := range opRules {
```
HerbertJordan, line 54: nit:
```suggestion
		for _, rule := range s.rules[opString] {
```

HerbertJordan (approved):

simonlechner (approved):

PR #527, opened 2024-06-11
LFVM: add BLOBHASH

facuMH:
This PR adds the `BLOBHASH` op implementation to lfvm, along with a `[ ]Hash` in the `TransactionContext` of `vm/interpreter` to hold the old blob hashes to be queried. 

This PR contributes to #468


go/vm/lfvm/instructions.go
```diff
@@ -677,6 +677,20 @@ func opBaseFee(c *context) {
 	}
 }
 
+func opBlobHash(c *context) {
```
LuisPH3, line 680:
```suggestion
if c.isCancun() {
	c.status = INVALID_INSTRUCTION
	return
}

index := c.stack.pop()
lenght :_= uint64(len(c.context.GetTransactionContext().BlobHashes)
if index.IsUint64() && index.Uint64() < lenght) {
     stack.push(c.context.GetTransactionContext().BlobHashes[index.Uint64()][:])
} else{
   stack.push(0)
}
```
this is completely optional, but the point here is:

1. early bail out, avoid nesting ifs
2. smaller boolean expressions, no call/compute in boolean code
3. explicit stack operations that reflect instruction spec.
4. positive boolean, instead of ! + || (some linters would ask for parenthesis here)

LuisPH3: plz. check replacement, there are some bugs. github does not allow me to edit it anymore.

LuisPH3 (approved):

go/vm/lfvm/instructions.go
HerbertJordan, line 680: Please add unit tests for this operation.

simonlechner (approved):

LuisPH3 (approved):

PR #528, opened 2024-06-11
evmzero: Add blobhash

facuMH:
This PR implements `BLOBHASH` in evmzero and a few tests for it.

This PR contributes to #468


HerbertJordan (approved):
Looks good, thanks!

cpp/vm/evmzero/interpreter.cc
```diff
@@ -976,6 +976,27 @@ struct Impl<OpCode::BASEFEE> {
...
+struct Impl<OpCode::BLOBHASH> {
+  constexpr static OpInfo kInfo{
+      .pops = 1,
+      .pushes = 1,
+      .static_gas = 3,
+      .introduced_in = EVMC_CANCUN,
+  };
+
+  static OpResult Run(uint256_t* top, Context& ctx) noexcept {
+    const auto index = static_cast<int64_t>(top[0]);
```
LuisPH3, line 989: the check in line 991 may be too late, since you already rounded up the value in this line. Can you check if this is true? Other instructions may ask up-front if the stack value fits in an int64 before reading and casting it. 
I think you may need to pop it into a int256 variable first, and ask if it overflows.

LuisPH3 (changes requested):

cpp/vm/evmzero/interpreter.cc
facuMH: the check on line 991 that checks regarding `int64_t` max, checks against the "uncasted" value. I simply cast it here so that if the max check passes, I don't need to cast inside the if, and then again to use it as index.

LuisPH3: I see now, I  got confused, because you sometimes use index and or top[0] this could use less complication and be a little more straight forward.

LuisPH3 (changes requested):

LuisPH3 (approved):

cpp/vm/evmzero/interpreter.cc
facuMH: nice offline discussion on how to improve !!! thanks

LuisPH3 (approved):

simonlechner (approved):

PR #531, opened 2024-06-12
CT: Add BlobHashes to TransactionContext

facuMH:
This PR adds a slice of hashes to `BlockContext`, and the corresponding generator. This is to store a list of hashes, which can later on be queried by `BLOBHASH` instruction. But considering the implementation of that instruction requires a new Condition and maybe more, I think it might be best to merge this changes first and deal with the Condition and Rules in separate PRs.

This PR contributes to #468 

EDIT: noticed that `blobbasefee` was not added to `adapter_test` so it is added in this PR as well even if not strictly related.


LuisPH3 (approved):

go/ct/utils/adapter_test.go
```diff
@@ -313,9 +313,9 @@ func TestAdapter_ParameterConversion(t *testing.T) {
 
 	for name, property := range properties {
 		t.Run(name, func(t *testing.T) {
-			state := st.State{}
-			property.set(&state)
-			params := ToVmParameters(&state)
+			state := st.NewState(st.NewCode([]byte{}))
+			property.set(state)
+			params := ToVmParameters(state)
```
HerbertJordan, line 318: This change looks unrelated ... why is it needed?

go/ct/st/block_context.go
```diff
@@ -32,6 +34,44 @@ type BlockContext struct {
...
+	if len(b.BlobHashes) != len(other.BlobHashes) {
+		isEqual = false
+	} else {
+		for i, hash := range b.BlobHashes {
+			if !slices.Equal(hash[:], other.BlobHashes[i][:]) {
+				isEqual = false
+				break
+			}
+		}
+	}
```
HerbertJordan, line 70: This should be the same as
```suggestion
	isEqual = isEqual && (slices.Equal(b.BlobHashes, other.BlobHashes))
```

go/ct/st/block_context.go
```diff
@@ -45,6 +85,16 @@ func (b *BlockContext) Diff(other *BlockContext) []string {
 		ret = append(ret, blockDifference+fmt.Sprintf("blob base fee: %v vs %v\n", b.BlobBaseFee, other.BlobBaseFee))
 	}
 
+	if len(b.BlobHashes) != len(other.BlobHashes) {
+		ret = append(ret, blockDifference+fmt.Sprintf("blob hashes: %v vs %v\n", b.BlobHashes, other.BlobHashes))
+	} else {
+		for i, hash := range b.BlobHashes {
+			if !slices.Equal(hash[:], other.BlobHashes[i][:]) {
+				ret = append(ret, blockDifference+fmt.Sprintf("blob hash: %v vs %v\n", hash, other.BlobHashes[i]))
```
HerbertJordan, line 93: Please include `i` in the output.

go/ct/st/block_context.go
```diff
@@ -45,6 +85,16 @@ func (b *BlockContext) Diff(other *BlockContext) []string {
 		ret = append(ret, blockDifference+fmt.Sprintf("blob base fee: %v vs %v\n", b.BlobBaseFee, other.BlobBaseFee))
 	}
 
+	if len(b.BlobHashes) != len(other.BlobHashes) {
+		ret = append(ret, blockDifference+fmt.Sprintf("blob hashes: %v vs %v\n", b.BlobHashes, other.BlobHashes))
+	} else {
+		for i, hash := range b.BlobHashes {
+			if !slices.Equal(hash[:], other.BlobHashes[i][:]) {
```
HerbertJordan, line 92: Simpler:
```suggestion
			if hash != other.BlobHashes[i] {
```

go/ct/st/block_context.go
```diff
@@ -32,6 +34,44 @@ type BlockContext struct {
 	TimeStamp   uint64     // Block's timestamp in unix time in seconds
 }
 
+// NewBlockContext creates a new block context
+func NewBlockContext() *BlockContext {
+	return &BlockContext{BlobHashes: []vm.Hash{}}
```
HerbertJordan, line 39: The handling of the `BlockContext` could be simplified if `nil` is considered a valid value for the list.

go/ct/st/block_context_test.go
```diff
@@ -74,3 +76,53 @@ func TestBlockContext_String(t *testing.T) {
...
+	b2 := b1.Clone()
+	if !b1.Equal(b2) {
+		t.Error("Cloned block context is not equal")
+	}
+}
+
+func TestBlockContext_ClonesAreIndependent(t *testing.T) {
+	b1 := BlockContext{}
+	b2 := b1.Clone()
+	b2.BlockNumber = 1
```
HerbertJordan, line 91: This does not cover the one field that is not trivially copyable.

go/ct/st/block_context.go
```diff
@@ -23,6 +24,7 @@ import (
 type BlockContext struct {
 	BaseFee     U256       // Base fee in wei
 	BlobBaseFee U256       // Blob Base fee in wei
+	BlobHashes  []vm.Hash  // List of blob hashes
```
HerbertJordan, line 27: I have not looked into this in detail, but as I understand the Blob system right now, blobs should be associated to the `CallContext`, not the `BlockContext`. Each transaction in a block has its own set of blobs.

go/ct/gen/block_context.go
```diff
@@ -175,32 +176,38 @@ func (b *BlockContextGenerator) bindVariablesInRangeConstraints(resultingBlockNu
...
 
 	chainId := RandU256(rnd)
 	coinbase, err := RandAddress(rnd)
 	if err != nil {
-		return st.BlockContext{}, err
+		return &st.BlockContext{}, err
+	}
+
+	blobHashes := []vm.Hash{}
+	blobHashesCount := rnd.Intn(st.MaxDataSize)
```
HerbertJordan, line 208: That is unreasonably large (MaxDataSize = 1024). Ethereum, right now, has a limit of up to 6 Blobs per transaction. So this limit should be single-digit.

HerbertJordan (changes requested):

go/ct/st/block_context.go
HerbertJordan: Thinking about it, it is neither part of the `BlockContext` nor of the `CallContext`. It is part of the `TransactionContext`, which is a new thing that could be added.

This would also move the generation challenge of the the constraint solver out of the `BlockContext` generator, avoiding adding additional complexity to it. Wdyt?

facuMH: well, since the the origin of `BlockContext` and `CallContext` it was mentioned both of them had info that was actually `TransactionContext`. Should we define more fields to be moved from either of those to a `TransactionContext` ? 
That should probably a separate PR. Maybe a follow-up PR and this one just adds `BlobHashes`, which are later on moved to the `TransactionContext` ?

go/ct/utils/adapter_test.go
facuMH: because just calling `st.State{}` does not allocate for underlying structures, such as `Storage`, `Accounts` or `BlockContext`, I noticed `set` functions do this, so I'll revert this change and add the necessary calls to `NewBlockContext`

HerbertJordan (commented):
Please re-add me as a reviewer when the conflicts are resolved, the CI builds have passed, and the PR is converted from a draft to a pull request.

go/ct/st/state_test.go
```diff
@@ -47,7 +47,7 @@ func getNewFilledState() *State {
 	s.Logs.AddLog([]byte{4, 5, 6}, NewU256(21), NewU256(22))
 	s.CallContext = CallContext{AccountAddress: vm.Address{0x01}}
 	s.BlockContext = BlockContext{BlockNumber: 1}
-	s.TransactionContext = TransactionContext{OriginAddress: vm.Address{0x01}}
+	s.TransactionContext = &TransactionContext{BlobHashes: []vm.Hash{{4, 3, 2, 1}}}
```
HerbertJordan, line 50: Should this add `BlobHashes` besides the `OriginAddress` or is the replacement intentional?

go/ct/st/transaction_context.go
```diff
@@ -12,30 +12,73 @@ package st
...
+		ret = append(ret, transactionContextDiff+fmt.Sprintf("origin address: %v vs. %v\n", t.OriginAddress,
 			other.OriginAddress))
 	}
 
+	if len(t.BlobHashes) != len(other.BlobHashes) {
+		ret = append(ret, transactionContextDiff+fmt.Sprintf("blob hashes: %v vs %v\n", t.BlobHashes, other.BlobHashes))
+	} else {
+		for i, hash := range t.BlobHashes {
+			if hash != other.BlobHashes[i] {
+				ret = append(ret, transactionContextDiff+fmt.Sprintf("blob hash at location%v: %v vs %v\n", i, hash, other.BlobHashes[i]))
```
HerbertJordan, line 61:
```suggestion
				ret = append(ret, transactionContextDiff+fmt.Sprintf("blob hash at location %v: %v vs %v\n", i, hash, other.BlobHashes[i]))
```

go/ct/st/transaction_context_test.go
```diff
@@ -43,16 +46,68 @@ func TestTransactionContext_String(t *testing.T) {
...
+		OriginAddress: vm.Address{1, 2, 3, 4},
+		BlobHashes:    []vm.Hash{{1, 2, 3, 4}},
+	}
+
+	t2 := t1.Clone()
+
+	if !t1.Equal(t2) {
+		t.Errorf("Cloned transaction context is not equal to original")
+	}
+}
```
HerbertJordan, line 80: It should test that the list of blob hashes is indeed deep-copied.

go/ct/st/transaction_context_test.go
```diff
@@ -43,16 +46,68 @@ func TestTransactionContext_String(t *testing.T) {
...
+
+func TestTransactionContext_Equal(t *testing.T) {
+
+	tests := map[string]struct {
+		change func(*TransactionContext, *TransactionContext)
```
HerbertJordan, line 85: Why the change function instead of two fields `a` and `b`?

HerbertJordan (dismissed):

go/ct/st/state_test.go
facuMH: is intentional, since the idea is to change something in the structure, and in this case, changing `BlobHashes` is more interesting than an address.

go/ct/st/transaction_context_test.go
facuMH: I thought change function conveyed the right meaning in this case, to modify depending on the test case.

go/ct/st/transaction_context_test.go
```diff
@@ -43,16 +47,78 @@ func TestTransactionContext_String(t *testing.T) {
...
+			t2.BlobHashes = []vm.Hash{{4}}
+		}, false},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			t1 := TransactionContext{}
+			t2 := TransactionContext{}
+			test.change(&t1, &t2)
```
simonlechner, line 118: You could add an equals check before applying the change, this would help with debugging if the Equals method is broken.

LuisPH3 (approved):

go/ct/st/transaction_context_test.go
HerbertJordan: A test for an `Equal` function should be about pairs of values as they are at the time of the comparison, not about the historic evolution how they become to be what hey are today -- what is implied with the name `change` and the way changes are used in other tests.

Also, using a pairs of values would make the code much shorter and simpler:
```
	tests := map[string]struct {
		a, b *TransactionContext
	}{
		"zero-values": {
			a: &TransactionContext{},
			b: &TransactionContext{},
		},
		"match": {
			a: &TransactionContext{BlobHashes: []vm.Hash{{1}}},
			b: &TransactionContext{BlobHashes: []vm.Hash{{1}}},
		},
		"mismatch": {
			a: &TransactionContext{BlobHashes: []vm.Hash{{1}}},
			b: &TransactionContext{BlobHashes: []vm.Hash{{4}}},
		},
		"mismatch-different-blob-hash-length": {
			a: &TransactionContext{BlobHashes: []vm.Hash{{1}, {2}}},
			b: &TransactionContext{BlobHashes: []vm.Hash{{4}}},
		},
	}

			a, b := test.a, test.b
			if got, want := a.Eq(b), reflect.DeepEqual(a, b); got != want {
				t.Errorf("failed to compare contexts %v and %v, wanted %t, got %t", a, b, want, got)
			}
		})
	}
```

go/ct/st/transaction_context_test.go
```diff
@@ -43,16 +47,81 @@ func TestTransactionContext_String(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+
+			t1 := TransactionContext{}
+			t2 := TransactionContext{}
+			if !t1.Eq(&t2) {
+				t.Error("transaction context equals is broken")
+			}
+			test.change(&t1, &t2)
+			if t1.Eq(&t2) != test.wanted {
+				t.Error("Block contexts are equal")
```
HerbertJordan, line 123: This error message is wrong.

go/ct/st/transaction_context_test.go
```diff
@@ -43,16 +47,81 @@ func TestTransactionContext_String(t *testing.T) {
...
+			t2.BlobHashes = []vm.Hash{{1}}
+		}, true},
+		"mismatch": {func(t1, t2 *TransactionContext) {
+			t1.BlobHashes = []vm.Hash{{1}}
+			t2.BlobHashes = []vm.Hash{{4}}
+		}, false},
+		"mismatch-different-blobhash-length": {func(t1, t2 *TransactionContext) {
+			t1.BlobHashes = []vm.Hash{{1}, {2}}
+			t2.BlobHashes = []vm.Hash{{4}}
+		}, false},
```
HerbertJordan, line 110: These tests do not cover the `OriginAddress`.

HerbertJordan (changes requested):

HerbertJordan (approved):

LuisPH3 (approved):

simonlechner (approved):

PR #532, opened 2024-06-12
CT: Add PresentBlobHash and AbsentBlobHash conditions

facuMH:
This PR adds the conditions `HasBlobHash` and `HasNotBlobHas` to force generation of states variables with a value with (or not) a value that can be used as an index of `blobHashes`

This PR contributes to #468


go/ct/rlz/conditions_test.go
```diff
@@ -428,3 +429,65 @@ func TestCondition_InOutOfRangeGetTestValues(t *testing.T) {
 		})
 	}
 }
+
+func TestCondition_BlobHashes_Restrict(t *testing.T) {
```
LuisPH3, line 433: Please check for unsatisfiable,

both has and has not for the same param?

LuisPH3 (changes requested):

go/ct/gen/block_context.go
HerbertJordan, line 1: Unit tests? No test-driven development?

go/ct/gen/state.go
```diff
@@ -257,6 +257,14 @@ func (g *StateGenerator) SetBlockNumberOffsetValue(variable Variable, value int6
 	g.blockContextGen.SetBlockNumberOffsetValue(variable, value)
 }
 
+func (g *StateGenerator) RestrictToHaveBlobHash(variable Variable) {
```
HerbertJordan, line 260: Please fix the grammar: maybe "RestrictToPresentBlobHashIndex"

go/ct/gen/block_context.go
```diff
@@ -205,8 +209,34 @@ func (b *BlockContextGenerator) Generate(assignment Assignment, rnd *rand.Rand)
 	}
 
 	blobHashes := []vm.Hash{}
-	blobHashesCount := rnd.Intn(st.MaxDataSize)
-	for i := 0; i < blobHashesCount; i++ {
+	// we do not want this count to be zero, so we set it to a random value between 1 and MaxDataSize
+	blobHashesCount := rnd.Uint64n(st.MaxDataSize-1) + 1
```
HerbertJordan, line 213: Heads-up: the number of generated blob hashes is part of the data produced by this generator. Depending on the set constraints, there may be a lower limit.

Also, the maximum should be single-digit, as the limit on Ethereum right now is 6.

HerbertJordan (changes requested):
I will review this once the base PR is closer to

go/ct/gen/state.go
facuMH: I actually thought the original name was good, since it is saying "restrict variable to have a blob hash". 
What if we drop the "Restrict" and just call them "PresentBlobHashIndex" and "AbsentBlobHashIndex" ?

HerbertJordan: The thing is: the variable will not be mapped in the assignment to a hash but to an index that has a hash assigned to it in the blob-hash list. That's why I think "RestritToHaveBlobHash" is not fitting.

HerbertJordan (changes requested):
It looks like this PR contains changes of preceding PRs that have been reviewed before.

Please re-base the PR and re-add me as a reviewer when ready.

go/ct/gen/transaction_context.go
```diff
@@ -19,34 +23,147 @@ import (
...
+				if assignedValue.IsUint64() && assignedValue.Uint64() < minMustNotHaveBlobHash {
+					minMustNotHaveBlobHash = assignedValue.Uint64()
+				}
+			}
+		}
+	}
+
+	if maxMustHaveBlobHash > minMustNotHaveBlobHash ||
+		maxMustHaveBlobHash == math.MaxUint64 ||
+		minMustNotHaveBlobHash == 0 {
```
LuisPH3, line 64: Minimum absent blob hash ==0 is valid if not present blobs hash are required, isn't it? 
I guess the intent is to generate always at least 1 element, but if the condition asks for a non-existing hash at position 0 ?

LuisPH3 (approved):
Please check If we really need to generate at least 1 element.

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +94,205 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+		} else {
+			if assignedValue.Uint64() < uint64(len(txCtx.BlobHashes)) {
+				t.Errorf("Assigned value for %v %v is not out of range.", variable.String(), assignedValue.Uint64())
+			}
+		}
+		if shouldBeAssigned {
+			if !assignedValue.Eq(value) {
+				t.Errorf("Assigned value for %v %vis not the expected value.", variable.String(), assignedValue.Uint64())
+			}
+		}
```
HerbertJordan, line 123: This last check seems to only make sure that what was assigned before the generator was executed remained present.

A more readable way to do this would be to
- remove this condition here and the two parameters `shouldBeAssigned` and `value` from this function
- in the test loop below, create a backup of the assignment before you pass it to the generator
- after the generator ran, check that the resulting assignment contains everything that is in the backup

This improved the readability of the `variableCheck` calls.

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +94,205 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+	tests := map[string]struct {
+		setup func(*TransactionContextGenerator, *Assignment)
+		check func(st.TransactionContext, Assignment, *testing.T)
+	}{
+		"present": {
+			setup: func(txCtxGen *TransactionContextGenerator, _ *Assignment) {
+				txCtxGen.PresentBlobHashIndex(Variable("v1"))
+			},
+			check: func(txCtx st.TransactionContext, assignment Assignment, t *testing.T) {
+				variableCheck(txCtx, assignment, t, Variable("v1"), true, false, common.NewU256(0))
```
HerbertJordan, line 135: If you follow the suggestion listed above, what remains of the "variableCheck" is a list of variables that are in our outside the branch range. In this case, it would be more readable to redefine the test struct as
```
struct {
   setup func(*TransactionContextGenerator, *Assignment)
   shouldBePresent []Variable
   shouldBeAbsend []Variable
```
and then inline the `variableCheck` function in the test loop below.

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +94,205 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+		}
+		if shouldBeAssigned {
+			if !assignedValue.Eq(value) {
+				t.Errorf("Assigned value for %v %vis not the expected value.", variable.String(), assignedValue.Uint64())
+			}
+		}
+	}
+
+	tests := map[string]struct {
+		setup func(*TransactionContextGenerator, *Assignment)
```
HerbertJordan, line 127: Since `Assignment` is a `map` type, and thus a reference type in Go, you do not need to pass a pointer. You can just pass an `Assignment`.

go/ct/rlz/conditions_test.go
```diff
@@ -426,3 +427,124 @@ func TestCondition_InOutOfRangeGetTestValues(t *testing.T) {
...
+
+				if checked, err := test.condition.Check(state); err != nil || !checked {
+					t.Errorf("failed to check condition: %v", err)
+				}
+			}
+		})
+	}
+}
+
+func TestCondition_BlobHashesProducesUnsatisfiable(t *testing.T) {
```
HerbertJordan, line 493: nit: the name seems off ...

go/ct/rlz/conditions_test.go
```diff
@@ -426,3 +427,124 @@ func TestCondition_InOutOfRangeGetTestValues(t *testing.T) {
...
+			test.condition.Restrict(gen)
+			_, err := gen.Generate(rand.New(0))
+			if err == nil {
+				t.Errorf("expected unsatisfiable condition")
+			}
+		})
+	}
+}
+
+func TestCondition_BlobHashesPRoducesGetTestValues(t *testing.T) {
```
HerbertJordan, line 527:
```suggestion
func TestCondition_HasBlobHashPoducesTestValues(t *testing.T) {
```

go/ct/rlz/conditions_test.go
```diff
@@ -426,3 +427,124 @@ func TestCondition_InOutOfRangeGetTestValues(t *testing.T) {
...
+			}
+
+			if testValues[0].(*testValue[bool]).value != true {
+				t.Errorf("unexpected test value, got %v, wanted true ", testValues[0].(*testValue[bool]).value)
+			}
+
+			if testValues[1].(*testValue[bool]).value != false {
+				t.Errorf("unexpected test value, got %v, wanted false ", testValues[1].(*testValue[bool]).value)
+			}
+
```
HerbertJordan, line 547: It is generally better to test for properties of the expected result, not implementation details. For instance, here, the implementation details like the number of generated test cases or their type should ideally not be exploited.

A better way to test this could look like this:
```suggestion
			var matchCovered, failCovered bool
			for _, value := range condition.GetTestValues() {
				gen := gen.NewStateGenerator()
				value.Restrict(gen)
				state, err := gen.Generate(rand.New(0))
				if err != nil {
					t.Fatalf("failed to build state: %v", err)
				}
				if matches, err := condition.Check(state); err != nil {
					t.Errorf("failed to check condition: %v", err)
				} else if matches {
					matchCovered = true
				} else {
					failCovered = true
				}
			}

			if !matchCovered {
				t.Errorf("no test value matched the condition")
			}

			if !failCovered {
				t.Errorf("no test value failed the condition")
			}
```
This way, the intention is tested and only public interfaces are utilized.

HerbertJordan (approved):
Looks good, I just added a few style remarks. Thanks!

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +93,192 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+				txCtxGen.IsAbsentBlobHashIndex(Variable("v1"))
+			},
+		},
+		"absent-and-present": {
+			setup: func(txCtxGen *TransactionContextGenerator, _ *Assignment) {
+				txCtxGen.IsAbsentBlobHashIndex(Variable("v1"))
+				txCtxGen.IsPresentBlobHashIndex(Variable("v1"))
+			},
+		},
+		"assigned-and-present": {
```
LuisPH3, line 238: nit:

duplicated? "assigned-max-present

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +93,192 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+			setup: func(txCtxGen *TransactionContextGenerator, assignment *Assignment) {
+				(*assignment)[Variable("v1")] = common.NewU256(math.MaxUint64)
+				txCtxGen.IsPresentBlobHashIndex(Variable("v1"))
+			},
+		},
+		"assigned-present-and-assigned-absent": {
```
LuisPH3, line 244: nit:
```suggestion
		"absent-less-than-assigned": {
```

go/ct/rlz/conditions_test.go
```diff
@@ -463,3 +464,134 @@ func TestCondition_RestrictTransientStorageAndCheck(t *testing.T) {
...
+		condition Condition
+		setup     func(*st.State)
+	}{
+		"hasBlobHash-first": {
+			condition: And(HasBlobHash(Param(0)), HasNoBlobHash(Param(0))),
+			setup: func(state *st.State) {
+				state.Stack.Push(NewU256(0))
+			},
+		},
+		"hasNotBlobHash-first": {
```
LuisPH3, line 542:
```suggestion
		"hasNoBlobHash-first": {
```

LuisPH3 (dismissed):

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +93,186 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			rnd := rand.New(0)
+			txCtxGen := NewTransactionContextGenerator()
+			assignment := Assignment{}
+			test.setup(txCtxGen, assignment)
+			assignmentBackup := assignment
```
HerbertJordan, line 186: The assignment is a `map`, and thus a reference type. The backup you create here is just an alias of the `assignment`, and all changes to the `assignment` are also affecting the backup.

To make a real backup, use `maps.Clone()`.

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +93,186 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			rnd := rand.New(0)
+			txCtxGen := NewTransactionContextGenerator()
+			assignment := Assignment{}
+			test.setup(txCtxGen, &assignment)
+			assignmentBackup := assignment
```
HerbertJordan, line 266: Same here, this is not really a backup.

go/ct/gen/transaction_context_test.go
```diff
@@ -33,3 +93,186 @@ func TestTransactionContext_Generate(t *testing.T) {
...
+				}
+			}
+		})
+	}
+}
+
+func TestTransactionContextGenerator_GenerateUnsatisfiable(t *testing.T) {
+
+	tests := map[string]struct {
+		setup func(*TransactionContextGenerator, *Assignment)
```
HerbertJordan, line 224: As in the test above, `*Assignment` could be just an `Assignment`.

HerbertJordan (changes requested):
Looks good, just two issues that I discovered in the tests. Please have a look.

simonlechner (approved):

HerbertJordan (approved):

simonlechner (approved):

PR #533, opened 2024-06-12
CT: Add BlobHash rules

facuMH:
This PR adds the rules for BlobHash 

it depends on #532 and #548


go/ct/spc/specification.go
```diff
@@ -1254,6 +1254,57 @@ func getAllRules() []Rule {
...
+		pushes:    1,
+		conditions: []Condition{
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+			HasBlobHash(Param(0)),
+		},
+		parameters: []Parameter{NumericParameter{}},
+		effect: func(s *st.State) {
+			indexU256 := s.Stack.Pop()
+			if indexU256.Gt(NewU256(uint64(len(s.TransactionContext.BlobHashes)))) {
+				s.Stack.Push(NewU256(0))
```
LuisPH3, line 1272: I am not sure if this is correctly modeled. If the condition is that a blobhash must exist, then the effect shall not provide the option to have an out of range case. 

I am not sure what should be doing instead, but according to the condition, this branch should never be executed.

LuisPH3 (changes requested):
I am concerned about the effect of the rule. I am not sure if this happen before and there is already a pattern to solve the issue.

LuisPH3 (dismissed):

go/ct/spc/specification.go
```diff
@@ -1254,6 +1254,53 @@ func getAllRules() []Rule {
...
+		pops:      1,
+		pushes:    1,
+		conditions: []Condition{
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+			HasBlobHash(Param(0)),
+		},
+		parameters: []Parameter{NumericParameter{}},
+		effect: func(s *st.State) {
+			indexU256 := s.Stack.Pop()
+			s.Stack.Push(NewU256FromBytes(s.TransactionContext.BlobHashes[indexU256.Uint64()][:]...))
```
HerbertJordan, line 1271: Maybe add a comment explaining why no overflow-check for the popped parameter is required. It took me a while to realize that the `HasBlobHash(..)` condition is making sure that there is no overflow.

HerbertJordan (approved):
Looks good, thanks.

simonlechner (dismissed):

simonlechner (approved):

go/ct/spc/specification.go
facuMH: Well, I don't think a comment is needed since it is the condition. if this were to fail because of an overflow, a test should fail for the condition and the transaction generator before reaching this point.

PR #534, opened 2024-06-12
Containerize build environment.

LuisPH3:
tl;dr:  This PR adds tools to prevent the famous "but it works in my machine".

This feature allows using different platforms and Toolchains to compile the project. At this stage, an initial Dockerfile is provided in the CI folder to describe an environment capable of building and running tests for the Tosca project. 
The workflow is purely optional.

Using this feature enables: 
- Containerized Jenkins Builds. Where the development environment is described in the git repository and does not depend on any particular server setup.
- Multiple simultaneous platform definitions.
- Different branches with different Toolchains, all of them building correctly the CI. This is important to prevent loosing CI support during Toolchain migrations. 
- Containerized development, using a number of IDEs that support them: https://containers.de

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,12 @@
 // this software will be governed by the GNU Lesser General Public License v3.
 
 pipeline {
-    agent { label 'quick' }
+    agent {
+        dockerfile {
+            filename 'CI/Dockerfile.build'
+            label 'quick'
+        }
+    }
 
     options {
         timestamps()
@@ -50,7 +55,7 @@ pipeline {
 
         stage('Check Go sources formatting') {
             steps {
-                sh "gofmt -s -d go"
+                sh 'gofmt -s -d go'
             }
         }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,6 +35,16 @@ pipeline {
             }
         }
 
+        stage('check environment') {
+            steps {
+                sh 'echo $PATH'
+                sh 'echo $GOPATH'
+                sh 'echo $GOCACHE'
+                sh 'echo $GOTMPDIR'
+                sh 'go version'
+                sh 'whereis gofmt'
+            }
+        }
         stage('Checkout code') {
             steps {
                 sh 'git submodule update --init --recursive'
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -1,13 +1,19 @@
...
+    }
 
     options {
         timestamps ()
         timeout(time: 2, unit: 'HOURS')
     }
 
     environment {
-        GOROOT = '/usr/lib/go-1.21/'
+        GOROOT = '/usr/lib/go-1.22/'
```
wsodsong, line 16: This line may not be needed. We installed a specific Go version in a custom directory on our vm, so we have to specify `GOROOT`.

LuisPH3: The point of this PR is to run the pipeline inside a container defined by the developers.  See line 4.
No tool in the host machine is to be used other than docker itself. This means the developers can install or remove versions of the toolchain without influencing other builds.

wsodsong: Sorry bad explanation. I wanted to point out that `/usr/lib/go-1.21/` is a custom directory specific to the host machine. The directory may or may not exist in a docker container. Setting `GOROOT` to `usr/lib/go-1.22/`  is most likely incorrect.

LuisPH3: I am a little concerned about this line, but because of a different reason: 

I am confident that /usr/lib/go-1.21/ exists in the environment because docker is installed in the container image in jenkinsfile line 10.  The container will run in a dedicated filesystem, where the only visible folder from the host is the source code checkout. No other system or home folder from the host will be visible. Only the tools described in the jenkinsfile will be installed (in the default system paths). 

On the other hand: I copied this line from the current Jenkinsfile, which uses the current VM system, and this worries me, because it has been working fine so far. We should fix that, what do you propose?

wsodsong: I tested creating the image on my local machine. It was my misunderstanding. Yes, `usr/lib/go-1.22` exists. I also noticed when running `go env` that `GOROOT` was automatically set to `usr/lib/go-1.22` on my container. The line may not be necessary. Can you try quickly by removing the line and see if the test still run successfully?

LuisPH3:
This change requires some time resources that we do not have right now. The development can be resumed if the changes are needed

PR #535, opened 2024-06-12
CT: Add hash256

facuMH:
This PR adds an immutable `hash256` type, analogous to `Bytes` to improve performance on state operations.

For instance in `Clone`
```
facu@:~/Tosca$ benchstat oldHashClone.log newHashClone.log
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/ct/st
cpu: AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
                    │ oldHashClone.log │          newHashClone.log           │
                    │      sec/op      │   sec/op     vs base                │
State_CloneState-16        9.839µ ± 5%   7.832µ ± 4%  -20.40% (p=0.000 n=25)

                    │ oldHashClone.log │           newHashClone.log           │
                    │       B/op       │     B/op      vs base                │
State_CloneState-16       41.67Ki ± 0%   32.91Ki ± 0%  -21.02% (p=0.000 n=25)

                    │ oldHashClone.log │        newHashClone.log        │
                    │    allocs/op     │ allocs/op   vs base            │
State_CloneState-16         13.00 ± 0%   13.00 ± 0%  ~ (p=1.000 n=25) ¹
¹ all samples are equal
```

And for `Eq`
```
facu@:~/Tosca$ benchstat oldHashEq.log newHashEq.log
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/ct/st
cpu: AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
                 │ oldHashEq.log │           newHashEq.log            │
                 │    sec/op     │   sec/op     vs base               │
State_EqState-16    1005.0n ± 3%   788.2n ± 6%  -21.57% (p=0.000 n=25)

                 │ oldHashEq.log │         newHashEq.log          │
                 │     B/op      │    B/op     vs base            │
State_EqState-16      0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=25) ¹
¹ all samples are equal

                 │ oldHashEq.log │         newHashEq.log          │
                 │   allocs/op   │ allocs/op   vs base            │
State_EqState-16      0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=25) ¹
¹ all samples are equal
```


LuisPH3:
I had an idea about this, but I don't know if it is possible or clean.

If the conditions do not contain any constraint for the `recentBlockHashes`  you may leave it nill, fill it to zeroes when translating into the VM state.
 
The blockhash history is a read only construct from the point of view of the VM, I am not sure if we even need to compare the state after calling the VM. 

What do you guys think?

go/ct/common/hash256.go
```diff
@@ -0,0 +1,100 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+	"pgregory.net/rand"
+)
+
+// Hash256 is an immutable slice of Hash256 that can be trivially cloned.
+type Hash256 struct {
+	data [256]vm.Hash
```
HerbertJordan, line 25: Objects of this struct may be immutable, but it is not cheap to compare or copy.

The memory layout of this type is 256 vm.Hash values next to each other. When you compare them, all 8 KiB of data is compared, and when you assign them 8 KiB of data is copied.

To make it light-weight, the object should contain a pointer to the data. This pointer can be trivially cloned, restored, and in comparisons you can first check the pointer to the data. If it is the same, there is no need to compare 8 KiB. If it is different, you need to compare the data. But the idea is to have it unlikely that in this case the data is actually equivalent, and the comparison will abort after a few bytes.

HerbertJordan (changes requested):

go/ct/common/hash256.go
facuMH: changes applied, benchmarks rerun.

go/ct/common/hash256_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestHash256_CanBeJsonEncoded(t *testing.T) {
+
+	zeroHashes := "["
+	for i := 0; i < 256; i++ {
+		zeroHashes += hash0
+		zeroHashes += ","
+	}
+	zeroHashes = zeroHashes[:len(zeroHashes)-1] + "]"
+
+	oneHashes := "[" + hash1 + ","
```
simonlechner, line 59:
```suggestion
	oneHash := "[" + hash1 + ","
```

go/ct/common/hash256_test.go
```diff
@@ -0,0 +1,98 @@
...
+	b2 := b1
+
+	if !b1.Equal(b2) {
+		t.Errorf("assigned value is not equal: %v vs %v", b1, b2)
+	}
+}
+
+const (
+	hash0 = "[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]"
+	hash1 = "[1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]"
```
simonlechner, line 47: Since they are only used in the function below their definition could be moved inside the function.

go/ct/common/hash256.go
```diff
@@ -0,0 +1,116 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+	"pgregory.net/rand"
+)
+
+// Hash256 is an immutable array of Hash256 that can be trivially cloned.
+type Hash256 struct {
+	data *[256]vm.Hash
```
LuisPH3, line 25: just a go question, this is immutable because the member is private and all getters return a copy by value, right?  no const data in go?

go/ct/common/hash256.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+	"pgregory.net/rand"
+)
+
+// Hash256 is an immutable array of Hash256 that can be trivially cloned.
+type Hash256 struct {
```
LuisPH3, line 24: nitpick: The type contains more than one hash, name should be plural. 
Comment should say array of 256 Hashes as well.
Why not BlockHashHistory or LastBlockHashes?

go/ct/common/hash256.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// Copy the slice into the Hash256 array
+	h.data = &hashes
+	return nil
+}
+
+func (b Hash256) Get(index uint64) vm.Hash {
+	if index >= 256 {
+		panic("index out of range")
```
LuisPH3, line 107: Is this really a panic? Of course, the caller side cannot call this function out of range, but If we ever reuse this function somewhere else, the reviewer will not see this error, and we will hide the panic under a combination of runtime values. If instead you return an error and handle it in the adapter properly, this code will never panic.

LuisPH3 (changes requested):
good speedup.
please rename the hash256 abstraction.

go/ct/common/hash256.go
facuMH: I think those names convey meaning of a particular use, while this is just an immutable array. I would rather call the type `Hashes256` and fix the comment, rather than talk about blocks here.

go/ct/common/hash256.go
facuMH: exactly

go/ct/common/hash256_test.go
```diff
@@ -0,0 +1,115 @@
...
+package common
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+)
+
+func TestHash256_EqualWhenContainingSameContent(t *testing.T) {
```
LuisPH3, line 22: sorry, both filename and the unit tests names are out of sync now with the type name.
(I really don't like to write 256 in the type name. But I guess this is my problem ;) )

go/ct/st/serialization.go
```diff
@@ -78,7 +78,7 @@ type stateSerializable struct {
 	CallJournal           *CallJournal
 	HasSelfDestructed     bool
 	SelfDestructedJournal []serializableSelfDestructEntry
-	RecentBlockHashes     [256]vm.Hash
+	RecentBlockHashes     Hashes256
```
LuisPH3, line 81: Recent block hashes is a good name.

go/ct/spc/specification.go
```diff
@@ -908,7 +908,9 @@ func getAllRules() []Rule {
 		effect: func(s *st.State) {
 			targetBlockNumber := s.Stack.Pop()
 			index := s.BlockContext.BlockNumber - targetBlockNumber.Uint64()
-			s.Stack.Push(NewU256FromBytes(s.RecentBlockHashes[index-1][:]...))
+			// we don't need to check for index ∈ [0..255] because InRange256FromCurrentBlock already does that
```
LuisPH3, line 911: until we change the implementation and we silently break this code.
no catch all, no ignore err, plz

LuisPH3 (changes requested):

go/ct/spc/specification.go
LuisPH3: As discussed offline, CT does not run in client and cannot produce failure there. It is kind of ok to ignore it here.

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const ErrorIndexOutOfRange = ConstErr("index out of range")
+
+// Hashes256 is an immutable array of 256 vm.Hash that can be trivially cloned/compared.
+type Hashes256 struct {
+	data *[256]vm.Hash
+}
+
+func NewHash256(hashes ...vm.Hash) Hashes256 {
```
LuisPH3, line 30: name :(

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	for i := 0; i < 256; i++ {
+		if b.data[i] != other.data[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func Randomhash256(rnd *rand.Rand) Hashes256 {
```
LuisPH3, line 51: name

go/ct/common/hashes256_test.go
```diff
@@ -0,0 +1,114 @@
...
+package common
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+)
+
+func TestHash256_EqualWhenContainingSameContent(t *testing.T) {
```
LuisPH3, line 22: name

go/ct/common/hashes256_test.go
```diff
@@ -0,0 +1,114 @@
...
+	if !b1.Equal(b2) {
+		t.Errorf("instances are not equal, got %v and %v", b1, b2)
+	}
+
+	if b1.Equal(b3) {
+		t.Errorf("instances are equal, got %v and %v", b1, b3)
+	}
+}
+
+func TestHash256_AssignmentProducesEqualValue(t *testing.T) {
```
LuisPH3, line 36: name

go/ct/common/hashes256_test.go
```diff
@@ -0,0 +1,114 @@
...
+	b1 := NewHash256(vm.Hash{1})
+	b2 := b1
+
+	if !b1.Equal(b2) {
+		t.Errorf("assigned value is not equal: %v vs %v", b1, b2)
+	}
+}
+
+func TestHash256_CanBeJsonEncoded(t *testing.T) {
```
LuisPH3, line 45: name

go/ct/common/hashes256_test.go
```diff
@@ -0,0 +1,114 @@
...
+				t.Fatalf("failed to restore Hash256: %v", err)
+			}
+			if !test.hash256.Equal(restored) {
+				t.Errorf("unexpected restored value, wanted %v, got %v", test.hash256, restored)
+			}
+		})
+	}
+}
+
+func TestHash256_Get(t *testing.T) {
```
LuisPH3, line 99: name

go/ct/common/hashes256_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	oneHash := vm.Hash{1}
+	hash256 := NewHash256(oneHash)
```
LuisPH3, line 102: variable name

LuisPH3 (changes requested):
Unfortunately, this not quite complete. Some types and files are renamed, but unit tests, constructor and some variables keep the old name. 
This is getting quite confusing now. I have to encourage you to find a better name.

plz check comments for mismatches or confusing uses.

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+	"pgregory.net/rand"
+)
+
+// ImmutableHashArray is an immutable array of 256 vm.Hash that can be trivially cloned/compared.
```
HerbertJordan, line 23: It is not true that it can be "trivially compared".

go/ct/common/hashes256_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+)
+
+func TestImmutableHashArray_EqualWhenContainingSameContent(t *testing.T) {
+	b1 := NewImmutableHashArray(vm.Hash{1})
+	b2 := b1
+	b3 := NewImmutableHashArray(vm.Hash{2})
+
+	if !b1.Equal(b2) {
```
HerbertJordan, line 27: The test is named "EqualWhenContainngSameContent" but what is tested is only the case where the content is the same pointer. You should also make sure that two independent instances created by two distinct calls to `NewImmutableHashArray` with the same arguments are equal.

go/ct/common/hashes256.go
HerbertJordan, line 1: As remarked by Luis, the file has not been aligned with the new name.

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if b.data == nil || other.data == nil {
+		return false
+	}
+	for i := 0; i < 256; i++ {
+		if b.data[i] != other.data[i] {
+			return false
+		}
+	}
+	return true
```
HerbertJordan, line 46: Shorter:
```suggestion
	return b.data == other.data || (b.data != nil && other.data != nil && *b.data == *other.data)
```

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+	return bytes
+}
+
+func (b ImmutableHashArray) FromBytes(bytes []byte) ImmutableHashArray {
+	hashes := ImmutableHashArray{}
+	for i := 0; i < 256; i++ {
+		copy(hashes.data[i][:], bytes[i*32:(i+1)*32])
+	}
+	return hashes
+}
```
HerbertJordan, line 72: These two functions are nowhere used, needed, nor tested.

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (b ImmutableHashArray) ToHash() [256]vm.Hash {
+	return [256]vm.Hash(*b.data)
+}
```
HerbertJordan, line 76: This function is also nowhere used nor tested.

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+func (b ImmutableHashArray) FromBytes(bytes []byte) ImmutableHashArray {
+	hashes := ImmutableHashArray{}
+	for i := 0; i < 256; i++ {
+		copy(hashes.data[i][:], bytes[i*32:(i+1)*32])
+	}
+	return hashes
+}
+
+func (b ImmutableHashArray) ToHash() [256]vm.Hash {
+	return [256]vm.Hash(*b.data)
```
HerbertJordan, line 75:
```suggestion
	return *b.data
```

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (b ImmutableHashArray) String() string {
+	return fmt.Sprintf("%x", b.data)
+}
+
+func (b ImmutableHashArray) MarshalJSON() ([]byte, error) {
+	return json.Marshal(b.data)
+}
+
+// Implement the UnmarshalJSON method for ImmutableHashArray
```
HerbertJordan, line 86: The comment is not formatted according to the general Go convention and does not provide any information that would not be obvious from the existence of this function.

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// Copy the slice into the ImmutableHashArray array
+	h.data = &hashes
+	return nil
+}
+
+// Get returns the hash at the given index or panics if out of range
+func (b ImmutableHashArray) Get(index uint64) vm.Hash {
+	return vm.Hash(b.data[index])
```
HerbertJordan, line 107: b.data might be `nil`, which leads to a panic;

go/ct/common/hashes256.go
```diff
@@ -0,0 +1,114 @@
...
+	for i := 0; i < 256; i++ {
+		bytes = append(bytes, b.data[i][:]...)
+	}
+	return bytes
+}
+
+func (b ImmutableHashArray) FromBytes(bytes []byte) ImmutableHashArray {
+	hashes := ImmutableHashArray{}
+	for i := 0; i < 256; i++ {
+		copy(hashes.data[i][:], bytes[i*32:(i+1)*32])
```
HerbertJordan, line 69: This most likely crashes because `hashes.data` is `nil`.

HerbertJordan (changes requested):
There are still lots of things to improve in this PR. See comments.

LuisPH3 (approved):

go/ct/common/immutable_hash_array_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}{
+		"empty": {
+			NewImmutableHashArray(), NewImmutableHashArray(), true,
+		},
+		"single": {
+			NewImmutableHashArray(vm.Hash{1}), NewImmutableHashArray(vm.Hash{1}), true,
+		},
+		"diff": {
+			NewImmutableHashArray(vm.Hash{1}), NewImmutableHashArray(vm.Hash{2}), false,
+		},
```
HerbertJordan, line 37: There should be test cases with default values where the `data` field is `nil`.

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Get returns the hash at the given index or panics if out of range
+func (b ImmutableHashArray) Get(index uint64) vm.Hash {
+	if b.data == nil {
+		return vm.Hash{}
+	}
+	return vm.Hash(b.data[index])
+}
+
+func GetRandomHash(rnd *rand.Rand) vm.Hash {
```
HerbertJordan, line 81: nit: nobody would probably look for a function like this here. Consider moving it into its own `hash.go` file in this directory.

go/ct/common/immutable_hash_array_test.go
```diff
@@ -0,0 +1,127 @@
...
+		oneHash += ","
+	}
+	oneHash = oneHash[:len(oneHash)-1] + "]"
+
+	tests := map[string]struct {
+		hashes  ImmutableHashArray
+		encoded string
+	}{
+		"empty": {
+			NewImmutableHashArray(vm.Hash{}), zeroHash,
```
HerbertJordan, line 83: You should also test a default value with `data == nil`.

go/ct/common/immutable_hash_array_test.go
```diff
@@ -0,0 +1,127 @@
...
+				t.Fatalf("failed to restore ImmutableHashArray: %v", err)
+			}
+			if !test.hashes.Equal(restored) {
+				t.Errorf("unexpected restored value, wanted %v, got %v", test.hashes, restored)
+			}
+		})
+	}
+}
+
+func TestImmutableHashArray_Get(t *testing.T) {
```
HerbertJordan, line 112: Also here, the default value should be tested.

Note, the way you implemented right now, the `ImmutableHashArray` these two instances behave differently:
```
a := ImmutableHashArray{}
b := NewImmutableHashArray{}
a.Get(256) // => returns 0
b.Get(256)  // => panic
```

HerbertJordan (changes requested):

LuisPH3 (approved):

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return nil
+}
+
+// Get returns the hash at the given index or panics if out of range
+func (b ImmutableHashArray) Get(index uint64) vm.Hash {
+	if b.data == nil && index < 256 {
+		return vm.Hash{}
+	} else if index >= 256 {
+		panic(fmt.Sprintf("index out of range: %d", index))
```
LuisPH3, line 89: nit: this panic is redundant, go checks slices anyway.

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,92 @@
...
+	if b.data == nil {
+		return *other.data == [256]vm.Hash{}
+	}
+	if other.data == nil {
+		return *b.data == [256]vm.Hash{}
+	}
+	return b.data == other.data || (b.data != nil && other.data != nil && *b.data == *other.data)
+}
+
+func RandomImmutableHashArray(rnd *rand.Rand) ImmutableHashArray {
```
LuisPH3, line 45: nit: this is a constructor, NewRandomImmutableHashArray

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,92 @@
...
+	} else {
+		h.data = &hashes
+	}
+	return nil
+}
+
+// Get returns the hash at the given index or panics if out of range
+func (b ImmutableHashArray) Get(index uint64) vm.Hash {
+	if b.data == nil && index < 256 {
+		return vm.Hash{}
```
LuisPH3, line 87: If we panic with an out-of-bounds access, we should not silently allow calling an object that is wrongly initialized. 
for consistency, we should panic here as well.

LuisPH3 (changes requested):
Please check the error handling of the Get function.

go/ct/common/immutable_hash_array.go
facuMH: discussed offline, will add documentation

go/ct/common/immutable_hash_array.go
facuMH: discussed offline, will add documentationv

LuisPH3 (approved):

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,94 @@
...
+	} else if index >= 256 {
+		// if data is nil, we still want to return ouf of range error
+		panic(fmt.Sprintf("index out of range: %d", index))
+	}
+	return vm.Hash(b.data[index])
```
HerbertJordan, line 93:
```suggestion
	return b.data[index]
```

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,94 @@
...
+		h.data = &hashes
+	}
+	return nil
+}
+
+// Get returns the hash at the given index or panics if out of range
+func (b ImmutableHashArray) Get(index uint64) vm.Hash {
+	if b.data == nil && index < 256 {
+		return vm.Hash{}
+	} else if index >= 256 {
```
HerbertJordan, line 89: nit: if a branch ends with a return, you can skip the `else`

go/ct/st/state.go
```diff
@@ -457,9 +458,13 @@ func (s *State) Diff(o *State) []string {
...
 
-	for i, want := range s.RecentBlockHashes {
-		if want != o.RecentBlockHashes[i] {
-			res = append(res, fmt.Sprintf("Different block number hash at index %d: %x vs %x", i, want, o.RecentBlockHashes[i]))
+	if !s.RecentBlockHashes.Equal(o.RecentBlockHashes) {
+		for i := 0; i < 256; i++ {
+			want := s.RecentBlockHashes.Get(uint64(i))
+			got := o.RecentBlockHashes.Get(uint64(i))
+			if want != got {
+				res = append(res, fmt.Sprintf("Different block number hash at index %d: %x vs %x", i, want, got))
```
HerbertJordan, line 466: nit:
```suggestion
				res = append(res, fmt.Sprintf("Different block hash at index %d: %x vs %x", i, want, got))
```

HerbertJordan (approved):
Looks good now, thanks for the improvements.

LuisPH3 (dismissed):
alright!

go/ct/common/immutable_hash_array_test.go
```diff
@@ -0,0 +1,144 @@
...
+		zeroHash += ","
+	}
+	zeroHash = zeroHash[:len(zeroHash)-1] + "]"
+
+	oneHash := "[" + hash1 + ","
+	for i := 0; i < 255; i++ {
+		oneHash += hash0
+		oneHash += ","
+	}
+	oneHash = oneHash[:len(oneHash)-1] + "]"
```
simonlechner, line 80:
```suggestion
	zeroHash := "[" + strings.Repeat(hash0+",", 255)
	zeroHash += hash0 + "]"

	oneHash := "[" + hash1 + ","
	oneHash += strings.Repeat(hash0+",", 254)
	oneHash += hash0 + "]"
```

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Check the length of the ImmutableHashArray data
+	if len(hashes) != len(h.data) {
+		return fmt.Errorf("invalid ImmutableHashArray length")
+	}
+
+	// Copy the slice into the ImmutableHashArray array
```
simonlechner, line 76:
```suggestion
	// Copy the slice into the ImmutableHashArray
```

simonlechner (dismissed):
Looks good, just some picky suggestions.

LuisPH3 (approved):

go/ct/common/immutable_hash_array.go
```diff
@@ -0,0 +1,95 @@
...
+	if b.data == nil {
+		return *other.data == [256]vm.Hash{}
+	}
+	if other.data == nil {
+		return *b.data == [256]vm.Hash{}
+	}
+	return b.data == other.data || (b.data != nil && other.data != nil && *b.data == *other.data)
+}
+
+func NewRandomImmutableHashArray(rnd *rand.Rand) ImmutableHashArray {
```
HerbertJordan, line 46: nit: keep factory functions grouped

HerbertJordan (approved):

HerbertJordan (approved):

LuisPH3 (approved):

Issue #466, opened 2024-05-21
Implement TLOAD and TSTORE instructions


HerbertJordan:
As part of the Cancun revision, the transient storage instructions TLOAD and TSTORE have been added ([EIP-1153](https://eips.ethereum.org/EIPS/eip-1153)).

To support this instruction in Tosca, the following steps are required:
- update geth dependency to a version that covers Shanghai (see #394)
- add TLOAD and TSTORE instruction support to LFVM + unit tests
- add TLOAD and TSTORE instruction support to the evmzero + unit tests
- add TLOAD and TSTORE specification to the CT spec

PR #536, opened 2024-06-13
LFVM: Add transient storage operations

simonlechner:
This PR adds transient storage operations TLOAD and TSTORE to the LFVM.

Contributes to #466.


facuMH (approved):

go/vm/lfvm/opcode.go
```diff
@@ -144,6 +144,8 @@ const (
 	// Storage
 	SLOAD
 	SSTORE
+	TLOAD
+	TSTORE
```
LuisPH3, line 148: I think this file was as well the op_to_string map, does it not?

go/vm/lfvm/interpreter.go
```diff
@@ -793,7 +797,8 @@ func isWriteInstruction(opCode OpCode) bool {
 		1<<(LOG4-SSTORE) |
 		1<<(CREATE-SSTORE) |
 		1<<(CREATE2-SSTORE) |
-		1<<(SELFDESTRUCT-SSTORE)
+		1<<(SELFDESTRUCT-SSTORE) |
+		1<<(TSTORE-SSTORE)
```
LuisPH3, line 801: this is obscure.... can you explain to me what this does?

LuisPH3 (changes requested):
looks good, but please check if there is a missing to_string entry

go/vm/lfvm/opcode.go
simonlechner: Missed this one, thank you.

go/vm/lfvm/interpreter.go
simonlechner: This also took me a few minutes to understand, it is used to check whether a opcodes writes and is therefore not allowed in static mode. 
It uses the fact that SSTORE is the first opcode that is not allowed in static mode. For each operation listed the bit mask is set using the difference to SSTORE as the offset. In the line of the return it is then checked whether the opcode is smaller than SSTORE and the bit mask is set at the given offset.
I am pretty sure this is a go work around for better performance, but I am not sure.

LuisPH3: yes, this looks like a pretty brainy optimization. Funny that there is a benchmark, but no unit test ;)

LuisPH3 (approved):

go/vm/lfvm/interpreter.go
HerbertJordan: Jup, I remember building this with Jan when all of this was just a early prototype. This bit-masking is way faster than a lookup in a map or slice.

Anyway: I think `TSTORE` should not be in this list. TSTORE has no lasting effect, and is thus not a `WriteInstruction` that should fail in a static context. Running TSTOREs in static contexts is fine (at least in my understanding, and evm.codes also does not list a error condition in case of being used in a read-only context).

HerbertJordan (approved):
Looks good, just the inclusion in the write-instruction set should be checked.

go/vm/lfvm/interpreter.go
simonlechner: I responded on the evmzero PR with links to the geth implementation and execution spec implementation.

facuMH (approved):

PR #540, opened 2024-06-14
CT: Add transient storage operations

### LINKED ISSUE #466: Implement TLOAD and TSTORE instructions

simonlechner:
This PR adds transient storage operations TLOAD and TSTORE to the CT.

Fixes #466.


go/ct/gen/transient_test.go
```diff
@@ -0,0 +1,88 @@
...
+	rnd := rand.New(0)
+	generator := NewTransientGenerator()
+	generator.BindSet(v1)
+	generator.BindNotSet(v1)
+
+	_, err := generator.Generate(assignment, rnd)
+	if err == nil {
+		t.Errorf("Expected error during generation, but got nil")
+	}
+}
```
HerbertJordan, line 88: A missing test case that is failing right now:
```
func TestTransient_FindMeABetterName(t *testing.T) {
	v1 := Variable("v1")
	v2 := Variable("v2")
	assignment := Assignment{}
	assignment[v1] = common.NewU256(42)
	assignment[v2] = common.NewU256(42)

	rnd := rand.New(0)
	generator := NewTransientGenerator()
	generator.BindSet(v1)
	generator.BindNotSet(v2)

	_, err := generator.Generate(assignment, rnd)
	if err == nil {
		t.Errorf("Expected error during generation, but got nil")
	}
}
```

go/ct/gen/transient.go
```diff
@@ -0,0 +1,135 @@
...
+	var parts []string
+
+	for key, set := range t.set {
+		if set {
+			parts = append(parts, fmt.Sprintf("set(%v)", key))
+		} else {
+			parts = append(parts, fmt.Sprintf("notSet(%v)", key))
+		}
+	}
+	return "{" + strings.Join(parts, ",") + "}"
```
HerbertJordan, line 134: Please sort the parts before joining them to produce a deterministic output (note: maps have a random order).

HerbertJordan, line 134: Reading `{set(v1),set(v2),notSet(v3)}` in the constraints is likely not very insightful for anybody to study or debug this code. Also, those constraints might eventually serve as the input for SMT solvers. Hence, we should try to stick to more accurate prints.

How about: `transient[v1]≠0 ∧ transient[v2]≠0 ∧ transient[v1]=0` ?

HerbertJordan, line 134: When the generator is `unsatisfiable` the print should reflect this.

Following the formula structure above, an `unsatisfiable` generator state should print `false` (a formula that can not be satisfied).

HerbertJordan (changes requested):

go/ct/gen/state_test.go
```diff
@@ -342,6 +342,7 @@ func TestStateGenerator_ClonesAreIndependent(t *testing.T) {
 		"stack={2≤size≤200}",
 		"memory={}",
 		"storage={}",
+		"transient={}",
```
facuMH, line 345: in at least one of the states, add a condition, so that we can check it is properly printed.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (t *TransientGenerator) BindSet(key Variable) {
+	if slices.Contains(maps.Keys(t.set), key) {
+		if !t.set[key] {
+			t.unsatisfiable = true
+		}
+	} else {
+		t.set[key] = true
+	}
```
facuMH, line 43:
```suggestion
	if value, exist := t.set[key], exist {
		if !value {
			t.unsatisfiable = true
			return
		}
	}
	t.set[key] = true
```

go/ct/gen/transient.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (t *TransientGenerator) BindNotSet(key Variable) {
+	if slices.Contains(maps.Keys(t.set), key) {
```
facuMH, line 47: same as above.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,135 @@
...
+			t.unsatisfiable = true
+		}
+	} else {
+		t.set[key] = false
+	}
+}
+
+func (t *TransientGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Transient, error) {
+	if t.unsatisfiable {
+		return nil, fmt.Errorf("%w, conflicting set/unset constraints", ErrUnsatisfiable)
```
facuMH, line 58: there is already an existing `Err.Unsatisfiable`, maybe use that one ?

go/ct/gen/transient.go
```diff
@@ -0,0 +1,135 @@
...
+			key = getUnusedKey()
+		}
+
+		if set {
+			transient.SetStorage(key, common.RandU256(rnd))
+			assignment[variable] = key
+		} else {
+			transient.DeleteStorage(key)
+			assignment[variable] = key
+		}
```
facuMH, line 95:
```suggestion
		if set {
			transient.SetStorage(key, common.RandU256(rnd))
		} else {
			transient.DeleteStorage(key)
		}
		assignment[variable] = key
```

go/ct/gen/transient_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	if transient == nil {
+		t.Errorf("Expected transient storage to be generated, but got nil")
+	}
+	if len(transient.GetStorageKeys()) == 0 {
+		t.Errorf("Expected transient storage to be non-empty, but got empty")
+	}
+}
+
+func TestTransient_SetConstraintIsEnforced(t *testing.T) {
```
facuMH, line 38: `SetConstraintIsEnforced` and `UnSetConstraintIsEnforced` can be merged into a test with a table

go/ct/rlz/conditions.go
```diff
@@ -578,6 +578,100 @@ func (c *storageConfiguration) String() string {
...
+func IsTransientNotSet(key BindableExpression[U256]) Condition {
+	return &isTransientNotSet{key}
+}
+
+func (c *isTransientNotSet) Check(s *st.State) (bool, error) {
+	key, err := c.key.Eval(s)
+	if err != nil {
+		return false, err
+	}
+	return s.Transient.IsNotSet(key), nil
```
facuMH, line 644:
```suggestion
    return !isTransientSet{c.key}.Check(s), nil
```

go/ct/rlz/conditions.go
```diff
@@ -578,6 +578,100 @@ func (c *storageConfiguration) String() string {
...
+		if isNotSet {
+			generator.BindIsTransientNotSet(key)
+		} else {
+			generator.BindIsTransientSet(key)
+		}
+	}
+	return []TestValue{
+		NewTestValue(property, domain, true, restrict),
+		NewTestValue(property, domain, false, restrict),
+	}
```
facuMH, line 668:
```suggestion
    return isTransientStorageSet{c.key}.GetTestValues()
```

go/ct/rlz/conditions.go
```diff
@@ -578,6 +578,100 @@ func (c *storageConfiguration) String() string {
...
+}
+
+func (c *isTransientSet) String() string {
+	return fmt.Sprintf("set(%v)", c.key)
```
facuMH, line 625:
```suggestion
	return fmt.Sprintf("is transient set(%v)", c.key)
```

go/ct/rlz/conditions.go
```diff
@@ -578,6 +578,100 @@ func (c *storageConfiguration) String() string {
...
+		}
+	}
+	return []TestValue{
+		NewTestValue(property, domain, true, restrict),
+		NewTestValue(property, domain, false, restrict),
+	}
+}
+
+func (c *isTransientNotSet) String() string {
+	return fmt.Sprintf("not set(%v)", c.key)
```
facuMH, line 672:
```suggestion
	return fmt.Sprintf("not transient set(%v)", c.key)
```

go/ct/spc/specification.go
```diff
@@ -719,6 +719,129 @@ func getAllRules() []Rule {
...
+		},
+	})...)
+
+	rules = append(rules, Rule{
+		Name: "tload_pre_cancun",
+		Condition: And(
+			RevisionBounds(R07_Istanbul, R11_Paris),
+			Eq(Status(), st.Running),
+			Eq(Op(Pc()), TLOAD),
+			Ge(Gas(), 100),
```
facuMH, line 768: gas, and stacksize should not be conditions here, since it should fail regardless of those values.

go/ct/spc/specification.go
```diff
@@ -719,6 +719,129 @@ func getAllRules() []Rule {
...
+			IsTransientSet(Param(0)),
+		},
+		effect: func(s *st.State) {
+			key := s.Stack.Pop()
+			value := s.Stack.Pop()
+			s.Transient.SetStorage(key, value)
+		},
+	})...)
+
+	rules = append(rules, rulesFor(instruction{
```
facuMH, line 798: duplicated rule ?

go/ct/spc/specification.go
```diff
@@ -719,6 +719,129 @@ func getAllRules() []Rule {
...
+
+	rules = append(rules, Rule{
+		Name: "tstore_pre_cancun",
+		Condition: And(
+			RevisionBounds(R07_Istanbul, R11_Paris),
+			Eq(Status(), st.Running),
+			Eq(Op(Pc()), TSTORE),
+			Ge(Gas(), 100),
+			Ge(StackSize(), 2),
+			Eq(ReadOnly(), false),
```
facuMH, line 827: same as before, remove this conditions so that the rules cover more states that should also fail.

go/ct/spc/specification.go
```diff
@@ -719,6 +719,129 @@ func getAllRules() []Rule {
...
+	})
+
+	rules = append(rules, Rule{
+		Name: "tstore_read_only",
+		Condition: And(
+			AnyKnownRevision(),
+			Eq(Status(), st.Running),
+			Eq(Op(Pc()), TSTORE),
+			Ge(Gas(), 100),
+			Ge(StackSize(), 2),
```
facuMH, line 839: same as the other fail cases.

go/ct/st/transient.go
```diff
@@ -0,0 +1,66 @@
...
+func (t *Transient) Eq(other *Transient) bool {
+	return mapEqualIgnoringZeroValues(t.storage, other.storage)
+}
+
+func (t *Transient) Diff(other *Transient) (res []string) {
+	res = append(res, mapDiffIgnoringZeroValues(t.storage, other.storage, "storage")...)
+	return
+}
+
+// For testing purposes only
```
facuMH, line 63: go documentation, 
```suggestion
// GetStorageKeys is here for testing purposes only
```

go/ct/st/transient_test.go
```diff
@@ -0,0 +1,110 @@
...
+	t1.SetStorage(NewU256(42), NewU256(1))
+
+	for name, change := range tests {
+		t.Run(name, func(t *testing.T) {
+			t2 := t1.Clone()
+			if !reflect.DeepEqual(t1, t2) {
+				t.Errorf("Clone is broken, want %v, got %v", t1, t2)
+			}
+			change(t2)
+			if reflect.DeepEqual(t1, t2) {
```
facuMH, line 45: nit: if you want to go one extra step more, check can be be parametrized and then you could have all 3 tests in one, since the structures of the all 3 are the same, just the check changes.

facuMH (changes requested):
looks good !! but I have some code style suggestions

go/ct/gen/transient.go
simonlechner: It is used, it is just wrapped with more information.

go/ct/spc/specification.go
simonlechner: No, one is for transient storage accesses on already set locations and the other for unset locations.

go/ct/st/transient_test.go
simonlechner: I think those functions are testing different things and it much easier to debug if they are separate.

go/ct/gen/state.go
```diff
@@ -227,6 +229,16 @@ func (g *StateGenerator) BindIsStorageCold(key Variable) {
 	g.storageGen.BindCold(key)
 }
 
+// BindIsTransientSet wraps TransientGenerator.BindSet.
+func (g *StateGenerator) BindIsTransientSet(key Variable) {
```
HerbertJordan, line 233: This function name has some grammar issues ... how would one read a line of code like this:
```
var := Variable("a")
gen.BindIsTransientSet(var)
```
This is not very intuitive. How about something like
```
var := Variable("a")
gen.RestrictToStorageLocationWithNonZeroTransientStorage(var)
```
Granted, it is not the shortest name (maybe you have a better idea), but I would argue that when one is reading this, it is clearer what this line is supposed to do.

In general, I think the naming in the generator package needs a substantial overhaul ...

go/ct/gen/state.go
```diff
@@ -227,6 +229,16 @@ func (g *StateGenerator) BindIsStorageCold(key Variable) {
 	g.storageGen.BindCold(key)
 }
 
+// BindIsTransientSet wraps TransientGenerator.BindSet.
```
HerbertJordan, line 232: nit: this document string is useless since `TransientGenerator.BindSet` is not documented either;

go/ct/gen/transient.go
```diff
@@ -0,0 +1,149 @@
...
+	"strings"
+
+	"golang.org/x/exp/maps"
+	"pgregory.net/rand"
+
+	"github.com/Fantom-foundation/Tosca/go/ct/common"
+	"github.com/Fantom-foundation/Tosca/go/ct/st"
+)
+
+type TransientGenerator struct {
```
HerbertJordan, line 27: `TransientStorageGenerator`? ... a `transient` generator sounds a bit misleading.

go/ct/st/state.go
```diff
@@ -284,6 +288,10 @@ func (s *State) String() string {
 	for k := range s.Storage.warm {
 		write("\t    [%v]\n", k)
 	}
+	write("\tTransient Storage:\n")
+	for k := range s.Transient.storage {
+		write("\t    [%v]\n", k)
```
HerbertJordan, line 293: Should it be printed like regular storage in the form of key/value pairs?

go/ct/st/transient.go
```diff
@@ -0,0 +1,66 @@
...
+	if t.storage == nil {
+		t.storage = make(map[U256]U256)
+		t.storage[key] = value
+	} else {
+		t.storage[key] = value
+	}
+}
+
+func (t *Transient) IsSet(key U256) bool {
+	_, ok := t.storage[key]
```
HerbertJordan, line 34: You either need to check for the `value` to be zero here as well or deal with `zero` values in the `SetStorage` call to be consistent.

go/ct/st/transient.go
```diff
@@ -0,0 +1,66 @@
...
+func (t *Transient) SetStorage(key U256, value U256) {
+	if t.storage == nil {
+		t.storage = make(map[U256]U256)
+		t.storage[key] = value
+	} else {
+		t.storage[key] = value
+	}
+}
+
+func (t *Transient) IsSet(key U256) bool {
```
HerbertJordan, line 33: How about renaming this to "IsEmpty" or "IsZero"?

go/ct/st/transient.go
```diff
@@ -0,0 +1,66 @@
...
+	_, ok := t.storage[key]
+	return ok
+}
+
+func (t *Transient) IsNotSet(key U256) bool {
```
HerbertJordan, line 38: nit: i would delete this function; users should just call `IsSet` (or renamed to `IsEmpty`) and negate the result as needed.

go/ct/st/transient.go
```diff
@@ -0,0 +1,66 @@
...
+	return !t.IsSet(key)
+}
+
+func (t *Transient) GetStorage(key U256) U256 {
+	return t.storage[key]
+}
+
+func (t *Transient) DeleteStorage(key U256) {
```
HerbertJordan, line 46: This should have the same effect as `SetStorage(key, Value{})`.

In the semantic of the transient storage (or the regular Storage), all keys actually always have a value. It is just that most of them are zero and not explicitly stored. However, logically they are all always present. They can not be created nor deleted.

It is more accurate to think about those as a huge array with 2^256 entries where most of them are zero. You would also not define an operation on the array deleting an index position. This is not really possible.

go/ct/st/transient.go
```diff
@@ -0,0 +1,66 @@
...
+	return mapEqualIgnoringZeroValues(t.storage, other.storage)
+}
+
+func (t *Transient) Diff(other *Transient) (res []string) {
+	res = append(res, mapDiffIgnoringZeroValues(t.storage, other.storage, "storage")...)
+	return
+}
+
+// GetStorageKeys is used for testing purposes only
+func (t *Transient) GetStorageKeys() []U256 {
```
HerbertJordan, line 64: If it is for testing only, it should be private.

go/ct/spc/specification.go
```diff
@@ -719,6 +719,129 @@ func getAllRules() []Rule {
...
+		op:        TLOAD,
+		staticGas: 100,
+		pops:      1,
+		pushes:    1,
+		parameters: []Parameter{
+			StorageAccessKeyParameter{},
+		},
+		conditions: []Condition{
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+			IsTransientSet(Param(0)),
```
HerbertJordan, line 734: `IsTransientSet` is not a good name. If somebody knows the data model of the EVM, it is not clear what this should mean. Any ideas for a better name?

go/ct/spc/specification.go
```diff
@@ -719,6 +719,129 @@ func getAllRules() []Rule {
...
+		parameters: []Parameter{
+			StorageAccessKeyParameter{},
+		},
+		conditions: []Condition{
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+			IsTransientNotSet(Param(0)),
+		},
+		effect: func(s *st.State) {
+			key := s.Stack.Pop()
+			value := s.Transient.GetStorage(key)
```
HerbertJordan, line 757: No need to look up the storage. The result must be zero. You can push zero straight on the stack to emphasis this.

HerbertJordan (commented):
I look through most of it but I need to call it a day now. Will continue tomorrow.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,149 @@
...
+		if pos, isBound := assignment[variable]; isBound {
+			key = pos
+		} else {
+			key = getUnusedKey()
+		}
+
+		if set {
+			transient.SetStorage(key, common.RandU256(rnd))
+		} else {
+			transient.DeleteStorage(key)
```
HerbertJordan, line 102: nit: I might be wrong, but this should not be necessary. The `key` to be deleted should at this point never be set before. If it would have, another constraint that got fulfilled by adding the key gets violated by removing it again.

Removing this simplifies the generator and may allow you to remove the `DeleteStorage` function.

HerbertJordan (changes requested):
A key step when implementing a new feature in CT is always to find a proper terminology for the components that get modeled -- and naming is hart.

Consider, for instance, an abstract data type like a stack. A stack is an object that has a `Push` and a `Pop` function, and maybe a `Size`. There may be many implementations for stacks, but the terminology is fixed, and developers get a clear idea what does operations do, without the need to understand that a `Push` function is internally appending something to a linked list, or growing a dynamic array.

In this PR the new concept that got added is the transient storage, commonly referred in this PR as `transient`. As a first step, for this `transient` operations needed to be fixed. You went with `SetStorage`, `GetStorage`, `IsSet`, `IsNotSet`, and `DeleteStorage`. This seems to reflect the chosen implementation, which is using a `map[Key]Value` as well as the intended use case for the types of constraints to be defined on it.

However, the concept in the blockchain that is modeled is not a key/value map. It is actually an array with 2^256 elements, initially initialized with zeros. The `TLOAD` and `TSTORE` operations can read and write to this array. Thus, a `Get` and a `Set` operation would be natural functions defined on the structure of a transient storage. For somebody familiar with this world state concept, it is not so obvious what a `IsSet` or `DeleteStorage` means for transient storage. In the end, those are helper functions that are equivalent to
```
IsSet(a) => Get(a) == 0
```
and
```
DeleteStorage(a) => Set(a,0)
```
However, by offering those explicitly, readers of the code may be mislead assuming that a transient storage is indeed a map, and that there is a difference between a storage location having the value `0` and not being set. As it would be, if the abstract model one should have about transient storage is indeed a `map`, not an `array`.

What I want to say with all of this (tl;dr): language and terminology is important. Being precise can simplify stuff, and reduce complexity.

Consider my comments in this PR in this context. If you agree, maybe you can work through it again and rethink and align the terminology. If not, let me know and I will finish the review for this PR as is.

go/ct/spc/specification.go
facuMH: `HasTransientStorageValue(Param(0))` ?

go/ct/spc/specification.go
facuMH: that is exactly my point, if you want it to fail explicitly because of the revision, you do not care about gas or stacksize.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"golang.org/x/exp/maps"
+	"pgregory.net/rand"
+
+	"github.com/Fantom-foundation/Tosca/go/ct/common"
+	"github.com/Fantom-foundation/Tosca/go/ct/st"
+)
+
+type TransientGenerator struct {
+	set           map[Variable]bool
```
LuisPH3, line 28: I would rather use constraints here or any other name related to constraints, like the other generators.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,149 @@
...
+
+	transient := &st.Transient{}
+
+	// Process set/unset constraints.
+	for variable, set := range t.set {
+
+		// Check different assignments with the same key if they conflict
+		for other, otherSet := range t.set {
+			if assignment[variable] == assignment[other] && set != otherSet {
+				return nil, fmt.Errorf("%w, conflicting set/unset constraints", ErrUnsatisfiable)
```
LuisPH3, line 88: Same error as unsatisfiable, difficult to diagnose

go/ct/gen/transient.go
```diff
@@ -0,0 +1,149 @@
...
+		if set {
+			transient.SetStorage(key, common.RandU256(rnd))
+		} else {
+			transient.DeleteStorage(key)
+		}
+		assignment[variable] = key
+	}
+
+	// Random entries
+	for i := 0; i < rnd.Intn(42); i++ {
```
LuisPH3, line 108: :wink: 42

go/ct/gen/transient.go
simonlechner: I reworked the whole PR and worked on naming, I these changes should make it much more readable.

go/ct/gen/state.go
```diff
@@ -69,6 +70,7 @@ func NewStateGenerator() *StateGenerator {
 		stackGen:              NewStackGenerator(),
 		memoryGen:             NewMemoryGenerator(),
 		storageGen:            NewStorageGenerator(),
+		transientGen:          NewTransientGenerator(),
```
HerbertJordan, line 73: The factory should also be renamed to `NewTransientStorageGenerator`.

go/ct/gen/state.go
```diff
@@ -411,6 +423,12 @@ func (g *StateGenerator) Generate(rnd *rand.Rand) (*st.State, error) {
 		return nil, err
 	}
 
+	// Invoke TransientGenerator
```
HerbertJordan, line 426: nit: name is outdated

go/ct/st/storage.go
```diff
@@ -161,34 +161,27 @@ func (a *Storage) Eq(b *Storage) bool {
 		maps.Equal(a.warm, b.warm)
 }
 
-func (a *Storage) Diff(b *Storage) (res []string) {
-	for key, valueA := range a.current {
-		valueB, contained := b.current[key]
-		if !contained && valueA != NewU256(0) {
-			res = append(res, fmt.Sprintf("Different current entry:\n\t[%v]=%v\n\tvs\n\tmissing", key, valueA))
+func mapDiffIgnoringZeroValues[K comparable](a map[K]U256, b map[K]U256, name string) (res []string) {
```
HerbertJordan, line 164: nice refactoring :+1:

go/ct/st/transient.go
```diff
@@ -0,0 +1,58 @@
...
+	return &TransientStorage{maps.Clone(t.storage)}
+}
+
+func (t *TransientStorage) Eq(other *TransientStorage) bool {
+	return mapEqualIgnoringZeroValues(t.storage, other.storage)
+}
+
+func (t *TransientStorage) Diff(other *TransientStorage) (res []string) {
+	res = append(res, mapDiffIgnoringZeroValues(t.storage, other.storage, "transient storage")...)
+	return
```
HerbertJordan, line 52: same as
```suggestion
	return mapDiffIgnoringZeroValues(t.storage, other.storage, "transient storage")
```

go/ct/st/transient.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// GetStorageKeys is used for testing purposes only
+func (t *TransientStorage) GetStorageKeys() []U256 {
```
HerbertJordan, line 56: nit: For the one use case where this is used a `IsAllZero` would be sufficient. This would eliminate the "key" reference in the name.

go/ct/spc/specification.go
```diff
@@ -717,6 +717,120 @@ func getAllRules() []Rule {
...
+		pops:      2,
+		pushes:    0,
+		parameters: []Parameter{
+			StorageAccessKeyParameter{},
+			NumericParameter{},
+		},
+		conditions: []Condition{
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+			Eq(ReadOnly(), false),
+			BindTransientStorageToNonZero(Param(1)),
```
HerbertJordan, line 783: The key should be `Param(0)`, right?

go/ct/spc/specification.go
```diff
@@ -717,6 +717,120 @@ func getAllRules() []Rule {
...
+		pops:      2,
+		pushes:    0,
+		parameters: []Parameter{
+			StorageAccessKeyParameter{},
+			NumericParameter{},
+		},
+		conditions: []Condition{
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+			Eq(ReadOnly(), false),
+			BindTransientStorageToZero(Param(1)),
```
HerbertJordan, line 804: same here, `Param(0)`, right?

HerbertJordan (changes requested):
I think the improvements look good now. Found a few minor issues though, please have a look.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+type TransientStorageGenerator struct {
+	nonZeroConstraints map[Variable]bool
+	unsatisfiable      bool
+}
+
+func NewTransientStorageGenerator() *TransientStorageGenerator {
+	return &TransientStorageGenerator{nonZeroConstraints: map[Variable]bool{}}
+}
+
+// BindToNonZero binds the transient storage location to a non-zero value.
```
LuisPH3, line 34:
```suggestion
// BindToNonZero binds a location in the transient storage to a non-zero value.
```

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+		if !value {
+			t.unsatisfiable = true
+			return
+		}
+	} else {
+		t.nonZeroConstraints[key] = true
+	}
+}
+
+// BindToZero binds the transient storage location to a zero value.
```
LuisPH3, line 46:
```suggestion
// BindToZero binds a location in the transient storage a zero value.
```

facuMH (changes requested):
looking good, but I still have a few comments and suggestions

go/ct/st/transient.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2024 Fantom Foundation
```
LuisPH3, line 1: filename transient_storage.go

go/ct/st/transient_test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) 2024 Fantom Foundation
```
LuisPH3, line 1: transient_storage_test.go

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+		// Check different assignments with the same key if they conflict
+		for other, otherNonZero := range t.nonZeroConstraints {
+			if assignment[variable] == assignment[other] && nonZero != otherNonZero {
+				return nil, fmt.Errorf("%w, conflicting constraints %v and %v on the same location", ErrUnsatisfiable, nonZero, otherNonZero)
+			}
+		}
+
+		var key common.U256
+		if pos, isBound := assignment[variable]; isBound {
+			key = pos
```
LuisPH3, line 94: add pos to keys in use.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+		} else {
+			key = getUnusedKey()
+		}
+
+		if nonZero {
+			transient.Set(key, common.RandU256(rnd))
```
LuisPH3, line 100: bind to zero means that the key will not be in the storage (transient)? 
or does it mean that a 0 value is stored?  if so, shall we write an entry in the storage with 0 content?

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+		for {
+			key := common.RandU256(rnd)
+			if _, isPresent := keysInUse[key]; !isPresent {
+				keysInUse[key] = true
+				return key
+			}
+		}
+	}
+
+	transient := &st.TransientStorage{}
```
LuisPH3, line 80: nit: transientStorage, or just storage. Transient is an adjective, you need a noun to name a thing.

LuisPH3 (changes requested):
please check if we need to account for assigned keys into the used keys set. and check if we need to 0 initialize bound to zero locations.

go/ct/gen/state_test.go
```diff
@@ -358,6 +360,7 @@ func TestStateGenerator_ClonesAreIndependent(t *testing.T) {
 		"stack={3≤size≤300}",
 		"memory={}",
 		"storage={}",
+		"transient={transient[$x]=0}",
```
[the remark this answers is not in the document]
simonlechner: As @HerbertJordan explained in the other comments on the this PR, transient storage is an array and this notation shows an array access.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Process zero/nonZero constraints.
+	for variable, nonZero := range t.nonZeroConstraints {
+
+		// Check different assignments with the same key if they conflict
+		for other, otherNonZero := range t.nonZeroConstraints {
+			if assignment[variable] == assignment[other] && nonZero != otherNonZero {
```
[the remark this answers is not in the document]
simonlechner: Good catch, updated it and added unit tests for these cases.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+	if t.unsatisfiable {
+		return "false"
+	}
+
+	var parts []string
+	keys := maps.Keys(t.nonZeroConstraints)
+	slices.Sort(keys)
+	for _, key := range keys {
+		if t.nonZeroConstraints[key] {
+			parts = append(parts, fmt.Sprintf("transient[%v]≠0", key))
```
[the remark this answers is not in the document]
simonlechner: See comment on state_test.go

go/ct/rlz/conditions.go
```diff
@@ -576,6 +576,83 @@ func (c *storageConfiguration) String() string {
 	return fmt.Sprintf("StorageConfiguration(%v,%v,%v)", c.config, c.key, c.newValue)
 }
 
+////////////////////////////////////////////////////////////
+// Bind Transient Storage to non zero value
+
+type bindTransientStorageToNonZero struct {
```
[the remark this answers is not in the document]
simonlechner: Also storage uses the word bind that's why I adapted it.

go/ct/spc/specification.go
```diff
@@ -717,6 +717,120 @@ func getAllRules() []Rule {
...
+		pushes:    1,
+		parameters: []Parameter{
+			StorageAccessKeyParameter{},
+		},
+		conditions: []Condition{
+			RevisionBounds(R13_Cancun, NewestSupportedRevision),
+			BindTransientStorageToNonZero(Param(0)),
+		},
+		effect: func(s *st.State) {
+			key := s.Stack.Pop()
```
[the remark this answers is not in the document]
simonlechner: I think it is more readable with key variable, you can directly see what the value poped from the stack is.

go/ct/st/state_test.go
```diff
@@ -40,6 +40,7 @@ func getNewFilledState() *State {
 		SetOriginal(NewU256(77), NewU256(4)).
 		SetWarm(NewU256(9), true).
 		Build()
+	s.Transient = &TransientStorage{}
```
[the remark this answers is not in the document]
simonlechner: st/transient_storage does not offer a new method since it is not required.

go/ct/st/transient.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	if t.storage == nil {
+		t.storage = make(map[U256]U256)
+	}
+	t.storage[key] = value
+}
+
+func (t *TransientStorage) Get(key U256) U256 {
+	return t.storage[key]
```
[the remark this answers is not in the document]
simonlechner: All of the listed functions to not need nil checks in go.

go/ct/gen/transient.go
simonlechner: good catch.

go/ct/gen/transient.go
simonlechner: For transient storage it does not make a difference whether 0 is stored or not. See earlier comments on this PR where we decided that it is not required to be saved.

go/ct/gen/transient.go
```diff
@@ -0,0 +1,146 @@
...
+
+// BindToNonZero binds the transient storage location to a non-zero value.
+func (t *TransientStorageGenerator) BindToNonZero(key Variable) {
+	if value, exits := t.nonZeroConstraints[key]; exits {
+		if !value {
+			t.unsatisfiable = true
+			return
+		}
+	} else {
+		t.nonZeroConstraints[key] = true
```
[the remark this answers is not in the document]
simonlechner: This should not be required since the map is not nil when using the NewTransientStorageGenerator method which is the indented use for this generator.

go/ct/st/state_test.go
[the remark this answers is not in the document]
LuisPH3: does this contradicts the comments about not having to check for nil?

go/ct/rlz/conditions.go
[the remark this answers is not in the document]
HerbertJordan: I discussed this before with Simon, the entire terminology related to variables is not well worked out and should be cleaned up as a whole. This is beyond the scope of this PR. For now, better to be consistent with pre-existing code.

go/ct/st/state_test.go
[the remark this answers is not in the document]
simonlechner: No, one is for the st/transient_storage and one for the transient storage generator.

HerbertJordan (commented):
Judging from the list of open comments I assume this is still work in progress.

Please re-request a review when all outstanding changes are complete. Thanks!

go/ct/gen/transient_storage.go
```diff
@@ -0,0 +1,148 @@
...
+
+	transientStorage := &st.TransientStorage{}
+
+	// Process zero/nonZero constraints.
+	for variable, nonZero := range t.nonZeroConstraints {
+
+		// Check different assignments with the same key if they conflict
+		for other, otherNonZero := range t.nonZeroConstraints {
+			_, isBound := assignment[variable]
+			if assignment[variable] == assignment[other] && nonZero != otherNonZero && isBound {
```
HerbertJordan, line 88: nit: the `isBound` check could be done before the loop; if it is not bound, the loop can be skipped.

go/ct/gen/transient_storage.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		if nonZero {
+			transientStorage.Set(key, common.RandU256(rnd))
+		}
+		assignment[variable] = key
+	}
+
+	// Random entries
+	for i := 0; i < rnd.Intn(42); i++ {
```
HerbertJordan, line 108: Up to 42 extra slots filled with data could introduce some overhead for the state enumeration since transient storage is deep-copied. But if this turns out to be an issue, it can be addressed in a follow-up performance PR.

go/ct/gen/transient_storage_test.go
```diff
@@ -0,0 +1,158 @@
...
+	rnd := rand.New(0)
+
+	generator := NewTransientStorageGenerator()
+	generator.BindToNonZero(v1)
+	generator.BindToZero(v2)
+	_, err := generator.Generate(Assignment{}, rnd)
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+}
```
HerbertJordan, line 142: This test should also check:
 - that the variables `v1` and `v2` are assigned to storage addresses in the assignment passed to the generator
 - the produced transient storage has a non-zero value at the address associated to `v1` and a zero value associated to `v2`

I do not see this generator outputs be tested in any other test in this file.

HerbertJordan (changes requested):
I think one test should be extended to check the generated result. Other than that, the PR is good to merge from my point of view.

go/ct/spc/specification_test.go
```diff
@@ -69,13 +69,13 @@ func TestSpecification_SpecificationIsComplete(t *testing.T) {
 }
 
 func TestSpecification_EachRuleProducesAMatchingTestCase(t *testing.T) {
+	rnd := rand.New(0)
```
LuisPH3, line 72: :eyes: follow-up issue? Any thought about why this happens?

LuisPH3 (approved):

go/ct/spc/specification_test.go
HerbertJordan: Is `rnd` actually thread save? The tests are run in parallel.

HerbertJordan (approved):

go/ct/gen/transient_storage.go
```diff
@@ -0,0 +1,149 @@
...
+	nonZeroConstraints map[Variable]bool
+	unsatisfiable      bool
+}
+
+func NewTransientStorageGenerator() *TransientStorageGenerator {
+	return &TransientStorageGenerator{nonZeroConstraints: map[Variable]bool{}}
+}
+
+// BindToNonZero binds a location in transient storage to a non-zero value.
+func (t *TransientStorageGenerator) BindToNonZero(key Variable) {
```
facuMH, line 35: nit: if generator is already unsatisfiable, return?

go/ct/spc/specification_test.go
simonlechner: This will be investigated in #550.

facuMH (approved):

go/ct/gen/transient_storage.go
simonlechner: The performance impact will be neglectable since this function will not be called a big number of times.

Issue #524, opened 2024-06-11
LFVM: Memory expansion overflow checks


simonlechner:
During the changes required for the limiting of the init code size in #498 an issue regarding the memory expansion calculation has been detected. The [toValidMemorySize](https://github.com/Fantom-foundation/Tosca/blob/2715eec8f7fffce39ac7817623fa492071f79900/go/vm/lfvm/memory.go#L31) overflows without any warning or error.

To close this issue it has to be fixed in the LFVM code and investigated why the CT did not produce such an error case.

HerbertJordan:
A similar issue was discovered in the CT specification itself recently, and was addressed by #517.

In particular, see this [comment](https://github.com/Fantom-foundation/Tosca/pull/517#discussion_r1633333850).

LuisPH3:
This issue affects as well   #541, where a unit test supposed to fail was not. 
This ticket has a more interesting take on this issue, why did the CT not find this issue? 
In order to reproduce the issue, revert #541 changes for go/vm/lfvm/memory.go

PR #541, opened 2024-06-14
Fix Bug crashing LFVM when passing large size parameter to LOGn Operation

LuisPH3:
This Bug was detected during early Fuzzing experiments. Tooling is not ready for release, unit test that reproduces bug is provided. 
The Gas computation for the word size cost is not checked for overflows. Memory expansion does, inverting the order prevents the bug from happening without introducing branches.


HerbertJordan (dismissed):

simonlechner (dismissed):

LuisPH3:
Cannot merge, this bug may be bigger than initially thought. 

- [ ] unit tests with max-int should fail, but they don't.  There should be not enough gas
- [ ] related to #524 
- [ ] maybe overflow as well in interpreter.go useGas.
- [ ] Ct does not detect?

facuMH:
as for the `CT does not detect` task, could you add a regression test for this ? that is generate a state that would trigger this bug ?

go/vm/lfvm/interpreter_test.go
```diff
@@ -69,24 +69,39 @@ func getContext(code Code, data []byte, runContext vm.RunContext, stackPtr int,
...
-	}
-	ok = ctx.UseGas(100)
-	if ok {
-		t.Errorf("expected failed useGas function, got ok")
-	}
-	if ctx.gas != 90 {
-		t.Errorf("expected gas in context is 90 also after failing, got %d", ctx.gas)
+
+	tests := []struct {
+		name      string
```
HerbertJordan, line 74: nit: when a case has a name it is custom to use the name as the key of a map; something like this:
```

	tests := map[string]struct {
		gas       vm.Gas
		amount    vm.Gas
		wantedGas vm.Gas
		expected  bool
		status    Status
	}{
		"Zero amount":       {100, 0, 100, true, RUNNING},
		"Sufficient gas":    {100, 10, 90, true, RUNNING},
		"Insufficient gas":  {10, 100, 10, false, OUT_OF_GAS},
		"All gas":           {100, 100, 0, true, RUNNING},
		"Negative amount":   {100, -100, 200, true, RUNNING},
		"Negative gas":      {-100, 100, -100, false, OUT_OF_GAS},
		"Negative Negative": {-100, -100, -100, false, OUT_OF_GAS},
	}

            ...
        }
```

go/vm/lfvm/interpreter_test.go
```diff
@@ -69,24 +69,39 @@ func getContext(code Code, data []byte, runContext vm.RunContext, stackPtr int,
...
+		gas       vm.Gas
+		amount    vm.Gas
+		wantedGas vm.Gas
+		expected  bool
+		status    Status
```
HerbertJordan, line 79: Renaming those fields would help readability:
- `gas` => `initialGas`
- `amount` => `cost` or `delta`
- `wantedGas` => `resultingGas`
- `expected` => not needed at all; can be computed as `initialGas` >= `amount` 
- `status` => not need, can be derived from other parameters; in particular `!expected` => `OUT_OF_GAS`

go/vm/lfvm/interpreter_test.go
```diff
@@ -69,24 +69,39 @@ func getContext(code Code, data []byte, runContext vm.RunContext, stackPtr int,
...
+	}{
+		{"Zero amount", 100, 0, 100, true, RUNNING},
+		{"Sufficient gas", 100, 10, 90, true, RUNNING},
+		{"Insufficient gas", 10, 100, 10, false, OUT_OF_GAS},
+		{"All gas", 100, 100, 0, true, RUNNING},
+		{"Negative amount", 100, -100, 200, true, RUNNING},
```
HerbertJordan, line 85: Not sure negative gas should be supported in a function that is named "UseGas". 

The problem is that such a behavior could potentially lead to non-termination of the EVM, if for some reason a bug produces some negative gas costs. The gas pool would no longer get strictly reduced.

I would suggest that a negative gas value is handled like a infinite gas cost. Wdyt?

go/vm/lfvm/memory.go
```diff
@@ -26,18 +27,29 @@ func NewMemory() *Memory {
...
 }
 
 func (m *Memory) ExpansionCosts(size uint64) vm.Gas {
 	if m.Len() >= size {
 		return 0
 	}
 	size = toValidMemorySize(size)
-	memory_size_word := uint64((size + 31) / 32)
+	memory_size_word := sizeInWords(size)
 	new_costs := vm.Gas((memory_size_word*memory_size_word)/512 + (3 * memory_size_word))
```
HerbertJordan, line 53: The multiplication of `memory_size_word*memory_size_word` can overflow. Can you add a TODO for this as well?

HerbertJordan (dismissed):

go/vm/lfvm/interpreter_test.go
LuisPH3: It is early for me to do these judgements, but I begin to think that negative gas values, when dealing with costs, are symptoms of overflows. 
I see two options, treat it as infinite gas, or introduce overflow checks on every gas computation. 

- The first assumes that a negative gas value is indeed an overflow.
- The latter quite a big change. 

I cannot tell you at this time how severe is this.

go/vm/lfvm/memory.go
facuMH: would it be ok to add that check in this PR  ?

facuMH (approved):
after the changes on the tests commented by herbert, it should be good to merge

go/vm/lfvm/memory.go
LuisPH3: Because this PR is getting too large, we will re-evaluate dynamic costs in #524

go/vm/lfvm/interpreter_test.go
LuisPH3: UseGas  shall no longer accept negative costs, by returning out of gas... unit testing and ct pass. This unblocks fuzzing.

LuisPH3:
The PR added extra changes:
- refactor UseGas as observed by @HerbertJordan 
- reuse SizeInWords for other lfvm ops. this was immediately detected by the fuzzer after the initial modification.

facuMH (approved):

simonlechner (approved):

PR #545, opened 2024-06-19
Prevent gas oveflow in the memory expansion cost computation

LuisPH3:
Use the same method as [Geth](https://github.com/Fantom-foundation/tosca-go-ethereum/blob/4db0f269681aa0137a5510532a8c4f0c423237e7/core/vm/gas_table.go#L39):

- Verify that overflow can still not happen, although LFVM gas is signed. 
- This is related to #524, may need to be reverted to investigate the cause of the CT not detecting this bug.

### DIFF
--- a/go/vm/lfvm/memory.go
+++ b/go/vm/lfvm/memory.go
@@ -44,14 +44,24 @@ func toValidMemorySize(size uint64) uint64 {
 	return fullWordsSize
 }
 
+const (
+	// Maximum memory size allowed
+	// This magic number comes from 'core/vm/gas_table.go' 'memoryGasCost' in geth
+	maxMemoryExpansionSize = 0x1FFFFFFFE0
+)
+
 func (m *Memory) ExpansionCosts(size uint64) vm.Gas {
 	if m.Len() >= size {
 		return 0
 	}
 	size = toValidMemorySize(size)
-	memory_size_word := sizeInWords(size)
-	// TODO: check for overflow, fix in #524
-	new_costs := vm.Gas((memory_size_word*memory_size_word)/512 + (3 * memory_size_word))
+
+	if size > maxMemoryExpansionSize {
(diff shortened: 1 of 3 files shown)


go/vm/lfvm/memory.go
```diff
@@ -44,14 +44,24 @@ func toValidMemorySize(size uint64) uint64 {
 	return fullWordsSize
 }
 
+const (
+	// Maximum memory size allowed
+	// This magic number comes from 'core/vm/gas_table.go' 'memoryGasCost' in geth
```
facuMH, line 49: thank you very much for this comment.

go/vm/lfvm/memory_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"math"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/vm"
+)
+
+func TestMagicNumber(t *testing.T) {
+
+	// Call me paranoid, but I am repeating the test for the magic number:
+	// The difference between geth and lfvm is that they use uint64 as gas
```
facuMH, line 23:
```suggestion
	// The difference between geth and lfvm is that geth uses uint64 as gas
```

facuMH (approved):
LGTM

facuMH (approved):

go/vm/lfvm/memory.go
```diff
@@ -44,14 +44,33 @@ func toValidMemorySize(size uint64) uint64 {
...
```
HerbertJordan, line 49: You could de-magicfy it a bit by stating that this is the largest amount of memory that can be allocated such that the expansion costs do not exceed `MaxInt64`.

HerbertJordan (approved):
Thanks for fixing this!

simonlechner (approved):

PR #546, opened 2024-06-19
CT driver, print progress before exit.

LuisPH3:
This commit prints enumeration status one last time before exiting. This is useful when running small filtered ct executions, to read the number of executed states.

A low-hanging fruit that improves tool usage:
- Filters resulting in an empty set will not warn about it but only report success, now the count of states can be seen.
- It is nice to see the final state count instead of the number 5 seconds before, and the completion time.

### DIFF
--- a/go/ct/spc/enumeration.go
+++ b/go/ct/spc/enumeration.go
@@ -57,34 +57,28 @@ func ForEachState(
 		startTime := time.Now()
 		lastTime := startTime
 		lastTestCounter := int64(0)
+
+		checkTimingAndPrint := func(now time.Time) {
+			cur := testCounter.Load()
+
+			diffCounter := cur - lastTestCounter
+			diffTime := now.Sub(lastTime)
+
+			lastTime = now
+			lastTestCounter = cur
+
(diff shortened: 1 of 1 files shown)


go/ct/spc/enumeration.go
```diff
@@ -60,6 +60,18 @@ func ForEachState(
...
+
+				diffCounter := cur - lastTestCounter
+				diffTime := curTime.Sub(lastTime)
+
+				lastTime = curTime
+				lastTestCounter = cur
+
+				relativeTime := curTime.Sub(startTime)
+				rate := float64(diffCounter) / diffTime.Seconds()
+				printIssueCounts(relativeTime, rate, cur)
```
HerbertJordan, line 74: This code is (almost) identical to the case below. It would be nice to factor this out into a lambda before the for loop.

HerbertJordan (dismissed):
Nice addition. The lack of it annoyed me more than once in the past :+1:

simonlechner (dismissed):
Nice addition!

simonlechner (approved):

HerbertJordan (approved):

PR #547, opened 2024-06-20
Add Fuzzer tests for Evm.StepN

LuisPH3:
This PR adds 3 fuzzer functions for the ct Evm.StepN interface, mostly targeting the LFVM implementation.

- FuzzLfvm: runs one instruction at the time and expects no panic.
- FuzzGeth: equivalent to FuzzLfvm, but intended to be used when errors are found as a reference. 
- FuzzDifferentialLfvmVsGeth: Executes both lfvm and geth with the same input state and compares resulting states, in the same fashion as the ct does, but driven by the mutation+coverage algorithm if the fuzzing engine. 

All fuzzing tests are automatically run as unit-tests with the initial test data + any test data files found in `go/ct/testdata/fuzz/NameOfTheTest/*`. Such files are automatically generated but the fuzzing engine when a failure or panic is found.

This PR does not:
- Script the Fuzzing as a CI step, this task requires figuring out a method to download the mentioned testdata files. 
- Decide if we commit directly offending testdata files or if they shall be manually converted into a unit-test
such tasks are still up for discussion.


BUILD.md
```diff
@@ -249,4 +249,18 @@ Overall coverage rate:
...
 The report will be generated in HTML form to alow visualization of each line of code coverage for each C++ file in the project. The entry HTML page to the report is located at the C++ build folder: `cpp/build/coverage/index.html`
 
-In VSCode, line by line coverage can be visualized using the extension [gcov-viewer](https://marketplace.visualstudio.com/items?itemName=JacquesLucke.gcov-viewer)
\ No newline at end of file
+In VSCode, line by line coverage can be visualized using the extension [gcov-viewer](https://marketplace.visualstudio.com/items?itemName=JacquesLucke.gcov-viewer)
+
+
+## Fuzzing
+
+Go code can be easily fuzzed: fuzzing attempts to find bugs by mutating a test input data set and measuring test coverage. This complementary technique aims stability issues.
```
facuMH, line 257:
```suggestion
Go code can be easily fuzzed: fuzzing attempts to find bugs by mutating a test input data set and measuring test coverage. This complementary technique targets stability issues.
```

BUILD.md
```diff
@@ -249,4 +249,18 @@ Overall coverage rate:
...
+
+### Fuzzing Evm.StepN interface
+
+The ct Evm.StepN interface is used to evaluate one N instructions in different EVM implementations. There are 2 Fuzzer tests against this interface:
```
facuMH, line 261:
```suggestion
The ct Evm.StepN interface is used to evaluate N instructions in different EVM implementations. There are 2 Fuzzer tests against this interface:
```

BUILD.md
```diff
@@ -249,4 +249,18 @@ Overall coverage rate:
...
+
+- Crash test: FuzzLfvm will execute instructions one at the time looking for a panic. ```make fuzz-lfvm```
+- Differential test: FuzzDifferentialLfvmVsGeth ```make fuzz-lfvm-diff```
```
facuMH, line 264: maybe one phrase about this command as well ?

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+
+const (
+	fuzzIdealStackSize     int = 7  // < max pops in a single instruction
+	fuzzMaximumCodeSegment int = 33 // < 1 instruction with 32 data bytes
+)
+
+// prepareInitialTestData is a helper function to be used by similar fuzzing tests
+// the arguments passed to the f.Add function needs to match the arguments
+// passed to the f.Fuzz function in type, position and number
+func prepareInitialTestData(f *testing.F, rnd *rand.Rand) {
```
HerbertJordan, line 117: rename to `prepareFuzzingSeeds`?

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+
+	for revision := R07_Istanbul; revision <= R13_Cancun; revision++ {
```
facuMH, line 119:
```suggestion
	for revision := R07_Istanbul; revision <= NewestSupportedRevision; revision++ {
```

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+		for op := 0x00; op < 0xFF; op++ {
+			for _, gas := range []int64{0, 1, 6, 10, 1000, math.MaxInt64} {
```
facuMH, line 121: magic numbers ? maybe call `gasDomain.SamplesForAll({})`

go/ct/evm_test.go
HerbertJordan, line 1: Initially I thought the fuzz tests should be placed in its own directory, something like `go/fuzz` next to `go/ct`. But while reading the code, I realized that the fuzzer makes heavy use of CT infrastructure, so I guess it is fine to be in this package.

Just, place the fuzzing code in its own file, something like `go/ct/fuzz_test.go`. Wdyt?

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+		original.Revision,
+		original.Pc, original.Code.HumanReadableString(0, 7),
+		resultState.Status, referenceState.Status,
+		original.Gas, resultState.Gas, referenceState.Gas,
+		original.Stack.Size(), resultState.Stack.Size(), referenceState.Stack.Size(),
+		original.Memory.Size(), resultState.Memory.Size(), referenceState.Memory.Size(),
+	)
+}
+
+func FuzzDifferentialLfvmVsGeth(f *testing.F) {
```
HerbertJordan, line 194: nit: I would favor having the "important" stuff, in this case the test targets, at the top of the file, and utility functions and implementation details like "prepareInitialTestData" at the bottom. This way, when opening the files the relevant part comes first.

I would assume that the habit of placing stuff you use before its use comes from years of C/C++ experience, but we can exploit that the order of declaration in Go is not relevant. Wdyt?

go/ct/evm_test.go
HerbertJordan: nit: use Min/Max revision constants

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+	fuzzMaximumCodeSegment int = 33 // < 1 instruction with 32 data bytes
+)
+
+// prepareInitialTestData is a helper function to be used by similar fuzzing tests
+// the arguments passed to the f.Add function needs to match the arguments
+// passed to the f.Fuzz function in type, position and number
+func prepareInitialTestData(f *testing.F, rnd *rand.Rand) {
+
+	for revision := R07_Istanbul; revision <= R13_Cancun; revision++ {
+		for op := 0x00; op < 0xFF; op++ {
```
HerbertJordan, line 120:
```suggestion
		for op := 0x00; op <= 0xFF; op++ {
```

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+
+		stack := st.NewStack()
+
+		for i := 0; i < len(stackBytes); i += 32 {
+			var stackValue [32]byte
+			if i+32 <= len(stackBytes) {
+				copy(stackValue[:], stackBytes[i:i+32])
+			} else {
+				copy(stackValue[:], stackBytes[i:])
+			}
```
facuMH, line 234:
```suggestion
			end := len(stackBytes)
			if i+32 <= len(stackBytes) {
				end = i+32
			} 
			copy(stackValue[:], stackBytes[i:end])
```

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+				)
+			}
+		}
+	}
+
+	// add one more with a full stack
+	ops := [fuzzMaximumCodeSegment]byte{}
+	rnd.Read(ops[:])
+	ops[0] = byte(0x00)
+	fullStack := [1024]byte{}
```
HerbertJordan, line 162: A full stack should be 1024*32 bytes, right?

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+		// Do not ignore stack sizes close to the overflow, as they are interesting
+		if len(stackBytes) > fuzzMaximumCodeSegment*32 && len(stackBytes) < (1024-fuzzMaximumCodeSegment) {
+			t.Skip("Uninteresting stack size", len(stackBytes))
+		}
+
+		stack := st.NewStack()
+
+		for i := 0; i < len(stackBytes); i += 32 {
+			var stackValue [32]byte
+			if i+32 <= len(stackBytes) {
```
facuMH, line 318: same as above

go/ct/st/code.go
```diff
@@ -142,3 +142,19 @@ func (c *Code) Copy() []byte {
 func (c *Code) String() string {
 	return fmt.Sprintf("%x", c.code)
 }
+
+// HumanReadableString returns a string with the length of the code and the
+// Human readable form for the opcodes in range [start, start+length).
```
facuMH, line 147:
```suggestion
// human readable form for the opcodes in range [start, start+length).
```

go/ct/evm_test.go
HerbertJordan, line 162: For those larger structures, using slices would be more adequate. It also saves you from converting the array into slices using the `[:]` operator.
```suggestion
	fullStack := make([]byte, 1024*32)
```

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+// the arguments passed to the f.Add function needs to match the arguments
+// passed to the f.Fuzz function in type, position and number
+func prepareInitialTestData(f *testing.F, rnd *rand.Rand) {
+
+	for revision := R07_Istanbul; revision <= R13_Cancun; revision++ {
+		for op := 0x00; op < 0xFF; op++ {
+			for _, gas := range []int64{0, 1, 6, 10, 1000, math.MaxInt64} {
+
+				// generate a code segment with the operation followed by 6 random values
+				ops := [fuzzMaximumCodeSegment]byte{}
```
HerbertJordan, line 124: You should probably prefer slices over arrays here and for the `stack` variable below.

go/ct/st/code_test.go
```diff
@@ -140,6 +140,57 @@ func TestCode_Printer(t *testing.T) {
...
+	}{
+
+		"empty": {
+			code:   NewCode([]byte{}),
+			start:  0,
+			length: 4,
+			want:   "len(0)",
+		},
+		"complete": {
+			code:   NewCode([]byte{byte(ADD), byte(PUSH1), 0x0C, byte(PUSH2)}),
```
facuMH, line 158: nit: since `[]byte{byte(ADD), byte(PUSH1), 0x0C, byte(PUSH2)}` is repeated for almost all cases, consider having a `baseCode` variable.

facuMH (changes requested):
Looks good, looking forward to the presentation !! that said, I still have a couple of comments

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+	)
+}
+
+// errorReportString is a helper function to print a summary of the test states.
+// Fuzzing generates a single error at the time, but when tunning the initial test
+// data, such states are evaluated in parallel and errors are produced in large numbers.
+// This function aims to present a compact overview of the most significant differences to
+// keep process output small and readable.
+func errorReportString(original *st.State, resultState *st.State, referenceState *st.State) string {
+	return fmt.Sprintf(`
```
HerbertJordan, line 178: `st.State` has a [`Diff`](https://github.com/Fantom-foundation/Tosca/blob/c90d7bfdd232c87e22a1238ead1f9a46d3f2c6f4/go/ct/st/state.go#L363) function that would produce a list of differences between two states. Would this be helpful for this function?

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+	rnd := rand.New(0)
+
+	prepareInitialTestData(f, rnd)
+
+	f.Fuzz(func(t *testing.T, opCodes []byte, gas int64, revision byte, stackBytes []byte) {
+		if gas < 0 {
+			t.Skip("negative gas", gas)
+		}
+
+		if Revision(revision) < R07_Istanbul || Revision(revision) > NewestSupportedRevision {
```
HerbertJordan, line 208: nit: `MinRevision` instead of `R07_Istanbul`

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+		}
+
+		if len(opCodes) > fuzzMaximumCodeSegment {
+			t.Skip("too many opCodes,  not interesting")
+		}
+
+		// Ignore stack sizes larger than 7 words, as they are not interesting
+		// Do not ignore stack sizes close to the overflow, as they are interesting
+		if len(stackBytes) > fuzzIdealStackSize*32 && len(stackBytes) < (1024-fuzzIdealStackSize) {
+			t.Skip("Uninteresting stack size", len(stackBytes))
```
HerbertJordan, line 223: nice :+1:

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+			}
+			stack.Push(NewU256FromBytes(stackValue[:]...))
+		}
+
+		code := st.NewCode(opCodes)
+		state := st.NewState(code)
+		state.Gas = vm.Gas(gas)
+		state.Revision = Revision(revision)
+		state.Stack = stack
+		state.BlockContext.TimeStamp = GetForkTime(state.Revision)
```
HerbertJordan, line 331: It looks like this conversion of fuzzing arguments to a state is very similar in the `fuzzVm` and `FuzzDifferentialLfvmVsGeth` case. This should be factored out into a utility function.

go/ct/evm_test.go
```diff
@@ -102,3 +105,246 @@ func TestCt_ExplicitCases(t *testing.T) {
...
+
+		result, _ := testee.StepN(state.Clone(), 1)
```
HerbertJordan, line 333: I guess cloning the state here is not really necessary, right?

BUILD.md
HerbertJordan:
```suggestion
Go provides fuzzing support through its standard library. Fuzzing attempts to find bugs by mutating a test input data set and measuring test coverage. This complementary technique aims to identify stability issues.
```

BUILD.md
```diff
@@ -249,4 +249,18 @@ Overall coverage rate:
...
+
+Notice lack of evmzero target: go utilizes native code coverage to inform the algorithm, which would use this information to drive the mutation process. There is currently no method to retrieve this information from the C++ library, this renders the C++/go fuzzer as a brute-force algorithm.
```
HerbertJordan, line 266: Running it in `brute-force` mode might be better than nothing, right?

go/ct/evm_test.go
HerbertJordan: `st.State` also has a [String](https://github.com/Fantom-foundation/Tosca/blob/c90d7bfdd232c87e22a1238ead1f9a46d3f2c6f4/go/ct/st/state.go#L243) function that (should) produces human readable output.

go/ct/st/code.go
```diff
@@ -142,3 +142,19 @@ func (c *Code) Copy() []byte {
...
+// - If the slice to be printed overflows the existing code, the overlapping code is printed.
+// - If start exceeds the code length, the length of the code is printed.
+func (c *Code) HumanReadableString(start int, length int) string {
+	result := fmt.Sprintf("len(%d)", len(c.code))
+	if start >= len(c.code) {
+		return result
+	}
+	end := min(length, len(c.code)-start) + start
+	for _, op := range c.code[start:end] {
+		result = fmt.Sprintf("%s; %s", result, OpCode(op).String())
```
HerbertJordan, line 157: This code would print data sections like operations. So a "PUSH2" followed by 2 data bytes would interpret the two bytes as operations.

go/ct/st/code_test.go
```diff
@@ -140,6 +140,57 @@ func TestCode_Printer(t *testing.T) {
...
+			start:  0,
+			length: 4,
+			want:   "len(4); ADD; PUSH1; op(0C); PUSH2",
```
HerbertJordan, line 161: nit: if you replace the `0x0C` with e.g. `0x01` the output would read `ADD, PUSH1, ADD, PUSH2` ...

go/ct/st/code.go
```diff
@@ -142,3 +142,19 @@ func (c *Code) Copy() []byte {
 func (c *Code) String() string {
 	return fmt.Sprintf("%x", c.code)
 }
+
+// HumanReadableString returns a string with the length of the code and the
+// Human readable form for the opcodes in range [start, start+length).
+// - If the slice to be printed overflows the existing code, the overlapping code is printed.
+// - If start exceeds the code length, the length of the code is printed.
+func (c *Code) HumanReadableString(start int, length int) string {
```
HerbertJordan, line 150: nit: This function name misses a verb ...

HerbertJordan (changes requested):
Looks good, just a few minor requests for improvement.

BUILD.md
LuisPH3: Actually writing this sentence got me thinking: Running the diff between geth and evmzero, will retrieve coverage reports from geth, and blindly stress similar regions in evmzero. I will make a couple of runs today and add a target for this purpose.

go/ct/st/code.go
LuisPH3: I wanted to call it To.... but since the to_string function is called just string in go, I tried to stay idiomatic.

go/ct/st/code_test.go
LuisPH3: That was my first idea: it broke other tests for the opcode. String function. I did not want to risk breaking something else by changing it.

BUILD.md
LuisPH3: evmzero differential fuzzer yielded results, just with the seed corpus. Those need to be fixed before we can even start the fuzzing: 
[#549](https://github.com/Fantom-foundation/Tosca/issues/549)

go/ct/st/code_test.go
LuisPH3: done

go/ct/st/code.go
HerbertJordan: I think the "String" function is called "String" because it produces a value of type "string".

Not sure about the idiomatic way in Go in this regard. I guess we are free to chose how we like it. Up to you.

go/ct/evm_test.go
LuisPH3: the goal here is:

- help the fuzzer to early discard mutations that do not increase coverage. 
- avoid cryptic encoding of states, as the future bugs will be stored, and they need to be potentially migrated or converted into a unit test.

go/ct/evm_test.go
LuisPH3: yes.

go/ct/evm_test.go
LuisPH3: Second time is not necessary as long as the VM does not modify it (to keep the input for the report). 
is this the case?

go/ct/evm_test.go
LuisPH3: - State.String is not used to keep output small, when tuning a new fuzzer with an existing corpus errors are produced in large numbers, I find is useful to see the whole of them instead of the details. Which is something you can always get with a debug. Example: #549 
- State.Diff is used in the last comparison printout.

HerbertJordan (approved):
Looks good, thanks!

go/ct/st/code.go
LuisPH3: Renamed to ToXXXXXX

- c++ calls these functions to_string
- rust uses "to" prefix when a new instance is created, and "as" when a new type is returned but no new memory is allocated.

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			t.Fatalf("failed to run test case in reference VM: %v", err)
+		}
+
+		if testeeResultState.Status != referenceResultState.Status {
+			t.Fatal("invalid result, status does not match reference status:", errorReportString(state, testeeResultState, referenceResultState))
+		}
+
+		// if result is other than running, further checks may be misleading
+		if testeeResultState.Status != st.Running {
```
facuMH, line 98: I would not consider `Failure` same as `Returned` or `Revert`. Perhaps this check can be a little more specialized

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,274 @@
...
+			// list just sketches a region of interest around which the fuzzer will generate
+			// more values. I found no measurable difference about being more accurate.
+			for _, gas := range []int64{0, 1, 6, 10, 1000, math.MaxInt64} {
+
+				// generate a code segment with the operation followed by 6 random values
+				ops := make([]byte, fuzzMaximumCodeSegment)
+				rand.Read(ops[:])
+				ops[0] = byte(op)
+
+				// generate a stack: the stack contains a mixture of values
```
facuMH, line 162: would it help to make use of `gen.StackGenerator` ?

go/ct/st/code.go
```diff
@@ -142,3 +143,29 @@ func (c *Code) Copy() []byte {
 func (c *Code) String() string {
 	return fmt.Sprintf("%x", c.code)
 }
+
+// ToHumanReadableString returns a string with the length of the code and the
+// human readable form for the opcodes in range [start, start+length).
+// - If the slice to be printed overflows the existing code, the overlapping code is printed.
+// - If start exceeds the code length, the length of the code is printed.
+// - Data found in the code is printed as decimal numbers (to differentiate from unused opcodes).
+func (c *Code) ToHumanReadableString(start int, length int) string {
```
facuMH, line 152: nice addition !!! probably should have been a separate PR.

facuMH (dismissed):
looks good !! just a couple of sugestions/comments

go/ct/evm_test.go
LuisPH3: After some experiments, it seems that if I try to release the original state, there is a double release situation.
This shows me that the original state is reused during execution and for better error reporting I will keep 3 instances of the state: original state, reference result, and testee result.
This can be optimized later on if neeeded.

go/ct/evm_fuzz_test.go
LuisPH3: No, This fuzzing tests intentionally leave the initial seed non-exhaustive. The interesting cases are in these tests found through the fuzzing itself and not because of the initial seed.

go/ct/st/code.go
LuisPH3: yes.

LuisPH3:
After review changes, differential test found a new error in the lfvm: #547

go/ct/evm_fuzz_test.go
LuisPH3: Good idea, unfortunately the Fuzzer fails at the moment, and I can not do this change and test it. I comment on it, and we can expand the fuzzer later on.

BUILD.md
LuisPH3: It should be fixed now. Thanks for checking.

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+	fuzzVm(geth.NewConformanceTestingTarget(), f)
+}
+
+// FuzzLfvm is a fuzzing test for lfvm
+func FuzzLfvm(f *testing.F) {
+	fuzzVm(lfvm.NewConformanceTestingTarget(), f)
+}
+
+// FuzzLfvm is a fuzzing test for evmzero
+// func FuzzEvmzero(f *testing.F) {
```
facuMH, line 42: Add a comment as for why is this function commented off.

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+// 	differentialFuzz(f,
+// 		evmzero.NewConformanceTestingTarget(),
+// 		geth.NewConformanceTestingTarget(),
+// 	)
+// }
+
+//////////////////////////////////////////////////////////////////////////////
+// Fuzzing helpers
+
+func differentialFuzz(f *testing.F, testeeVm ct.Evm, referenceVm ct.Evm) {
```
facuMH, line 68:
```suggestion
func differentialFuzz(f *testing.F, testeeVm, referenceVm ct.Evm) {
```

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+	rnd := rand.New(0)
+
+	prepareFuzzingSeeds(f, rnd)
+
+	f.Fuzz(func(t *testing.T, opCodes []byte, gas int64, revision byte, stackBytes []byte) {
```
facuMH, line 74: why not use corresponding `vm`  package types for opcodes, gas, revision ?

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+		// Hack: lfvm does a pc transformation, but for code smaller than the required jump the pc will point to a different location
+		// - Geth will point to pc+offset, whenever it is an overflow
+		// - Lfvm will point min(pc+offset, len(code))
+		if testeeResultState.Pc == uint16(len(opCodes)) &&
+			testeeResultState.Pc != referenceResultState.Pc {
+			testeeResultState.Pc = referenceResultState.Pc
+		}
+
+		if !testeeResultState.Eq(referenceResultState) {
+			t.Fatal("invalid result, result state does not match reference state:", testeeResultState.Diff(referenceResultState), errorReportString(state, testeeResultState, referenceResultState))
```
facuMH, line 126: maybe it's a nit thing, but I don't like `invalid result, result state`, maybe `resulting state` or `test result state`. wdyt ?

facuMH (dismissed):

go/ct/evm_fuzz_test.go
LuisPH3: this is a requirement form the fuzzer, there is only a list of allowed types: 
-     string, []byte
-     int, int8, int16, int32/rune, int64
-     uint, uint8/byte, uint16, uint32, uint64
-     float32, float64
-     bool

go/ct/evm_fuzz_test.go
LuisPH3:
```suggestion
			t.Fatal("invalid result, resulting state does not match reference state:", testeeResultState.Diff(referenceResultState), errorReportString(state, testeeResultState, referenceResultState))
```

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+func FuzzGeth(f *testing.F) {
+	fuzzVm(geth.NewConformanceTestingTarget(), f)
+}
+
+// FuzzLfvm is a fuzzing test for lfvm
+func FuzzLfvm(f *testing.F) {
+	fuzzVm(lfvm.NewConformanceTestingTarget(), f)
+}
+
+// FuzzLfvm is a fuzzing test for evmzero
```
LuisPH3, line 41:
```suggestion
// FuzzLfvm is a fuzzing test for evmzero, (issue #549 )
```

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/Fantom-foundation/Tosca/go/ct"
+	. "github.com/Fantom-foundation/Tosca/go/ct/common"
+	"github.com/Fantom-foundation/Tosca/go/ct/st"
+	"github.com/Fantom-foundation/Tosca/go/vm"
+	"github.com/Fantom-foundation/Tosca/go/vm/geth"
+	"github.com/Fantom-foundation/Tosca/go/vm/lfvm"
+)
+
+// FuzzGeth is a fuzzing test for the geth EVM implementation
+// TODO: it would be interesting to have a method to extend the corpus of the fuzzer
```
HerbertJordan, line 29: An idea for this (both for follow-up work, not this PR): run the fuzzer on a list of EVMs instead of a single one; this way, you only have one entry point and the corpus is shared

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+// 		geth.NewConformanceTestingTarget(),
+// 	)
+// }
+
+//////////////////////////////////////////////////////////////////////////////
+// Fuzzing helpers
+
+func differentialFuzz(f *testing.F, testeeVm, referenceVm ct.Evm) {
+
+	rnd := rand.New(0)
```
HerbertJordan, line 70: nit: Since this random number generator is only used by the seed generation, I would move its creation inside the `prepareFuzzingSeeds` function.

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+		testeeResultState, err := testeeVm.StepN(state.Clone(), 1)
+		defer func() {
+			if testeeResultState != nil {
+				testeeResultState.Release()
+			}
+		}()
+		if err != nil {
+			t.Fatalf("failed to run test case: %v", err)
+		}
```
HerbertJordan, line 90: Would this be the same?
```suggestion
		if err != nil {
			t.Fatalf("failed to run test case: %v", err)
		}
		defer testeeResultState.Release()
```

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+	prepareFuzzingSeeds(f, rnd)
+
+	f.Fuzz(func(t *testing.T, opCodes []byte, gas int64, revision byte, stackBytes []byte) {
+		state, err := corpusInputDataToCtState(opCodes, gas, revision, stackBytes)
+		if err != nil {
+			t.Skip(err)
+		}
+
+		result, _ := testee.StepN(state, 1)
```
HerbertJordan, line 143: The error returned by `StepN` here should be checked. I guess any error is to be considered a failure.

go/ct/evm_fuzz_test.go
```diff
@@ -0,0 +1,285 @@
...
+	rnd.Read(fullStack[:])
+	f.Add(
+		ops,                // opCodes
+		int64(0),           // gas
+		byte(R07_Istanbul), // revision
+		fullStack,          // stack
+	)
+}
+
+func corpusInputDataToCtState(opCodes []byte, gas int64, revision byte, stackBytes []byte) (*st.State, error) {
```
HerbertJordan, line 218: This logic of this function might be complex enough to justify a unit test to make sure it is working as intended.
The test could be added in this file.

HerbertJordan (approved):

go/ct/evm_fuzz_test.go
LuisPH3: Yes, I thought about it. I would have to run some experiment to see how does this scale, and it is nice to use when a bug is found.
Another competing idea is to link or copy found errors corpus from `/testdata/FuzzingTest1/datahash` to testdata/fuzzingTest2/datahash`, since they follow the same layout they should be easily exchangeable between any crash test, and diff tests. Such files are generated when an error is found, I guess they should be committed, I have a couple of unsolved ones which I will commit once evmzero bugs are fixed.

go/ct/evm_fuzz_test.go
LuisPH3: This assumes that we cannot return both error and value !=  nil.
So far it is the pattern everywhere, I guess it is a convention. 
I'll write it like everywhere else.

go/ct/evm_fuzz_test.go
LuisPH3: yes, this routine was written before the diffing case, and I would not filter for invalid inputs. The assumption here is that the StepN routine would be robust against invalid inputs and not panic. 
To make this compatible with the diffing test, the input is now filtered for sane data.
I'll add a check.

facuMH (dismissed):

facuMH (dismissed):

HerbertJordan (approved):

simonlechner (approved):

facuMH (approved):

HerbertJordan (approved):

Issue #420, opened 2024-04-13
evmc: introduce generic void context parameter


HerbertJordan:
Tosca depends on a modified version of the EVMC package mainly to facilitate CT integration. These modifications are quite extensive and are unlikely to ever by upstreamed to the main EVMC repository.

To minimise the needed modification, to keep maintenance work limited, the number of changes should be reduced. This could be facilitated by establishing a generic context-passing mechanism through a void-pointer in the package. This way, special applications like the step-wise execution of code could be channeled through a minimal modification with a chance of being up-streamed.

This idea should be explored and tested for our CT use case. If it works, the EVMC modifications should be updated accordingly.

simonlechner:
Migrated to [#45](https://github.com/0xsoniclabs/tosca/issues/45).

PR #548, opened 2024-06-20
CT: update evmc  for blobhash support

facuMH:
This PR adds `EvmcBlobHashes []evmc.Hash` to `HostContext`  in `evmc_interpreter`.

These slice is filled in `evmc_steppable_interpreter` 
And is returned in `GetTxContext()` instead of the one in `params`

This is an ugly fix needed because:
- when passing memory from Go to C, if it's a slice, you need to pin the memory, to insure it will note moved or deleted by Go's garbage collector.
- the problem is that (because they are different types), when calling `evmc_interpter.go:GetTxContext`, in there, it needs to translate from `vm.Hash` in `vm.Params` to `evmc.hash` defined type by `evmc`
- but this means that we cannot `unpin` the `evmc.hash` slice's memory (because it is returned to the C side)
- so we need to pin the `evmcBlobHashes` before calling `C.step_n_wrapper`,
This way we can have this extra slice in `HostContext` defined in evmc_interpreter.

This is a first approach, open to suggestions on how to do this in a better way. 

Q: is it event possible to reinterpret cast the same memory without a copy ?

NOTE: unit tests will test until patch has been merged in evmc (https://github.com/Fantom-foundation/evmc/pull/1)

### DIFF
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -986,7 +986,7 @@ struct Impl<OpCode::BLOBHASH> {
 
   static OpResult Run(uint256_t* top, Context& ctx) noexcept {
     const auto tx_context = ctx.host->get_tx_context();
-    if (static_cast<size_t>(top[0]) < tx_context.blob_hashes_count) {
+    if (top[0] < tx_context.blob_hashes_count) {
       top[0] = ToUint256(tx_context.blob_hashes[static_cast<size_t>(top[0])]);
     } else {
       top[0] = 0;
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -986,8 +986,8 @@ struct Impl<OpCode::BLOBHASH> {
 
   static OpResult Run(uint256_t* top, Context& ctx) noexcept {
     const auto tx_context = ctx.host->get_tx_context();
-    if (top[0] < tx_context.blob_hashes_count) {
-      top[0] = ToUint256(tx_context.blob_hashes[static_cast<int64_t>(top[0])]);
+    if (static_cast<size_t>(top[0]) < tx_context.blob_hashes_count) {
+      top[0] = ToUint256(tx_context.blob_hashes[static_cast<size_t>(top[0])]);
     } else {
       top[0] = 0;
     }
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -987,7 +987,7 @@ struct Impl<OpCode::BLOBHASH> {
   static OpResult Run(uint256_t* top, Context& ctx) noexcept {
     const auto tx_context = ctx.host->get_tx_context();
     if (top[0] < tx_context.blob_hashes_count) {
-      top[0] = ToUint256(tx_context.blob_hashes[static_cast<size_t>(top[0])]);
+      top[0] = ToUint256(tx_context.blob_hashes[static_cast<int64_t>(top[0])]);
     } else {
       top[0] = 0;
     }
--- a/go/vm/evmc/evmc_interpreter.go
+++ b/go/vm/evmc/evmc_interpreter.go
@@ -155,8 +155,9 @@ func toEvmcRevision(revision vm.Revision) (evmc.Revision, error) {
 // context and chain state information external to the the interpreter.
 // It implements the host interface of evmc's Go bindings.
 type hostContext struct {
-	params  vm.Parameters
-	context vm.RunContext
+	params         vm.Parameters
+	context        vm.RunContext
+	EvmcBlobHashes []evmc.Hash
 }
 
 func (ctx *hostContext) AccountExists(addr evmc.Address) bool {
@@ -224,15 +225,17 @@ func (ctx *hostContext) Selfdestruct(addr evmc.Address, beneficiary evmc.Address
 func (ctx *hostContext) GetTxContext() evmc.TxContext {
 	params := ctx.params
 	return evmc.TxContext{
-		GasPrice:   evmc.Hash(params.GasPrice),
-		Origin:     evmc.Address(params.Origin),
-		Coinbase:   evmc.Address(params.Coinbase),
-		Number:     params.BlockNumber,
-		Timestamp:  params.Timestamp,
-		GasLimit:   int64(params.GasLimit),
-		PrevRandao: evmc.Hash(params.PrevRandao),
-		ChainID:    evmc.Hash(params.ChainID),
-		BaseFee:    evmc.Hash(params.BaseFee),
+		GasPrice:    evmc.Hash(params.GasPrice),
+		Origin:      evmc.Address(params.Origin),
+		Coinbase:    evmc.Address(params.Coinbase),
+		Number:      params.BlockNumber,
+		Timestamp:   params.Timestamp,
+		GasLimit:    int64(params.GasLimit),
+		PrevRandao:  evmc.Hash(params.PrevRandao),
+		ChainID:     evmc.Hash(params.ChainID),
+		BaseFee:     evmc.Hash(params.BaseFee),
+		BlobBaseFee: evmc.Hash(params.BlobBaseFee),
+		BlobHashes:  ctx.EvmcBlobHashes,
 	}
 }
 
--- a/go/vm/evmc/evmc_interpreter.go
+++ b/go/vm/evmc/evmc_interpreter.go
@@ -157,7 +157,7 @@ func toEvmcRevision(revision vm.Revision) (evmc.Revision, error) {
 type hostContext struct {
 	params         vm.Parameters
 	context        vm.RunContext
-	EvmcBlobHashes []evmc.Hash
+	evmcBlobHashes []evmc.Hash
 }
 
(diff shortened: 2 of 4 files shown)


cpp/vm/evmzero/interpreter.cc
```diff
@@ -986,8 +986,8 @@ struct Impl<OpCode::BLOBHASH> {
 
   static OpResult Run(uint256_t* top, Context& ctx) noexcept {
     const auto tx_context = ctx.host->get_tx_context();
-    if (top[0] < tx_context.blob_hashes_count) {
-      top[0] = ToUint256(tx_context.blob_hashes[static_cast<int64_t>(top[0])]);
+    if (static_cast<size_t>(top[0]) < tx_context.blob_hashes_count) {
```
HerbertJordan, line 989: I think this change is eliminating an overflow protection.

In the old version, the `top[0]` value was a `uint256_t` while the `tx_context.blob_hash_count` is a `size_t`. In the implementation of the `operator<` of the `uint256_t` the `size_t` value is extended to a 256-bit value and then compared.

Now, in the new version, the `uint256_t` value is converted into a `size_t`, ignoring overflows, and the compared to the number of hashes. I guess this is not the intention, is it?

go/vm/evmc/evmc_interpreter.go
```diff
@@ -155,8 +155,9 @@ func toEvmcRevision(revision vm.Revision) (evmc.Revision, error) {
 // context and chain state information external to the the interpreter.
 // It implements the host interface of evmc's Go bindings.
 type hostContext struct {
-	params  vm.Parameters
-	context vm.RunContext
+	params         vm.Parameters
+	context        vm.RunContext
+	EvmcBlobHashes []evmc.Hash
```
HerbertJordan, line 160: this field does not need to be public

go/vm/evmc/evmc_interpreter.go
```diff
@@ -224,15 +225,17 @@ func (ctx *hostContext) Selfdestruct(addr evmc.Address, beneficiary evmc.Address
...
+		Origin:      evmc.Address(params.Origin),
+		Coinbase:    evmc.Address(params.Coinbase),
+		Number:      params.BlockNumber,
+		Timestamp:   params.Timestamp,
+		GasLimit:    int64(params.GasLimit),
+		PrevRandao:  evmc.Hash(params.PrevRandao),
+		ChainID:     evmc.Hash(params.ChainID),
+		BaseFee:     evmc.Hash(params.BaseFee),
+		BlobBaseFee: evmc.Hash(params.BlobBaseFee),
+		BlobHashes:  ctx.EvmcBlobHashes,
```
HerbertJordan, line 238: The `./third_party/evmc` submodule needs to be updated to include this field.

go/vm/evmc/evmc_steppable_interpreter.go
```diff
@@ -62,6 +62,12 @@ func (e *SteppableEvmcInterpreter) StepN(
 		return nil, err
 	}
 
+	evmcBlobHashes := []evmc.Hash{}
```
HerbertJordan, line 65: In cases where you know the resulting size of the slice it should be allocated accordingly to avoid growing steps:
```suggestion
	evmcBlobHashes := make([]evmc.Hash, 0, len(params.BlobHashes))
```

go/vm/evmc/evmc_steppable_interpreter.go
```diff
@@ -62,6 +62,12 @@ func (e *SteppableEvmcInterpreter) StepN(
...
+	for _, hash := range params.BlobHashes {
+		evmcBlobHashes = append(evmcBlobHashes, evmc.Hash(hash[:]))
```
HerbertJordan, line 67: nit:
```suggestion
		evmcBlobHashes = append(evmcBlobHashes, evmc.Hash(hash))
```

go/vm/evmc/evmc_steppable_interpreter.go
```diff
@@ -62,6 +62,12 @@ func (e *SteppableEvmcInterpreter) StepN(
...
+	}
+	host_ctx.EvmcBlobHashes = evmcBlobHashes
```
HerbertJordan, line 69: nit: can you move this up to line 48 to keep the struct initialization together?

HerbertJordan (changes requested):
Looks good to me. Please extend this change also to the `Run` case, to support this feature also in a setup that does not use `StepN`.

go/vm/evmc/evmc_interpreter.go
facuMH: yes, how should this need for update be reflected ?

go/vm/evmc/evmc_steppable_interpreter.go
```diff
@@ -46,6 +46,12 @@ func (e *SteppableEvmcInterpreter) StepN(
 		context: params.Context,
 	}
 
+	evmcBlobHashes := make([]evmc.Hash, 0, state.CallData.Length())
+	for _, hash := range params.BlobHashes {
+		evmcBlobHashes = append(evmcBlobHashes, evmc.Hash(hash))
+	}
+	host_ctx.evmcBlobHashes = evmcBlobHashes
```
LuisPH3, line 53: I am not sure if this solves the problem, and if it does I have concerns about stability and being idiomatic:
- On the one hand host_ctx will keep the member pointer (slice) evmcBlobHashes alive over the StepN call. 
- On the other hand cgo **requires** to pin pointers passed to C, and I am afraid that they do check if the memory is pinned before the call.  

I guess it is not enough with having the variable bound and alive during C program execution because cgo wants to provide extra guarantees (what happens if the C side keeps a copy of that pointer after the C function has been called? )

Reference: https://pkg.go.dev/cmd/cgo
```
All Go pointers passed to C must point to pinned Go memory. Go pointers passed as function arguments to C 
functions have the memory they point to implicitly pinned for the duration of the call. Go memory reachable 
from these function arguments must be pinned as long as the C code has access to it. Whether Go memory is 
pinned is a dynamic property of that memory region; it has nothing to do with the type of the pointer. 

```

go/vm/evmc/evmc_interpreter.go
```diff
@@ -155,8 +155,9 @@ func toEvmcRevision(revision vm.Revision) (evmc.Revision, error) {
 // context and chain state information external to the the interpreter.
 // It implements the host interface of evmc's Go bindings.
 type hostContext struct {
-	params  vm.Parameters
-	context vm.RunContext
+	params         vm.Parameters
+	context        vm.RunContext
+	evmcBlobHashes []evmc.Hash
```
LuisPH3, line 160: I don't think this interface is clean. I understand where does this come from, but I don't think we should be mixing vm and evmc types like this.

go/vm/evmc/evmc_steppable_interpreter.go
LuisPH3: Regarding readability, this is not the place to convert/copy data, neither. 
The responsibility of this function is to trigger the correct conversions and handle logic flow. Data transformations belong somewhere else.

LuisPH3 (changes requested):
I don't think this solves the issue:

- cgo requires pinning.
- evmc shall data shall not be mixed with vm data, specially when converted ad-hoc in a new code location.

LuisPH3 (changes requested):
Now that I was pointed out to https://github.com/Fantom-foundation/evmc/pull/1/files I can see that memory is pinned. Nevertheless, I don't think this is a good solution:

- Allocation of the evmc type in the host context is ad-hoc. It does not share any property with the other members other than lifetime.
- Obscure mechanisms: we rely on someone doing the right thing, but still we need to declare early enough because we know what is going to happen. No encapsulation, this won't scale either in the future with other implementations. 
- Unintuitive: the code happens in complete different areas of responsibility.

IMO a more solid implementation would require the following changes:
- pinner is a property of the evmc host interface, allowing to access it from re-entrant callbacks. This allows us to continue doing data transformation at the same locations where we used to. In the callback, from parameters to evmc types. 
- evmc provides and releases the pinner
- code transformation goes back to the GetTxContext function and is no longer constructed out of place. 

These changes will make sure that the next time we have a memory issue across the boundary, we have the tools to resolve without having to patch the interface again. 

I am willing to compromise in this PR, but I want to make sure that we all understand that this change will not solve future issues and may produce extra confusion.

LuisPH3:
As discussed offline, IMO the future prof option would be to carry on with [#420](https://github.com/Fantom-foundation/Tosca/issues/420) and use that opaque context to pass a pinner when necessary.

go/vm/evmc/evmc_interpreter.go
HerbertJordan: It is a git submodule. You need to update by
```
cd ./third_party/evmc
git checkout <version you need>
```
and then add it to this commit by
```
cd <tosca directory>
git add ./third_party/evmc
git commit -m "Update evmc dependency"
```

facuMH: ah right, I thought you meant a comment or something else. But before the git submodule change, another PR needs to be merged in the dependency repo.

HerbertJordan (dismissed):
Thanks for adding support for this and including the tests!

HerbertJordan (approved):

PR #555, opened 2024-06-25
Consolidate Value type utilities

HerbertJordan:
This PR is consolidating all `Value` type utilities into the `types.go` file and adds additional utility functions required by the upcoming opera EVM Processor port.

New features added:
- conversion to `big.Int` and `uint256.Int`
- comparison between values
- addition and subtraction of operations on values

Changes:
- the print format of values now uses decimal numbers instead of fixed-length hex values
- the function `Uint256ToValue` is renamed to `ValueFromUint256`

Mid-term we should consider defining a common library for basic types like `Value`, `Address`, `Key` and `Hash` to be shared among all Fantom projects.


go/vm/types.go
```diff
@@ -29,8 +32,42 @@ func (a *Address) UnmarshalText(data []byte) error {
...
+	return new(uint256.Int).SetBytes(v[:])
+}
+
 func (v Value) String() string {
-	return fmt.Sprintf("0x%x", v[:])
+	return v.ToUint256().String()
+}
+
+func (v Value) Cmp(o Value) int {
+	return v.ToUint256().Cmp(o.ToUint256())
```
facuMH, line 48: does this means 2 copies for every call to `Cmp` ?

go/vm/types.go
```diff
@@ -29,8 +32,42 @@ func (a *Address) UnmarshalText(data []byte) error {
...
+}
+
+func ValueFromUint64(value uint64) Value {
+	return ValueFromUint256(new(uint256.Int).SetUint64(value))
```
facuMH, line 52:
```suggestion
	return ValueFromUint256(value.ToUint256())
```
Would this also work ?

go/vm/types.go
```diff
@@ -29,8 +32,42 @@ func (a *Address) UnmarshalText(data []byte) error {
...
+func ValueFromUint256(value *uint256.Int) (result Value) {
+	if value == nil {
+		return result
+	}
+	result = value.Bytes32()
+	return result
+}
+
+func Add(a, b Value) Value {
+	return ValueFromUint256(new(uint256.Int).Add(a.ToUint256(), b.ToUint256()))
```
facuMH, line 66:
```suggestion
	return ValueFromUint256(a.ToUint256().Add(b.ToUint256())))
```
Would this also work?  same for sub ?

facuMH (approved):
looks good, just a couple of suggestions. I like the idea of centralizing types for all our projects.

go/vm/types.go
HerbertJordan: No, the `value` here has type `uint64`.

go/vm/types.go
HerbertJordan: There is no `Add` function defined for the `*uint256.Int` type that accepts a single input. Only the version with two inputs.

go/vm/types.go
HerbertJordan: Yes, I went for the simple solution and would have looked into performance issues as they are showing up. But it is easy to fix, to get rid of the copies. I made the corresponding change.

facuMH (approved):

simonlechner (approved):

PR #558, opened 2024-06-26
CT: IsAddressWarm/Cold small refactor & benchmark

facuMH:
This PR is a response to the test `TestSpecification_EachRuleProducesAMatchingTestCase` taking too long for the development cycle compared to other tests. 
Running the commands
```
go test ./go/ct/spc -run ^TestSpecification_EachRuleProducesAMatchingTestCase$ -cpuprofile cpu.log
go tool pprof -http "localhost:8000" ./cpu.log
```
and a followed analysis showed  that `IsAddressWarm` and the `Cold` counterpart were taking about 21% of the whole run. 
Unfortunately the changes proposed in this PR only lowered this to 17% of the full run of the test, but this results are not shown in the benchmark comparisons. 

That said, I do think these changes do help code maintainability.

### DIFF
--- a/go/ct/rlz/conditions.go
+++ b/go/ct/rlz/conditions.go
@@ -713,11 +713,8 @@ func IsAddressCold(key BindableExpression[U256]) Condition {
 }
 
 func (c *isAddressCold) Check(s *st.State) (bool, error) {
-	key, err := c.key.Eval(s)
-	if err != nil {
-		return false, err
-	}
-	return s.Accounts.IsCold(NewAddress(key)), nil
+	res, err := IsAddressWarm(c.key).Check(s)
+	return !res, err
 }
 
 func (c *isAddressCold) Restrict(generator *gen.StateGenerator) {
--- a/go/ct/rlz/conditions_test.go
+++ b/go/ct/rlz/conditions_test.go
@@ -611,6 +611,7 @@ func BenchmarkCondition_IsAddressWarmCheckWarm(b *testing.B) {
 func BenchmarkCondition_IsAddressWarmCheckCold(b *testing.B) {
 	state := st.NewState(st.NewCode([]byte{}))
 	state.Accounts.MarkWarm(NewAddress(NewU256(42)))
+	state.Stack.Push(NewU256(1))
 	condition := IsAddressWarm(Param(0))
 
 	b.ResetTimer()
--- a/go/ct/rlz/conditions_test.go
+++ b/go/ct/rlz/conditions_test.go
@@ -595,3 +595,26 @@ func TestCondition_BlobHashesProducesGetTestValues(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkCondition_IsAddressWarmCheckWarm(b *testing.B) {
+	state := st.NewState(st.NewCode([]byte{}))
+	state.Accounts.MarkWarm(NewAddress(NewU256(42)))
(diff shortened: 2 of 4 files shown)


go/ct/rlz/conditions.go
```diff
@@ -669,7 +669,7 @@ func (c *isAddressWarm) Check(s *st.State) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return s.Accounts.IsWarm(NewAddress(key)), nil
+	return s.Accounts.IsWarm(key.Bytes20be()), nil
```
HerbertJordan, line 672: This is weakening readability. I would instead update the `NewAddress` implementation to use `Bytes20be()`. This may also be beneficial for other places where `NewAddress` is used.

Also, the effect of this change seems to be noise ... at least on my system there is no significant difference:
```
cpu: Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
                                   │   old.log   │              new.log               │
Condition_IsAddressWarmCheckWarm-4   41.70n ± 8%   42.02n ± 4%       ~ (p=0.684 n=10)
Condition_IsAddressWarmCheckCold-4   16.64n ± 6%   16.57n ± 2%       ~ (p=0.739 n=10)
geomean                              26.34n        26.39n       +0.20%

```

go/ct/rlz/conditions_test.go
```diff
@@ -463,3 +463,26 @@ func TestCondition_RestrictTransientStorageAndCheck(t *testing.T) {
...
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		condition.Check(state)
+	}
+}
+
+func BenchmarkCondition_IsAddressWarmCheckCold(b *testing.B) {
+	state := st.NewState(st.NewCode([]byte{}))
+	state.Accounts.MarkWarm(NewAddress(NewU256(42)))
+	condition := IsAddressWarm(Param(0))
```
HerbertJordan, line 482: This benchmark checks the same property as the one above.

HerbertJordan, line 482: This test case does not set up `Param(0)` on the stack, so what it actually tests is how fast it can check that the stack is empty.

go/ct/rlz/expressions.go
```diff
@@ -326,11 +326,10 @@ func (p param) Property() Property { return Property(p.String()) }
 func (p param) Domain() Domain[U256] { return p.domain }
 
 func (p param) Eval(s *st.State) (U256, error) {
-	stack := s.Stack
-	if p.position >= stack.Size() {
+	if p.position >= s.Stack.Size() {
 		return NewU256(0), ErrStackOutOfBoundsAccess
 	}
-	return stack.Get(p.position), nil
+	return s.Stack.Get(p.position), nil
```
HerbertJordan, line 332: Can you verify with a benchmark that this really makes a difference?

HerbertJordan (changes requested):
I like the simplifications of the conditions checks calling their complement as well as the simplification of the `IsCold` check in the accounts, but I do not think any of those changes have a significant performance impact.

go/ct/rlz/conditions_test.go
facuMH: the intent was to check the performance of `isAddressWarm`  and observe if the result makes a difference in performance

HerbertJordan (approved):

PR #559, opened 2024-06-26
CT: reusable sample slices

facuMH:
This PR is a response to the test `TestSpecification_EachRuleProducesAMatchingTestCase` taking too long for the development cycle compared to other tests.
Running the commands
```
go test ./go/ct/spc -run ^TestSpecification_EachRuleProducesAMatchingTestCase$ -cpuprofile cpu.log
go tool pprof -http "localhost:8000" ./cpu.log
````
and a followed analysis showed that `MemorySizeParameter.Samples()`  took about 30% of the whole run.
Making the slice a global variable to prevent multiple allocations reduced the runtime of the whole test in ~45%
Making the slice for `MemoryOffsetParameter.Samples()` reusable as well reduced another ~5%. 
The rest of the parameters are made reusable as well for consistency. 

This change is "safe" only because the `.Samples` methods are only called in a `for _, value := range param.Samples()` so the slices can not be modified. A comment is added to reflect this warning.

### DIFF
--- a/go/ct/rlz/parameters.go
+++ b/go/ct/rlz/parameters.go
@@ -15,81 +15,95 @@ import (
 )
 
 type Parameter interface {
+	// Samples returns a list of test values for an operation parameter.
+	// For efficiency reasons, the resulting slice may be shared among
+	// multiple calls. Thus, neither the slice itself nor its elements must
+	// be modified by users of this function.
 	Samples() []U256
 }
 
 type NumericParameter struct{}
 
+var numericParameterSamples = []U256{
+	NewU256(0),
+	NewU256(1),
+	NewU256(1 << 8),
+	NewU256(1 << 16),
+	NewU256(1 << 32),
+	NewU256(1 << 48),
+	NewU256(1).Shl(NewU256(64)),
+	NewU256(1).Shl(NewU256(128)),
+	NewU256(1).Shl(NewU256(192)),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: go/ct/rlz/parameters.go [hunk 1]
```diff
@@ -15,105 +15,123 @@ import (
 )
 
 type Parameter interface {
+
+	// Samples method returns a reference to global slices,
+	// since Go can not insure that slices are not modified
+	// this has to be insured by the users of this interface.
```
HerbertJordan, line 21:
```suggestion
	// Samples returns a list of test values for an operation parameter.
	// For efficiency reasons, the resulting slice may be shared among
	// multiple calls. Thus, neither the slice itself nor its elements must
	// be modified by users of this function.
```

go/ct/rlz/parameters.go
```diff
@@ -15,105 +15,123 @@ import (
...
 	Samples() []U256
 }
 
 type NumericParameter struct{}
 
+var NumericParameterSamples = []U256{
```
HerbertJordan, line 27: This and all other lists should be private.

HerbertJordan (changes requested):

simonlechner (approved):
Nice change with surprisingly big impact on the runtime.

HerbertJordan (approved):

PR #563, opened 2024-06-28
CT: EachRuleProducesAMatchingTestCase flaky test

### LINKED ISSUE #550: Failing Specification Test: EachRuleProducesAMatchingTestCase
During the development of #540 the specification unit test `TestSpecification_EachRuleProducesAMatchingTestCase` started to fail due to a missing rule for `sstore_regular_London_StorageDeleted_cold`. A quick fix to hid the problem has been implemented to stop the CI from failing. The reason for this failure should be investigated and fixed.

facuMH:
This PR fixes some of the state generation issues that made the `TestSpecification_EachRuleProducesAMatchingTestCase` test randomly fail (depending on the state of the RNG). 

Also the `TestSpecification_SpecifiedRuleProducesMatchingTestCases` test is there to use iteratively in case any other rule fails as well, and to debug the generation. 

The 2 specific issues thar are solved in this changes are:
- limiting code size to be bigger than 2 if we are restrained by `isData` condition (if we need data there has to be at least one op followed by data)
- randomly setting an address as warm, but not "cleaning" it when "isStorageCold" condition is also present.

This PR fixes #550

### DIFF
--- a/go/ct/gen/code.go
+++ b/go/ct/gen/code.go
@@ -143,6 +143,9 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 		// extend the runtime but are expected to reveal limited extra code coverage.
 		const expectedSize float64 = 200
 		size = int(rnd.ExpFloat64()/(1/expectedSize)) + minSize
+		if len(g.varIsDataConstraints) > 0 && size < 2 {
+			size = 2
+		}
 		if size > st.MaxCodeSize {
 			size = st.MaxCodeSize
 		}
@@ -375,7 +378,7 @@ func (s *varCodeConstraintSolver) fits(pos int, op OpCode) bool {
 // maximum is 33 since this is the largest instruction we have.
 func (s *varCodeConstraintSolver) largestFit(pos int) int {
 	n := 0
-	for ; n <= 33 && pos+n < s.codeSize; n++ {
+	for ; n < 33 && pos+n < s.codeSize; n++ {
 		if s.usedPositions[pos+n] != isUnused {
 			break
 		}
--- a/go/ct/gen/code.go
+++ b/go/ct/gen/code.go
@@ -143,6 +143,8 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 		// extend the runtime but are expected to reveal limited extra code coverage.
 		const expectedSize float64 = 200
 		size = int(rnd.ExpFloat64()/(1/expectedSize)) + minSize
+		// if there are data constraints, we need at least 2 bytes,
+		// one for an op, and the other for data.
 		if len(g.varIsDataConstraints) > 0 && size < 2 {
 			size = 2
 		}
--- a/go/ct/gen/code.go
+++ b/go/ct/gen/code.go
@@ -132,6 +132,12 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 		minSize = 1
 	}
 
+	// if there are data constraints, we need at least 2 bytes,
+	// one for an op, and the other for data.
+	if len(g.varIsDataConstraints) > 0 && minSize < 2 {
+		minSize = 2
+	}
+
 	var size int
 	if g.codeSize != nil {
 		if *g.codeSize < minSize {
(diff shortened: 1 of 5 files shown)


simonlechner (approved):

go/ct/gen/code.go
```diff
@@ -143,6 +143,11 @@ func (g *CodeGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.Cod
 		// extend the runtime but are expected to reveal limited extra code coverage.
 		const expectedSize float64 = 200
 		size = int(rnd.ExpFloat64()/(1/expectedSize)) + minSize
+		// if there are data constraints, we need at least 2 bytes,
+		// one for an op, and the other for data.
+		if len(g.varIsDataConstraints) > 0 && size < 2 {
```
HerbertJordan, line 148: I would like to see a unit test in `code_test.go` that fails before this modification and passes after.

go/ct/gen/code.go
```diff
@@ -375,7 +380,7 @@ func (s *varCodeConstraintSolver) fits(pos int, op OpCode) bool {
 // maximum is 33 since this is the largest instruction we have.
 func (s *varCodeConstraintSolver) largestFit(pos int) int {
 	n := 0
-	for ; n <= 33 && pos+n < s.codeSize; n++ {
+	for ; n < 33 && pos+n < s.codeSize; n++ {
```
HerbertJordan, line 383: Same for this, can you add a unit test?

go/ct/gen/storage.go
```diff
@@ -225,7 +225,6 @@ func (g *StorageGenerator) Generate(assignment Assignment, rnd *rand.Rand) (*st.
 
 		builder.SetOriginal(key, orgValue)
 		builder.SetCurrent(key, curValue)
-		builder.SetWarm(key, rnd.Intn(2) == 1)
```
HerbertJordan, line 228: Also here, you should have a unit test for the storage generator that stress-tests this situation.

go/ct/spc/specification_test.go
```diff
@@ -101,10 +101,14 @@ func TestSpecification_EachRuleProducesAMatchingTestCase(t *testing.T) {
 	}
 }
 
-func TestSpecification_SstoreWithTooLittleGasProducesMatchingTestCases(t *testing.T) {
+// the following test has been used to verify stability of a specific rule.
+// so far it has been used for:
+// - sstore_with_too_little_gas_
+// - pc_on_data_is_ignored
```
HerbertJordan, line 107: nit: This is code comment not documentation and should be placed inside the function

go/ct/spc/specification_test.go
```diff
@@ -101,10 +101,14 @@ func TestSpecification_EachRuleProducesAMatchingTestCase(t *testing.T) {
...
+func TestSpecification_SpecifiedRuleProducesMatchingTestCases(t *testing.T) {
 	rnd := rand.New(0)
 	rules := Spec.GetRules()
-	filter := regexp.MustCompile("sstore_with_too_little_gas_")
+	filter := regexp.MustCompile("sstore_with_too_little_gas_Berlin_StorageDeleted_cold")
```
HerbertJordan, line 111: How about making this a table-driven test where you list the rules this test should cover?

This way, you would preserve the coverage of the old rule, add coverage for the new rule, and make future re-use easy.

HerbertJordan (changes requested):

go/ct/gen/code_test.go
```diff
@@ -599,3 +601,54 @@ func TestVarCodeConstraintSolver_fitsOnUsed(t *testing.T) {
...
+			generator := NewCodeGenerator()
+			generator.AddIsData(variable)
+			generator.codeSize = new(int)
+			*generator.codeSize = size
+
+			assignment := Assignment{}
+
+			state, err := generator.Generate(assignment, rand.New(0))
+			if size < 2 {
+				if !strings.Contains(err.Error(), ErrUnsatisfiable.Error()) {
```
HerbertJordan, line 622: nit
```suggestion
				if !errors.Is(err, ErrUnsatisfiable) {
```

HerbertJordan (dismissed):

go/ct/gen/code_test.go
facuMH: error from code generator in case of test contains more info than just error. https://github.com/Fantom-foundation/Tosca/blob/main/go/ct/gen/code.go#L138

HerbertJordan: The line you are referencing is annotating an `ErrUnsatisifiable` with extra information, but it is still identifiable as such. Please study the [errors](https://pkg.go.dev/errors) package and the `%w` verb of [fmt.Errorf](https://pkg.go.dev/fmt#Errorf).

HerbertJordan (approved):

PR #564, opened 2024-07-01
CT: improve lfvm coverage

facuMH:
Thanks to the recent coverage command we could observe certain edge cases not being covered.
The changes in this PR enable the generation of states that would reach those edge cases however unlikely they might be in a real use case. 

In particular:
- extra values in `GasDomain` and `MemoryOffsetParameter` enable `EnsureCapacity` to fail because there is enough Gas to execute a certain rule, but not enough to expand it with reasonable size/offset. 
- `NumericParameters` is there because any instruction that pops should have a defined set of parameter values.

### DIFF
--- a/go/ct/rlz/domains.go
+++ b/go/ct/rlz/domains.go
@@ -374,7 +374,7 @@ func (d gasDomain) Samples(a vm.Gas) []vm.Gas {
 }
 
 func (gasDomain) SamplesForAll(as []vm.Gas) []vm.Gas {
-	res := []vm.Gas{0, st.MaxGas}
+	res := []vm.Gas{0, 100, 200, st.MaxGas}
 
 	// Test every element off by one.
 	for _, a := range as {
--- a/go/ct/rlz/parameters.go
+++ b/go/ct/rlz/parameters.go
@@ -36,6 +36,7 @@ var numericParameterSamples = []U256{
 	NewU256(1).Shl(NewU256(192)),
 	NewU256(1).Shl(NewU256(255)),
 	NewU256(0).Not(),
+	NewU256(1, 1),
 }
 
 func (NumericParameter) Samples() []U256 {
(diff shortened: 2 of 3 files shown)


go/ct/rlz/domains.go
```diff
@@ -374,7 +374,7 @@ func (d gasDomain) Samples(a vm.Gas) []vm.Gas {
 }
 
 func (gasDomain) SamplesForAll(as []vm.Gas) []vm.Gas {
-	res := []vm.Gas{0, st.MaxGas}
+	res := []vm.Gas{0, 100, 200, st.MaxGas}
```
simonlechner, line 377: Since almost all rules have a condition on the gas this change will almost double the amount of test cases we produce. It should be investigated whether a change with less performance impact would be sufficient.

simonlechner (dismissed):

go/ct/rlz/domains.go
facuMH: this is a great observation, I thought it would have an impact but not as big as it did. After a little experimenting, I removed the `100`, but in order to hit those edge cases an intermediate value independent from the rules' base cost is needed.

HerbertJordan (approved):

go/ct/rlz/parameters.go
```diff
@@ -52,6 +53,10 @@ var memoryOffsetParameterSamples = []U256{
 	NewU256(31),
 	NewU256(32),
 	NewU256(1, 0),
+	// Samples stressing the max init code size introduced with Shanghai
+	NewU256(2*24576 - 1),
+	NewU256(2 * 24576),
+	NewU256(2*24576 + 1),
```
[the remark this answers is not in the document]
facuMH: since they were added to stress edge cases of `MCOPY` I used it for `MLOAD` and `MSTORE` as well.

HerbertJordan (approved):

PR #586, opened 2024-07-04
Tosca: unify revision type

facuMH:
This PR unifies the `Revision` type and enforces its use in all packages removing the need for a conversion between `ct` and `vm`  revision.


go/tosca/interpreter.go
```diff
@@ -190,6 +190,7 @@ const (
 	R11_Paris
 	R12_Shanghai
 	R13_Cancun
+	R99_UnknownNextRevision
```
HerbertJordan, line 193: Since the `R99_UnknownNextRevision` is not a real revision, I think it should be defined in `go/ct/common/revisions.go` together with other constants like `Min` and `Max` revision. It is OK if this leads to a print that looks like `Revision(99)` instead of a `UnknownNextRevision`.

go/ct/utils/adapter.go
```diff
@@ -32,17 +32,17 @@ func ToVmParameters(state *st.State) tosca.Parameters {
...
+	case tosca.R11_Paris:
 		revision = tosca.R11_Paris
-	case cc.R12_Shanghai:
+	case tosca.R12_Shanghai:
 		revision = tosca.R12_Shanghai
-	case cc.R13_Cancun:
+	case tosca.R13_Cancun:
 		revision = tosca.R13_Cancun
 	default:
 		revision = tosca.Revision(state.Revision)
```
HerbertJordan, line 48: I guess this entire switch can be replaced by a a simple
```
revision := state.Revision
```

HerbertJordan (changes requested):
Looks good, thanks for refactoring this. I have two minor remarks that I think should be addressed before merging. Please have a look.

go/ct/common/revisions.go
```diff
@@ -11,110 +11,38 @@
...
 
-// Newest Revision currently supported by the CT specification
-const NewestSupportedRevision = R13_Cancun
-const NewestFullySupportedRevision = R12_Shanghai
+// Newest tosca.Revision currently supported by the CT specification
+const NewestSupportedRevision = tosca.R13_Cancun
+const NewestFullySupportedRevision = tosca.R12_Shanghai
 
-const MinRevision = R07_Istanbul
+const R99_UnknownNextRevision = NewestSupportedRevision + 1
```
HerbertJordan, line 24: Since the revision is stored in JSON files as part of regression tests, this should be a constant (also over time). Thus, when new revisions are added, this should not collide with new revision numbers nor being altered.

I'd suggest to assign it the value 99, as it is prefixed with.

HerbertJordan (changes requested):

HerbertJordan (approved):

go/ct/rlz/conditions.go
```diff
@@ -314,23 +314,23 @@ func (c *ge[T]) String() string {
...
 
-func IsRevision(revision Revision) Condition {
+func IsRevision(revision tosca.Revision) Condition {
 	return RevisionBounds(revision, revision)
 }
 
 // AnyKnownRevision restricts the revision to any revision covered by the CT specification.
 func AnyKnownRevision() Condition {
-	return RevisionBounds(Revision(0), NewestSupportedRevision)
+	return RevisionBounds(tosca.Revision(0), NewestSupportedRevision)
```
HerbertJordan, line 333: nit: `tosca.Revision(0)` => `MinRevision`

HerbertJordan (approved):

simonlechner (approved):

HerbertJordan (approved):

simonlechner (approved):

PR #590, opened 2024-07-04
CT: Reduce number of produced test cases

simonlechner:
By adding the support for new revisions the number of generated test cases during a CT run increased to an unmanageable number. By evaluating the specification, conditions on read only mode, warm accounts and revisions could be changed. The number of test cases is reduced from 510.530.430.373 to 18.288.391.943 by these changes.

### DIFF
--- a/go/ct/rlz/conditions.go
+++ b/go/ct/rlz/conditions.go
@@ -328,6 +328,11 @@ func IsRevision(revision tosca.Revision) Condition {
 	return RevisionBounds(revision, revision)
 }
 
+func IsRevisionCondition(condition Condition) bool {
+	_, ok := condition.(*revisionBounds)
+	return ok
+}
+
 // AnyKnownRevision restricts the revision to any revision covered by the CT specification.
 func AnyKnownRevision() Condition {
 	return RevisionBounds(MinRevision, NewestSupportedRevision)
@@ -347,10 +352,19 @@ func (e *revisionBounds) GetTestValues() []TestValue {
 	restrict := func(generator *gen.StateGenerator, revision tosca.Revision) {
 		generator.SetRevision(revision)
 	}
-	res := []TestValue{NewTestValue(property, domain, R99_UnknownNextRevision, restrict)}
+	res := []TestValue{}
+	// If the revision is set to a specific value, only test this value,
+	// except if it is the unknown next revision.
+	if e.min == e.max && e.min != R99_UnknownNextRevision {
+		res = append(res, NewTestValue(property, domain, e.min, restrict))
+		return res
+	}
+
+	res = append(res, NewTestValue(property, domain, R99_UnknownNextRevision, restrict))
 	for r := tosca.Revision(0); r <= NewestSupportedRevision; r++ {
 		res = append(res, NewTestValue(property, domain, r, restrict))
 	}
+
 	return res
 }
 
@@ -680,25 +694,14 @@ func (c *isAddressWarm) Restrict(generator *gen.StateGenerator) {
 }
 
 func (c *isAddressWarm) GetTestValues() []TestValue {
-	property := Property(c.String())
-	domain := boolDomain{}
-	restrict := func(generator *gen.StateGenerator, isWarm bool) {
-		key := c.key.GetVariable()
-		c.key.BindTo(generator)
-		if isWarm {
-			generator.BindToWarmAddress(key)
-		} else {
-			generator.BindToColdAddress(key)
-		}
-	}
+	property := Property(fmt.Sprintf("warm(%v)", c.key))
 	return []TestValue{
-		NewTestValue(property, domain, true, restrict),
-		NewTestValue(property, domain, false, restrict),
+		NewTestValue(property, boolDomain{}, true, restrictAccountWarmCold(c.key)),
 	}
 }
 
 func (c *isAddressWarm) String() string {
-	return fmt.Sprintf("warm(%v)", c.key)
+	return fmt.Sprintf("account warm(%v)", c.key)
 }
 
 ////////////////////////////////////////////////////////////
@@ -724,11 +727,26 @@ func (c *isAddressCold) Restrict(generator *gen.StateGenerator) {
 }
 
 func (c *isAddressCold) GetTestValues() []TestValue {
-	return IsAddressWarm(c.key).GetTestValues()
+	property := Property(fmt.Sprintf("warm(%v)", c.key))
+	return []TestValue{
+		NewTestValue(property, boolDomain{}, false, restrictAccountWarmCold(c.key)),
+	}
 }
 
 func (c *isAddressCold) String() string {
-	return fmt.Sprintf("cold(%v)", c.key)
+	return fmt.Sprintf("account cold(%v)", c.key)
+}
+
+func restrictAccountWarmCold(bindKey BindableExpression[U256]) func(generator *gen.StateGenerator, isWarm bool) {
+	return func(generator *gen.StateGenerator, isWarm bool) {
+		key := bindKey.GetVariable()
+		bindKey.BindTo(generator)
+		if isWarm {
+			generator.BindToWarmAddress(key)
+		} else {
+			generator.BindToColdAddress(key)
+		}
+	}
 }
 
 ////////////////////////////////////////////////////////////
--- a/go/ct/rlz/conditions_test.go
+++ b/go/ct/rlz/conditions_test.go
@@ -11,6 +11,7 @@
 package rlz
 
 import (
+	"errors"
 	"math"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


go/ct/rlz/conditions.go
```diff
@@ -724,7 +732,20 @@ func (c *isAddressCold) Restrict(generator *gen.StateGenerator) {
 }
 
 func (c *isAddressCold) GetTestValues() []TestValue {
-	return IsAddressWarm(c.key).GetTestValues()
+	property := Property(c.String())
+	domain := boolDomain{}
+	restrict := func(generator *gen.StateGenerator, isWarm bool) {
+		key := c.key.GetVariable()
+		c.key.BindTo(generator)
+		if isWarm {
```
facuMH, line 740: could this ever be `true`? can the one in `IsAddressWarm` ever be false ? same for `Is/IsNotReadOnly`?

go/ct/rlz/conditions_test.go
```diff
@@ -596,6 +596,40 @@ func TestCondition_BlobHashesProducesGetTestValues(t *testing.T) {
...
+
+func TestCondition_IsReadOnlyUnsatisfiable(t *testing.T) {
+	isReadOnly := IsReadOnly()
+	isNotReadOnly := IsNotReadOnly()
+
+	gen := gen.NewStateGenerator()
+	isReadOnly.Restrict(gen)
+	isNotReadOnly.Restrict(gen)
+	_, err := gen.Generate(rand.New(0))
+	if err == nil {
```
facuMH, line 628: nit: `!errors.Is(err, ErrUnsatisfiable)`

facuMH (approved):
Looks good !! thanks for reducing this large number !! just a couple of comments.

go/ct/rlz/conditions.go
```diff
@@ -724,7 +734,20 @@ func (c *isAddressCold) Restrict(generator *gen.StateGenerator) {
 }
 
 func (c *isAddressCold) GetTestValues() []TestValue {
-	return IsAddressWarm(c.key).GetTestValues()
+	property := Property(c.String())
```
HerbertJordan, line 737: The "Property" targeted by the test value should be the same as for the `isAddressWarm` condition.

go/ct/rlz/conditions.go
```diff
@@ -724,7 +734,20 @@ func (c *isAddressCold) Restrict(generator *gen.StateGenerator) {
...
+		c.key.BindTo(generator)
+		if isWarm {
+			generator.BindToWarmAddress(key)
+		} else {
+			generator.BindToColdAddress(key)
+		}
+	}
+	return []TestValue{
+		NewTestValue(property, domain, false, restrict),
+	}
```
HerbertJordan, line 750: This code is mostly a duplicate of `isAddressWarm`'s `GetTestValue`. Please factor it out into a common utility function.

go/ct/rlz/conditions.go
```diff
@@ -962,3 +985,77 @@ func (c *hasNoBlobHash) String() string {
...
+func (c *isNotReadOnly) Check(s *st.State) (bool, error) {
+	return !s.ReadOnly, nil
+}
+
+func (c *isNotReadOnly) Restrict(generator *gen.StateGenerator) {
+	generator.SetReadOnly(false)
+}
+
+func (c *isNotReadOnly) GetTestValues() []TestValue {
+	property := Property(c.String())
```
HerbertJordan, line 1043: Same here, the positive (`IsReadOnly`) and negative (`IsNotReadOnly`) conditions should reference the same `property`. Also, please avoid code duplication by, for instance, providing a stand-alone version of this function accepting a `[]bool` listing all the `TestValue`s that should be used.

go/ct/rlz/conditions_test.go
```diff
@@ -596,6 +596,40 @@ func TestCondition_BlobHashesProducesGetTestValues(t *testing.T) {
...
+				t.Fatalf("failed to build state: %v", err)
+			}
+			if checked, err := condition.Check(state); err != nil || !checked {
+				t.Errorf("failed to restrict and check condition: %v", err)
+			}
+		})
+	}
+}
+
+func TestCondition_IsReadOnlyUnsatisfiable(t *testing.T) {
```
HerbertJordan, line 620:
```suggestion
func TestCondition_ConflictingReadOnlyConditionsAreUnsatisfiable(t *testing.T) {
```

go/ct/spc/specification.go
```diff
@@ -785,7 +786,7 @@ func getAllRules() []Rule {
 		},
 		conditions: []Condition{
 			RevisionBounds(R13_Cancun, NewestSupportedRevision),
-			Eq(ReadOnly(), false),
+			IsNotReadOnly(),
```
HerbertJordan, line 789: I might have never addressed this explicitly, but I never liked this pattern we established with the pairs of conditions like `IsCode` and `IsData`, or `IsStorageCold` and `IsStorageWarm`. I always thought, when everything is settled I would take a day or so and rewrite this into the form that `ReadOnly` has here, with an `Eq(ReadOnly(), false)` or `Eq(ReadOnly(), true)`. I think structurally this is the more concise form, with less types/code to be defined and tested, benefiting from composition.

Replacing `Eq(ReadOnly(), false)` here with an `IsNotReadOnly()` kinda goes against this idea. If the goal is to restrict the domain of test cases produced for the `ReadOnly` domain, I think this could be addressed differently.

The [ReadOnly](https://github.com/Fantom-foundation/Tosca/blob/dd9a9f0ee94d2f53198c33a8e79d7afbd8cfae3b/go/ct/rlz/expressions.go#L202) expression right now has the [boolDomain](https://github.com/Fantom-foundation/Tosca/blob/dd9a9f0ee94d2f53198c33a8e79d7afbd8cfae3b/go/ct/rlz/domains.go#L35) which when being [sampled](https://github.com/Fantom-foundation/Tosca/blob/dd9a9f0ee94d2f53198c33a8e79d7afbd8cfae3b/go/ct/rlz/domains.go#L46-L52) for test cases produces both cases -- `true` and `false`.

If for the static / read-only state the domain should be restricted, you could define a specialized domain
```
type readOnlyDomain struct {
	boolDomain
}

func (readOnlyDomain) Samples(value bool) []bool {
	return []bool{value}
}

func (readOnlyDomain) SamplesForAll(as []bool) []bool {
	return as
}
```
and use this domain for the `ReadOnly()` expression. This has the same effect of limiting the test cases to the value the `ReadOnly` property is restricted to.

HerbertJordan (changes requested):
I think there is a nicer way to reduce the domain of the `ReadOnly` flag. Please see the corresponding comment and let me know what you think.

go/ct/rlz/conditions.go
simonlechner: The restrict function is now reused for isAddressWarm and isAddressCold, therefore both cases are possible.

go/ct/rlz/conditions.go
simonlechner: The String() function new return the same string.

go/ct/rlz/conditions.go
```diff
@@ -724,11 +724,26 @@ func (c *isAddressCold) Restrict(generator *gen.StateGenerator) {
...
-	return IsAddressWarm(c.key).GetTestValues()
+	property := Property(c.String())
+	return []TestValue{
+		NewTestValue(property, boolDomain{}, false, restrictAccountWarmCold(c.key)),
+	}
 }
 
 func (c *isAddressCold) String() string {
-	return fmt.Sprintf("cold(%v)", c.key)
+	return fmt.Sprintf("account warm(%v)", c.key)
```
HerbertJordan, line 734: This makes `isAddressCold(param(0))` and `isAddressWarm(param(0))` indistinguishable when being listed e.g. as part of the `generator-info` command.

go/ct/rlz/conditions.go
HerbertJordan: That's not good (see comment for the `String` function). There is no need to link the `Property` to the string function of the condition. It can be something like
```
Property(fmt.Sprintf("warm(%v)", c.key))
```
here and for the complement constraint.

HerbertJordan (changes requested):

go/ct/rlz/conditions_test.go
```diff
@@ -596,6 +597,40 @@ func TestCondition_BlobHashesProducesGetTestValues(t *testing.T) {
 	}
 }
 
+func TestCondition_IAccountWarmRestrictAndCheck(t *testing.T) {
```
facuMH, line 600:
```suggestion
func TestCondition_IsAccountWarmRestrictAndCheck(t *testing.T) {
```

facuMH (approved):
small typo, otherwise good.

go/ct/rlz/conditions.go
```diff
@@ -724,11 +727,26 @@ func (c *isAddressCold) Restrict(generator *gen.StateGenerator) {
 }
 
 func (c *isAddressCold) GetTestValues() []TestValue {
-	return IsAddressWarm(c.key).GetTestValues()
+	property := Property(fmt.Sprintf("warm(%v)", c.key))
```
facuMH, line 730: cold

facuMH (approved):
looks good, just one thing to correct

go/ct/rlz/conditions.go
simonlechner: As discussed with Herbert in the other conversations, the `property` of `isAddressWarm` and `isAddressCold` has to be the same because it is used to unify constraints on the same property.

HerbertJordan (approved):

PR #591, opened 2024-07-04
CT: extcodehash pre berlin

facuMH:
This PR improves the coverage of `EXTCODEHASH` rule by dividing the `preBerlin` rule into `warm` and `cold` cases, ensuring all branches are covered.

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -1341,9 +1341,11 @@ func getAllRules() []Rule {
 
 	// --- EXTCODEHASH ---
 
-	extCodeHashEffect := func(s *st.State, address tosca.Address) {
+	extCodeHashEffect := func(s *st.State) tosca.Address {
+		address := NewAddress(s.Stack.Pop())
 		hash := s.Accounts.GetCodeHash(address)
 		s.Stack.Push(NewU256FromBytes(hash[:]...))
+		return address
 	}
 
 	// cold
@@ -1361,8 +1363,7 @@ func getAllRules() []Rule {
 			AddressParameter{},
 		},
 		effect: func(s *st.State) {
-			address := NewAddress(s.Stack.Pop())
-			extCodeHashEffect(s, address)
+			address := extCodeHashEffect(s)
 			s.Accounts.MarkWarm(address)
 		},
 	})...)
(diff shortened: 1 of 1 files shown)


go/ct/spc/specification.go
```diff
@@ -1384,12 +1384,33 @@ func getAllRules() []Rule {
...
 			IsRevision(R07_Istanbul),
+			IsAddressWarm(Param(0)),
+		},
+		parameters: []Parameter{
+			AddressParameter{},
+		},
+		effect: func(s *st.State) {
+			address := NewAddress(s.Stack.Pop())
+			hash := s.Accounts.GetCodeHash(address)
+			s.Stack.Push(NewU256FromBytes(hash[:]...))
```
HerbertJordan, line 1401: I am not sure which branch coverage is increased by this change. Can you add a before / after to the description of the PR?

If the branch that was not covered was the first branch in
```
if !s.Accounts.Exist(address) {
  s.Stack.Push(NewU256(0))
} else {
  hash := s.Accounts.GetCodeHash(address)
  s.Stack.Push(NewU256FromBytes(hash[:]...))
}
```
wouldn't it be easiest to just replace the effect with
```
address := NewAddress(s.Stack.Pop())
hash := s.Accounts.GetCodeHash(address)
s.Stack.Push(NewU256FromBytes(hash[:]...))
```
and eliminate the branch that was not covered?

It seems odd to add a condition on whether an address is warm or cold to the rule if in the effect this property is not used. This kinda indicates that a `IsAddressWarm` condition has a side effect that is exploited here -- namely that any any warm account exists. This would be a bug in the constraint solver handling warm addresses. There may be warm addresses that not point to existing accounts.

HerbertJordan (changes requested):

go/ct/spc/specification.go
facuMH: after a brief conversation online, we agreed that the interpreter should not check whether the account exists or not, and simply get the code hash from the stateDB, and the latter should return 0 when needed.

go/ct/spc/specification.go
```diff
@@ -1334,6 +1334,11 @@ func getAllRules() []Rule {
 
 	// --- EXTCODEHASH ---
 
+	extCodeHashEffect := func(s *st.State, address tosca.Address) {
```
HerbertJordan, line 1337: nit: popping the address from the stack in the state could be also moved into this function.

HerbertJordan (dismissed):

HerbertJordan (approved):

simonlechner (approved):

PR #593, opened 2024-07-04
LFVM: remove unused functions and move super instructions to own file

facuMH:
This PR removes functions that are not used (as shown by coverage test results) and moves super instructions to separate file, facilitating a more accurate coverage of the original set of instructions. 

The removed instructions (`noop`, `data` and `invalid`) are never reached because of an early filter in the interpreter using `NUM_EXECUTABLE_OPCODES`

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -12,9 +12,7 @@ package lfvm
 
 import (
 	"bytes"
-	"errors"
 	"math"
-	"math/bits"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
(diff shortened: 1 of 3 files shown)


HerbertJordan (approved):

simonlechner (approved):
Did you check whether INVALID can be removed completely, since it is a defined opcode? Otherwise looks good.

facuMH:
> Did you check whether INVALID can be removed completely, since it is a defined opcode? Otherwise looks good.

the dispatcher table is just a switch case, that in the case of op being INVALID would never reach the switch, because it would do an early return in the if of line 395

PR #594, opened 2024-07-04
CT: Add rules for jump over int32

facuMH:
Coverage reported that there was no hit for the checks of jump/jumpi's destination overflow. 
This PR adds rules testing exactly that case.

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -592,6 +592,9 @@ func getAllRules() []Rule {
 
 		{
 			Name: "jump_to_invalid_destination",
+			Parameter: []Parameter{
+				JumpTargetParameter{},
+			},
 			Condition: And(
 				AnyKnownRevision(),
 				Eq(Status(), st.Running),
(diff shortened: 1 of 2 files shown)


HerbertJordan (changes requested):

go/ct/spc/specification.go
```diff
@@ -663,6 +666,10 @@ func getAllRules() []Rule {
 
 		{
 			Name: "jumpi_to_invalid_destination",
+			Parameter: []Parameter{
+				JumpTargetParameter{},
+				JumpTargetParameter{},
```
HerbertJordan, line 671: This is wrong, the second parameter is not the jump target, it is a condition for the jump. A good domain for this might be  a [NumericParameter](https://github.com/Fantom-foundation/Tosca/blob/dd9a9f0ee94d2f53198c33a8e79d7afbd8cfae3b/go/ct/rlz/parameters.go#L25C6-L25C22).

Please also fix this in related rules.

HerbertJordan (changes requested):

HerbertJordan (approved):

simonlechner (approved):

PR #596, opened 2024-07-05
LFVM: merge memory set/copy with EnsureCapacity

facuMH:
This PR aims to improve coverage by including calls to `EnsureCapacity` in functions that need it to not fail.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -169,7 +169,7 @@ func opMstore(c *context) {
 		c.status = ERROR
 		return
 	}
-	if err := c.memory.SetWord(offset, value, 32, c); err != nil {
+	if err := c.memory.SetWord(offset, value, c); err != nil {
 		c.SignalError(err)
 	}
 }
@@ -183,7 +183,7 @@ func opMstore8(c *context) {
 		c.status = ERROR
 		return
 	}
-	if err := c.memory.SetByte(offset, byte(value.Uint64()), 1, c); err != nil {
+	if err := c.memory.SetByte(offset, byte(value.Uint64()), c); err != nil {
 		c.SignalError(err)
 	}
 }
@@ -218,15 +218,13 @@ func opMcopy(c *context) {
 		return
 	}
 
-	if c.memory.EnsureCapacity(destOffset, size, c) != nil {
+	data, err := c.memory.GetSliceWithCapacityAndGas(srcOffset, size, c)
+	if err != nil {
 		return
 	}
-	if c.memory.EnsureCapacity(srcOffset, size, c) != nil {
+	if err := c.memory.SetWithCapacityAndGasCheck(destOffset, size, data, c); err != nil {
 		return
 	}
-
-	data := c.memory.GetSlice(srcOffset, size)
-	c.memory.Set(destOffset, size, data, c, false)
 }
 
 func opMload(c *context) {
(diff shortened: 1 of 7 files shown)


go/interpreter/lfvm/memory.go
```diff
@@ -167,7 +177,13 @@ func (m *Memory) SetWord(offset uint64, value *uint256.Int) error {
 	return nil
 }
 
-func (m *Memory) Set(offset, size uint64, value []byte) error {
+func (m *Memory) Set(offset, size uint64, value []byte, c *context, shouldEnsureCapacity bool) error {
```
HerbertJordan, line 180: I am not a fan of the `shouldEnsureCapacity` parameter. It is not very intuitive what this final boolean flag is supposed to be for at each call side. I would prefer it always to be enabled or, if this causes problems, have a dedicated `SetWithoutCapacityCheck(offset, size, value)` .

go/interpreter/lfvm/memory.go
```diff
@@ -180,7 +196,11 @@ func (m *Memory) Set(offset, size uint64, value []byte) error {
 	return nil
 }
 
-func (m *Memory) CopyWord(offset uint64, trg *uint256.Int) error {
+func (m *Memory) CopyWord(offset uint64, trg *uint256.Int, size uint64, c *context) error {
```
HerbertJordan, line 199: Also, here, since it is a word, size should be 32 - always.

go/interpreter/lfvm/memory.go
HerbertJordan, line 1: I would even go one step further, and integrate the `EnsureCapacity` function also in functions like `GetSlice` and try to eliminate any use as many calls to `EnsureCapacity` in instructions as possible.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory.go
facuMH: I had originally done so, but GetSlice is sometimes called after `EnsureCapacity` and sometimes after `EnsureCapacityWithoutGas`, hence adding an extra parameter to determine whether we cared about gas or not in that case. Also, `GetSlice` does not return any `error`, so it would either return `nil` when `EnsureCapacity` failed, hence hiding the actual error, or add the error handling after each call of `GetSlice`. Since did not particularly like either approach I decided to leave `GetSlice` as is, but if you have any suggestions, I would be glad to do so.

go/interpreter/lfvm/memory.go
facuMH: I removed the bool and added `SetWithoutCapacityCheck`. Used similar pattern for `GetSlice`

HerbertJordan (dismissed):
Looks good, but please have a look at the comment on the `opCreate` and verify that this is not causing problems.

simonlechner (approved):

go/interpreter/lfvm/instructions.go
```diff
@@ -1118,8 +1098,6 @@ func opCall(c *context) {
 
 	// first use static and dynamic gas cost and then resize the memory
 	// when out of gas is happening, then mem should not be resized
-	c.memory.EnsureCapacityWithoutGas(needed_memory_size, c)
```
HerbertJordan, line 1121: Removing this line here has the same problem as when it was removed in the create call.

go/interpreter/lfvm/instructions.go
```diff
@@ -1234,7 +1212,6 @@ func opCallCode(c *context) {
...
```
HerbertJordan, line 1237: Same here.

HerbertJordan (changes requested):

simonlechner (approved):

HerbertJordan (approved):

simonlechner (approved):

Issue #293, opened 2024-02-14
CT: Sparse memory


BlackMark29A:
The current memory implementation of the CT uses dense memory. This means that to allocate 32 bytes at a very large offset will require the entire memory up to this offset to also be allocated, even though only 32 bytes are actually required.

For very large offsets this is not an issue, because there will not be enough gas available to grow the memory to this size, and so this is expected to fail with "out of gas". However, there is a Goldilocks zone of states that can be generated, where there is enough gas, but there won't be enough memory on the target hardware, so that running this will result in an out of memory failure.

For now we are limiting the maximum gas to something that will result in states that can be executed on hardware, however a better solution would be to change the memory implementation to be sparse. This way any unused memory will not have to be allocated, and small buffers at arbitrary offsets will only have a negligible memory overhead.

simonlechner:
With the introduction of a gas upper bound we did not have a problem with the dense memory.

PR #598, opened 2024-07-08
CT: Max Memory Size Check

facuMH:
Thanks to the fuzzing experiments an unchecked overflow was found and fixed in LFVM memory expansion calculation function. 

This PRs add that fix in the analogous CT function and a corresponding test case that fails without this newly introduced check.

### DIFF
--- a/go/ct/st/memory.go
+++ b/go/ct/st/memory.go
@@ -83,9 +83,24 @@ func (m *Memory) Grow(offset, size uint64) {
 	}
 }
 
+const (
+	// Maximum memory size allowed
+	// This is the maximum value that, when squared, does not produce an overflow.
+	// This number comes from 'core/vm/gas_table.go' 'memoryGasCost' in geth
+	maxMemoryExpansionSize = 0x1FFFFFFFE0
+)
+
 // ExpansionCosts calculates the expansion costs for the given offset and size.
 // It does not grow memory. It also returns offset and size converted to uint64.
 func (m *Memory) ExpansionCosts(offset_u256, size_u256 U256) (memCost tosca.Gas, offset, size uint64) {
+	// static assert
+	const (
+		// Memory expansion cost is done using unsigned arithmetic,
+		// check for the maximum memory expansion size, not overflowing int64 after computing costs
+		maxInWords uint64 = (uint64(maxMemoryExpansionSize) + 31) / 32
+		_                 = int64(maxInWords*maxInWords/512 + 3*maxInWords)
+	)
+
 	if !size_u256.IsUint64() {
 		return math.MaxInt64, 0, 0
 	}
@@ -111,6 +126,12 @@ func (m *Memory) ExpansionCosts(offset_u256, size_u256 U256) (memCost tosca.Gas,
 		memCost = 0
 		return
 	}
+
+	if newSize > maxMemoryExpansionSize {
+		memCost = MaxGas
+		return
+	}
+
 	calcMemoryCost := func(size uint64) tosca.Gas {
 		memorySizeWord := size / 32
 		if size%32 != 0 {
@@ -118,6 +139,7 @@ func (m *Memory) ExpansionCosts(offset_u256, size_u256 U256) (memCost tosca.Gas,
 		}
 		return tosca.Gas((memorySizeWord*memorySizeWord)/512 + (3 * memorySizeWord))
 	}
+
 	memCost = calcMemoryCost(newSize) - calcMemoryCost(uint64(m.Size()))
 	return
 }
--- a/go/ct/st/memory_test.go
+++ b/go/ct/st/memory_test.go
@@ -139,15 +139,19 @@ func TestMemory_MemoryExpansionCosts(t *testing.T) {
 		wantOffset uint64
 		wantSize   uint64
 	}{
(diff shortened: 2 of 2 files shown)


go/ct/st/memory_test.go
```diff
@@ -139,15 +139,19 @@ func TestMemory_MemoryExpansionCosts(t *testing.T) {
...
+		"large size no overflow":            {NewU256(0), NewU256(math.MaxUint64), MaxGas, 0, math.MaxUint64},
+		"large offset no overflow":          {NewU256(math.MaxUint64 - 1), NewU256(1), MaxGas, math.MaxUint64 - 1, 1},
+		"large offset and size no overflow": {NewU256(math.MaxUint64 / 2), NewU256(math.MaxUint64 / 2), MaxGas, math.MaxUint64 / 2, math.MaxUint64 / 2},
 		"size overflow":                     {NewU256(0), NewU256(1, 0), tosca.Gas(math.MaxInt64), 0, 0},
 		"offset overflow":                   {NewU256(1, 0), NewU256(1), tosca.Gas(math.MaxInt64), 0, 0},
 		"large offset and size overflow":    {NewU256(math.MaxUint64/2 + 1), NewU256(math.MaxUint64/2 + 1), tosca.Gas(math.MaxInt64), math.MaxUint64/2 + 1, math.MaxUint64/2 + 1},
 		"zero size":                         {NewU256(0), NewU256(0), tosca.Gas(0), 0, 0},
 		"zero size offset":                  {NewU256(1024), NewU256(0), tosca.Gas(0), 1024, 0},
 		"zero size offset overflow":         {NewU256(1, 0), NewU256(0), tosca.Gas(0), 0, 0},
+		"max memory size allowed":           {NewU256(0), NewU256(0x1FFFFFFFE1), MaxGas, 0, 0x1FFFFFFFE1},
```
simonlechner, line 151: Maybe add a test case with the max memory size + 1 that should overflow.

simonlechner (dismissed):
Looks good.

go/ct/st/memory_test.go
HerbertJordan, line 151: To improve readability, I would use `maxMemoryExpansionSize` instead of `0x1FFFFFFFE0` and `maxMemoryExpansionSize+1` instead of `0x1FFFFFFFE1`

go/ct/st/memory_test.go
```diff
@@ -139,15 +139,19 @@ func TestMemory_MemoryExpansionCosts(t *testing.T) {
...
+		"acceptable size":                   {NewU256(0), NewU256(0x1FFFFFFFE0), tosca.Gas(36028809887088637), 0, 0x1FFFFFFFE0},
+		"acceptable offset":                 {NewU256(0x1FFFFFFFE0), NewU256(1), tosca.Gas(1152921504606846976), 0x1FFFFFFFE0, 1},
```
HerbertJordan, line 153: I do not understand this line. Why would this be acceptable?

go/ct/st/memory.go
```diff
@@ -83,9 +83,23 @@ func (m *Memory) Grow(offset, size uint64) {
 	}
 }
 
+const (
+	// Maximum memory size allowed
+	// This magic number comes from 'core/vm/gas_table.go' 'memoryGasCost' in geth
```
HerbertJordan, line 88: Please provide a cleaner definition for this value. Saying that it is a "magic number from some source" is not building up trust.

go/ct/st/memory.go
```diff
@@ -111,13 +125,20 @@ func (m *Memory) ExpansionCosts(offset_u256, size_u256 U256) (memCost tosca.Gas,
 		memCost = 0
 		return
 	}
+
+	if newSize > maxMemoryExpansionSize {
+		memCost = MaxGas
```
HerbertJordan, line 130: In the other cases where we have saturated memory costs we return `math.MaxInt64`. Here, it returns `MaxGas`. This is not consistent. I think this here should be `math.MaxInt64` as well.

Note that `MaxGas` is not `tosca.Gas(math.MaxInt64)`.

HerbertJordan (changes requested):

go/ct/st/memory_test.go
facuMH: the `maxMemoryExpansionSize` is `0x1FFFFFFFE0`, so this case is actually testing the "+1" case. I will use the variable to make it obvious.

go/ct/st/memory.go
facuMH: I am aware of the difference, I thought using `MaxGas` was better since it is smaller than `MaxInt64` but this is intentional, as described in #293. I think we should keep it consistent, yes, but that given that there is an existing issue about it, I thought sticking with `MaxGas` was the safer path. Should we remove the issue and remove all uses of `MaxGas` or stick with it and change other uses of `MaxInt64` to `MaxGas` ?

HerbertJordan (approved):
I belief you when you are saying this change is fixing something, but I do not see it in the PR. Can you add in the description something that lists a test case or a command that was not working before and is working now, that was not introduced specifically to check the function changed in this PR?

simonlechner (approved):

PR #600, opened 2024-07-09
tosca: change `ValueFromUint64` to `NewValue`

### LINKED ISSUE #589: tosca: rename functions like `ValueFromUint64` for `NewValue` or similar
Some functions are called in a lot of places and with long names reading code can get cumbersome when it could be simplified. 
For example: 
- `ValueFromUint64` to `NewValue` 

similar function names can be added here and should get their own PRs

facuMH:
This PR changes the old `ValueFromUint64` to `NewValue`, keeping the old representation and decimal printing, but allowing the create a new value from up to 4 `uint64` much like `NewU256` 

This PR fixes #589

### DIFF
--- a/go/tosca/types.go
+++ b/go/tosca/types.go
@@ -12,10 +12,12 @@ package tosca
 
 import (
 	"bytes"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"math/bits"
 	"strings"
 
 	"github.com/holiman/uint256"
@@ -57,8 +59,20 @@ func (v Value) Cmp(o Value) int {
 	return bytes.Compare(v[:], o[:])
 }
 
-func ValueFromUint64(value uint64) Value {
-	return ValueFromUint256(new(uint256.Int).SetUint64(value))
+// NewValue creates a new Value instance from up to 4 uint64 arguments. The
+// arguments are given in the order from most significant to least significant
+// by padding leading zeros as needed. No argument results in a value of zero.
+func NewValue(args ...uint64) (result Value) {
+	if len(args) > 4 {
+		panic("Too many arguments")
+	}
+	offset := 4 - len(args)
+	for i := 0; i < len(args) && i < 4; i++ {
+		start := (offset * 8) + i*8
+		end := start + 8
+		binary.BigEndian.PutUint64(result[start:end], args[i])
+	}
+	return
 }
 
 // ValueFromUint256 converts a *uint256.Int to a Value.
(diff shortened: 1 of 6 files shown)


go/tosca/types.go
```diff
@@ -67,16 +87,44 @@ func ValueFromUint256(value *uint256.Int) (result Value) {
...
-	result = value.Bytes32()
+	valueBytes := value.Bytes32()
+	for i := 0; i < 4; i++ {
+		copy(result[i*8:], valueBytes[i*8:])
+	}
 	return result
 }
 
-func Add(a, b Value) Value {
-	return ValueFromUint256(new(uint256.Int).Add(a.ToUint256(), b.ToUint256()))
```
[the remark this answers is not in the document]
facuMH: I originally disliked casting so much to `uint256`. So I also added a benchmark, and this is the result.
```
             │ oldAdd.bench │              new.bench              │
             │    sec/op    │   sec/op     vs base                │
Value_Add-16   12.130n ± 1%   7.454n ± 1%  -38.55% (p=0.000 n=25)

             │ oldAdd.bench │           new.bench            │
             │     B/op     │    B/op     vs base            │
Value_Add-16     0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=25) ¹
¹ all samples are equal

             │ oldAdd.bench │           new.bench            │
             │  allocs/op   │ allocs/op   vs base            │
Value_Add-16     0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=25) ¹
¹ all samples are equal
```
No memory allocation either way, but the new implementation is 38% faster. 
Sadly this function is not called in many places yet. 
That said, I would still keep the new implementation if it is ok with all reviewers

go/tosca/types.go
```diff
@@ -49,6 +51,12 @@ func (v Value) ToUint256() *uint256.Int {
 	return new(uint256.Int).SetBytes(v[:])
 }
 
+func (v Value) getInternalUint64(index int) uint64 {
```
HerbertJordan, line 54: nit: please move utility functions to the bottom

go/tosca/types_test.go
HerbertJordan, line 1: nit: There should be a test that `NewValue()` produces zero as advertised

go/tosca/types_test.go
```diff
@@ -87,15 +87,61 @@ func TestAddress_JSON_InvalidValueDecodingFails(t *testing.T) {
...
+
+			if test.value[test.index] != 1 {
+				t.Errorf("NewValue failed to set the correct value.")
+			}
+
+		})
+	}
+}
+
+func TestValue_Get(t *testing.T) {
```
HerbertJordan, line 113: Please improve the name of this test. There is no function `Get` and even if, a test should be named by the property it is testing, like `TestValue_StringProducesDecimalPrint` below.

go/tosca/types_test.go
```diff
@@ -87,15 +87,61 @@ func TestAddress_JSON_InvalidValueDecodingFails(t *testing.T) {
...
+	}
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("%v[%d]", test.value, test.index), func(t *testing.T) {
+			if got := test.value.getInternalUint64(test.index); test.want != got {
+				t.Errorf("wanted %v but got %v.", test.want, got)
+			}
+		})
+	}
+}
```
HerbertJordan, line 133: The two new tests are checking that the provides `uint64` values are converted to the 32 bytes in the type using big endian order, but I think there should be a test that also makes sure that for the arithmetic operations this order is used. Could be something simpler like
```
Sub(NewValue(1,0),NewValue(1)) == NewValue(math.MaxUint64)
Sub(NewValue(1,0,0),NewValue(1)) == NewValue(math.MaxUint64,math.MaxUint64)
Sub(NewValue(1,0,0,0),NewValue(1)) == NewValue(math.MaxUint64,math.MaxUint64,math.MaxUint64)
```
and the same for `Add`.

go/tosca/types.go
```diff
@@ -67,16 +87,41 @@ func ValueFromUint256(value *uint256.Int) (result Value) {
...
 	}
-	result = value.Bytes32()
-	return result
+	return value.Bytes32()
 }
 
-func Add(a, b Value) Value {
-	return ValueFromUint256(new(uint256.Int).Add(a.ToUint256(), b.ToUint256()))
+func Add(a, b Value) (z Value) {
+	// return ValueFromUint256(new(uint256.Int).Add(a.ToUint256(), b.ToUint256()))
```
HerbertJordan, line 94: If the new version is accepted, please remove the comment.

HerbertJordan (approved):

go/tosca/types_test.go
facuMH: in `TestValue_StringProducesDecimalPrint` there is a case of `{NewValue(), "0"},`

go/tosca/types_test.go
facuMH: tests added. nice suggestion

HerbertJordan (approved):

simonlechner (approved):

PR #601, opened 2024-07-09
Floria: Registration and simple value transfer

### LINKED ISSUE #568: Floria: Support value transfer call
Define a testcase in the integration end to end testing with a simple value transfer and make Floria (and the reference implementation) pass this test.

simonlechner:
This PR registers Floria (Tosca's own processor implementation) to the processors to test. Functionality for a simple value transfer is also added. 
Fixes #568.

### DIFF
--- a/go/processor/floria/processor.go
+++ b/go/processor/floria/processor.go
@@ -1,0 +1,31 @@
+package floria
+
+import (
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+)
+
+func init() {
+	tosca.RegisterProcessorFactory("floria", newProcessor)
+}
+
+func newProcessor(interpreter tosca.Interpreter) tosca.Processor {
+	if interpreter == nil {
+		panic("interpreter is required")
+	}
+
+	return &processor{
+		interpreter: interpreter,
+	}
+}
+
+type processor struct {
+	interpreter tosca.Interpreter
+}
+
+func (p *processor) Run(
+	blockParams tosca.BlockParameters,
+	transaction tosca.Transaction,
+	context tosca.TransactionContext,
+) (tosca.Receipt, error) {
+	return tosca.Receipt{}, nil
+}
--- a/go/processor/floria/processor.go
+++ b/go/processor/floria/processor.go
@@ -1,6 +1,18 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
 package floria
 
 import (
+	"fmt"
+
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 )
 
@@ -9,10 +21,6 @@ func init() {
 }
 
 func newProcessor(interpreter tosca.Interpreter) tosca.Processor {
-	if interpreter == nil {
-		panic("interpreter is required")
-	}
-
 	return &processor{
 		interpreter: interpreter,
 	}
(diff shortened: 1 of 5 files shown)


go/processor/floria/processor.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/holiman/uint256"
+)
+
+func init() {
+	tosca.RegisterProcessorFactory("floria", newProcessor)
+}
+
+func newProcessor(interpreter tosca.Interpreter) tosca.Processor {
+	if interpreter == nil {
+		panic("interpreter is required")
```
facuMH, line 26: mmmm I think that if you run the program and see it crashing and the panic is just "interpreter is requires" seems a little out of context. Perhaps make a more self contained panic message, such as "interpreter is required to start a new processor" or something else. wdyt ?

go/processor/floria/processor.go
```diff
@@ -0,0 +1,115 @@
...
+
+func buyGas(transaction tosca.Transaction, context tosca.TransactionContext) error {
+	gasPrice := transaction.GasPrice.ToUint256()
+	gas := uint256.NewInt(uint64(transaction.GasLimit))
+	gas = gas.Mul(gas, gasPrice)
+
+	// Buy gas
+	senderBalance := context.GetBalance(transaction.Sender).ToUint256()
+	if senderBalance.Cmp(gas) < 0 {
+		return fmt.Errorf("insufficient balance: %v < %v", senderBalance, gas)
```
facuMH, line 91: should we already setup an `errors.go` this could be a simple call to `errInsufficientBalance(senderBalance,gas)`

go/processor/floria/processor.go
```diff
@@ -0,0 +1,115 @@
...
+	if senderBalance.Cmp(transferValue) < 0 {
+		return fmt.Errorf("insufficient balance: %v < %v", senderBalance, transferValue)
+	}
+
+	senderBalance = senderBalance.Sub(senderBalance, transferValue)
+	context.SetBalance(transaction.Sender, tosca.ValueFromUint256(senderBalance))
+
+	receiverBalance := context.GetBalance(*transaction.Recipient).ToUint256()
+	receiverBalance = receiverBalance.Add(receiverBalance, transferValue)
+	context.SetBalance(*transaction.Recipient, tosca.ValueFromUint256(receiverBalance))
```
facuMH, line 112:
```suggestion
	senderBalance := context.GetBalance(transaction.Sender)
	transferValue := transaction.Value
	if senderBalance.Cmp(transferValue) < 0 {
		return fmt.Errorf("insufficient balance: %v < %v", senderBalance, transferValue)
	}

	senderBalance = senderBalance.Sub(senderBalance, transferValue)
	context.SetBalance(transaction.Sender, senderBalance)

	receiverBalance := context.GetBalance(*transaction.Recipient)
	receiverBalance = receiverBalance.Add(receiverBalance, transferValue)
	context.SetBalance(*transaction.Recipient, receiverBalance)
```
Maybe I'm missing something, but why are all the conversions to/from `U256` needed?

go/processor/floria/processor_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	context := test_processor.NewScenarioContext()
+	context.SetBalance(transaction.Sender, tosca.ValueFromUint64(1000))
+
+	err := buyGas(transaction, context)
+	if err != nil {
+		t.Errorf("buyGas returned an error: %v", err)
+	}
+	if context.GetBalance(transaction.Sender).Cmp(tosca.ValueFromUint64(1000-200*2)) != 0 {
```
facuMH, line 86: maybe you could rewrite this as `balance-gaslimit*gasprice` ?

facuMH (commented):
looks like a great first step !! just a couple of minor comments

go/processor/floria/processor_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func TestProcessor_HandleNonce(t *testing.T) {
+	transaction := tosca.Transaction{
+		Sender: tosca.Address{1},
+		Nonce:  9,
+	}
+
+	context := test_processor.NewScenarioContext()
```
HerbertJordan, line 38: I would argue that at the scope of unit testing, mocks for the `TransactionContext` may be more adequate. They provide a more fine-grained control on the expected effects and simplify the testing of the post-condition.
```
        ctrl := gomock.NewController(t)
	context := tosca.NewMockTransactionContext(ctrl)

	context.EXPECT().GetNonce(tosca.Address{1}).Return(uint64(9))
	context.EXPECT().SetNonce(tosca.Address{1}, uint64(10))

	transaction := tosca.Transaction{
		Sender: tosca.Address{1},
		Nonce:  9,
	}

	err := handleNonce(transaction, context)
	if err != nil {
		t.Errorf("handleNonce returned an error: %v", err)
        }
```

HerbertJordan (changes requested):
Looks good. Just added a few remarks on style decisions you might want to consider since this PR may implicitly initialize the guidelines for the Floria package.

go/processor/floria/processor.go
simonlechner: I didn't consider that tosca.value also implements these functions. I also changed it in the other functions.

go/processor/floria/processor.go
```diff
@@ -0,0 +1,115 @@
...
+	// Increment nonce
+	context.SetNonce(tosca.Address(transaction.Sender), context.GetNonce(tosca.Address(transaction.Sender))+1)
+
+	return nil
+}
+
+func buyGas(transaction tosca.Transaction, context tosca.TransactionContext) error {
+	gasPrice := transaction.GasPrice.ToUint256()
+	gas := uint256.NewInt(uint64(transaction.GasLimit))
+	gas = gas.Mul(gas, gasPrice)
```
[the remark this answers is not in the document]
simonlechner: I think using the Value type will be more readable in the long run since it will get rid of all these casts.

go/processor/floria/processor.go
simonlechner: As discussed in the standup, these errors are not going to be returned and are only used internal for this processor implementation.

go/processor/floria/processor_test.go
simonlechner: Right, changed it to mocks for all unit tests.

go/tosca/types.go
```diff
@@ -79,6 +79,10 @@ func Sub(a, b Value) Value {
 	return ValueFromUint256(new(uint256.Int).Sub(a.ToUint256(), b.ToUint256()))
 }
 
+func Mul(a, b Value) Value {
```
HerbertJordan, line 82: Something to be considered, but no need to address this in this PR: technically, Value is a number with a unit. The unit is "wei", and when multiplying two values with a unit, you would get the product of the units, in this case "wei^2", which is obviously a strange construct.

In reality, we probably never have any use case where we multiply two values with each other. What we actually do is scale a value by a unit-less scalar (e.g., GasPrice * Gas is a Value * scalar).

You may have noticed this problem because you have been forced to convert one of the parameters in the gas price computation to a value. This is a consequence of this. A way out, that at the first sight may look odd though, would be to define
```
func Mul(a Value, s uint64) Value { ... }
```
or
```
func (v Value) Scale(s uint64) Value { ... }
```
By defining it like this, it makes it easier to do the right operations (scaling a value) and harder to do something likely to be wrong (like multiplying values).

HerbertJordan (approved):

go/tosca/types.go
simonlechner: Thank you for the remark. I think this is a good change because, as you mentioned, it makes it harder to do wrong things in the future. I changed it to the `Scale` function.

facuMH (approved):

PR #604, opened 2024-07-11
tosca: unify opcodes

### LINKED ISSUE #599: Tosca Opcode
opcodes are defined and used in multiple projects, this leads to unnecessary conversions.
To fix these a `tosca.opcode` should be defined and the different projects should use this.

facuMH:
This PR unifies the opcodes used by all tosca projects. 
This PR fixes #599

### DIFF
--- a/go/integration_test/processor/world_state_test.go
+++ b/go/integration_test/processor/world_state_test.go
@@ -18,10 +18,6 @@ import (
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
-
-	// This is only imported to get the EVM opcode definitions.
-	// TODO: write up our own op-code definition and remove this dependency.
-	op "github.com/ethereum/go-ethereum/core/vm"
 )
 
 func TestWorldState_Equal(t *testing.T) {
@@ -264,7 +260,7 @@ func TestAccount_Diff(t *testing.T) {
 		},
 		"different_code": {
 			a:        Account{Code: tosca.Code{byte(vm.STOP)}},
-			b:        Account{Code: tosca.Code{byte(vm.ADD), byte(op.MUL)}},
+			b:        Account{Code: tosca.Code{byte(vm.ADD), byte(vm.MUL)}},
 			expected: []string{"different code: 0x00 != 0x0102"},
 		},
 		"different_storage": {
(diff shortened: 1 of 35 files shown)


go/tosca/vm/opcodes_test.go
```diff
@@ -81,28 +79,12 @@ func TestOpCode_NumberOfOpCodes(t *testing.T) {
...
-	case tosca.R11_Paris:
-		currentOpCodes = append(currentOpCodes, []OpCode{}...) // Paris
-		fallthrough
-	case tosca.R10_London:
-		currentOpCodes = append(currentOpCodes, []OpCode{BASEFEE}...) // London
-		fallthrough
-	case tosca.R09_Berlin:
-		currentOpCodes = append(currentOpCodes, []OpCode{}...) // Berlin
-	}
+	currentOpCodes = append(currentOpCodes, []OpCode{}...)                                            // Berlin
```
LuisPH3, line 83: As discussed, I am not sure about the use case of this unit test. But I would leave it as it was.

LuisPH3 (dismissed):

go/tosca/vm/opcodes_test.go
facuMH: I do not have strong opinions on the matter. I also don't think leaving a `switch` with all `fallthrought` that will pass, does not make much sense either.

HerbertJordan (dismissed):
Thanks for cleaning this up!

simonlechner (approved):
Looks good

go/tosca/vm/opcodes_test.go
simonlechner: This unit test was added before the new revision updates were applied. I think the switch case can and should be removed now that we have full support.

go/integration_test/processor/world_state_test.go
```diff
@@ -17,6 +17,7 @@ import (
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
 
 	// This is only imported to get the EVM opcode definitions.
 	// TODO: write up our own op-code definition and remove this dependency.
```
simonlechner, line 23: Since this import is no longer required it should be removed as well as the comment. Please use your IDE search function to find all other files where this is applicable.

simonlechner (changes requested):

go/integration_test/processor/world_state_test.go
facuMH: nice catch !!

simonlechner (approved):

LuisPH3 (approved):

HerbertJordan (approved):

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #606, opened 2024-07-12
Floria: Support for non recursive interpreter call

### LINKED ISSUE #602: Floria: support non-recursive calls to interpreter
Tosca's processor implementation should be able to call the underlying processor and handle its result.

simonlechner:
This PR adds support for a single call to the processor's interpreter. Since the value transfer is performed per call, it is moved into the call function. Gas is not yet handled or tested, this will be added in a follow up PR. Only simple code exampled have been tested for this first version, `context` access in the interpreter requires extensive testing and is therefore not part of this PR.
Fixes #602.


go/processor/floria/calls.go
```diff
@@ -0,0 +1,77 @@
...
+
+	result, err := interpreter.Run(params)
+	if err != nil || !result.Success {
+		context.RestoreSnapshot(snapshot)
+	}
+
+	return result, err
+}
+
+func transferValue(transaction tosca.Transaction, context tosca.TransactionContext) error {
```
facuMH, line 62: if `transferValue` will be needed in multiple call functions, will they all happen here or should it be in a more neutral file where it can be reutilized? maybe this is a premature optimization.

go/processor/floria/calls_test.go
```diff
@@ -0,0 +1,182 @@
...
+	transaction := tosca.Transaction{
+		Sender:    tosca.Address{1},
+		Recipient: &tosca.Address{2},
+	}
+
+	ctrl := gomock.NewController(t)
+	context := tosca.NewMockTransactionContext(ctrl)
+	interpreter := tosca.NewMockInterpreter(ctrl)
+
+	senderBalance := tosca.NewValue(0)
```
facuMH, line 56: shouldn't this be a non zero value ?

go/processor/floria/calls_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+import (
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"go.uber.org/mock/gomock"
+)
+
+func TestCalls_InterpreterResultIsHandledCorrectly(t *testing.T) {
+	tests := map[string]struct {
```
facuMH, line 21: no test for error case ?

go/processor/floria/processor.go
```diff
@@ -76,7 +90,7 @@ func handleNonce(transaction tosca.Transaction, context tosca.TransactionContext
 }
 
 func buyGas(transaction tosca.Transaction, context tosca.TransactionContext) error {
-	gas := transaction.GasPrice.Scale(uint64(transaction.GasLimit))
+	gas := transaction.GasPrice.Scale((uint64(transaction.GasLimit)))
```
facuMH, line 93: are the extra parenthesis necessary ?

facuMH (approved):
looks good, but I have a couple of comments

go/processor/floria/calls.go
simonlechner: I think all of the call implementations should be in this file. Maybe we are even able to merge the different calls into a single function to reduce code duplication. The `transferValue` function should not be required out of this file.

go/processor/floria/calls_test.go
simonlechner: These tests to not stress the value transfer and gas has already been purchased when the `call` function is executed, therefore this value should not have an impact on this test. This is explicitly shown by setting the value to 0.

go/processor/floria/calls_test.go
simonlechner: There is a case for a failed interpreter execution in the test but it is only checking the success of the result. I think the output should not be considered in a failed case therefore I don't want to require a fixed value for it.

go/processor/floria/calls.go
```diff
@@ -0,0 +1,77 @@
...
+
+package floria
+
+import (
+	"fmt"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+)
+
+func Call(interpreter tosca.Interpreter, transaction tosca.Transaction, context tosca.TransactionContext) (tosca.Result, error) {
```
HerbertJordan, line 19: nit: this function could be package private

go/processor/floria/calls.go
```diff
@@ -0,0 +1,77 @@
...
+		Input:     transaction.Input,
+		Value:     transaction.Value,
+		CodeHash:  &codeHash,
+		Code:      code,
+	}
+
+	snapshot := context.CreateSnapshot()
+	if err := transferValue(transaction, context); err != nil {
+		context.RestoreSnapshot(snapshot)
+		return tosca.Result{}, err
```
HerbertJordan, line 51: Something to be covered in a follow-up: we need to check whether in this case the full Gas price should be charged.

HerbertJordan, line 51: nit: Not sure if a failed transaction should result in a cal error, since it might be considered normal operation.

go/processor/floria/calls_test.go
```diff
@@ -0,0 +1,182 @@
...
+				t.Errorf("Unexpected success value from interpreter call")
+			}
+			if string(result.Output) != string(test.output) {
+				t.Errorf("Unexpected output value from interpreter call")
+			}
+		})
+	}
+}
+
+func TestProcessor_TransferValue(t *testing.T) {
```
HerbertJordan, line 86:
```suggestion
func TestTransferValue_SuccessfulValueTransfer(t *testing.T) {
```

go/processor/floria/calls_test.go
```diff
@@ -0,0 +1,182 @@
...
+	err := transferValue(transaction, context)
+	if err != nil {
+		t.Errorf("transferValue returned an error: %v", err)
+	}
+	if context.GetBalance(transaction.Sender).Cmp(tosca.NewValue(90)) != 0 {
+		t.Errorf("Sender balance was not decremented")
+	}
+	if context.GetBalance(*transaction.Recipient).Cmp(tosca.NewValue(10)) != 0 {
+		t.Errorf("Recipient balance was not incremented")
+	}
```
HerbertJordan, line 112: These checks only check whether you set the expectations above right for those calls. They do not test anything in the `transfereValue` function, and can be removed.

go/processor/floria/calls_test.go
```diff
@@ -0,0 +1,182 @@
...
+	_, err := Call(interpreter, transaction, context)
+	if err != nil {
+		t.Errorf("transferValue returned an error: %v", err)
+	}
+	if context.GetBalance(transaction.Sender).Cmp(tosca.NewValue(90)) != 0 {
+		t.Errorf("Sender balance was not decremented")
+	}
+	if context.GetBalance(*transaction.Recipient).Cmp(tosca.NewValue(10)) != 0 {
+		t.Errorf("Recipient balance was not incremented")
+	}
```
HerbertJordan, line 148: Same here.

go/processor/floria/calls_test.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func TestProcessor_TransferValueInCall(t *testing.T) {
```
HerbertJordan, line 115:
```suggestion
func TestCall_TransferValueInCall(t *testing.T) {
```

go/processor/floria/calls_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	context.EXPECT().GetBalance(transaction.Sender).Return(tosca.NewValue(0))
+	if context.GetBalance(transaction.Sender).Cmp(tosca.NewValue(0)) != 0 {
+		t.Errorf("Sender balance was changed")
+	}
+
+	context.EXPECT().GetBalance(*transaction.Recipient).Return(tosca.NewValue(0))
+	if context.GetBalance(*transaction.Recipient).Cmp(tosca.NewValue(0)) != 0 {
+		t.Errorf("Recipient balance was changed")
+	}
```
HerbertJordan, line 181: Also here. Here you are only testing the mock implementation.

HerbertJordan (changes requested):
The design looks good, just a few concerns about the tests. Please have a look.

go/processor/floria/calls.go
simonlechner: In [geth](https://github.com/ethereum/go-ethereum/blob/cf0378499f1bcae65c093c58cd6ca8225e91b125/core/vm/evm.go#L203), the call returns with a insufficient balance error if there is not enough balance.

simonlechner: In [geth](https://github.com/ethereum/go-ethereum/blob/cf0378499f1bcae65c093c58cd6ca8225e91b125/core/vm/evm.go#L204), the unchanged gas in returned if there is not enough balance. Checks for this will be added once the gas computation and billing is implemented.

go/processor/floria/calls_test.go
simonlechner: I already thought so, but wanted to test more than just the returned error. I removed them and the ones in the other tests.

go/processor/floria/calls.go
HerbertJordan: We have been derivating from the `geth` implementation in the past as well if it lead to cleaner interface definitions -- in particular regarding error handling. As we discussed in the past, signaling the result of perfectly acceptable operations through errors is not good practice.

When we say we have to be `geth` compatible and use `geth` as a reference, this is limited to the observable behavior. We are free to change internal aspects as needed.

The decision how to handle insufficient funds for a value transfer will be more clear once this code is reused for nested calls. When covering this, a consistent setup between the [Processor](https://github.com/Fantom-foundation/Tosca/blob/31731009a10dfe16344549cafbd26d25c0763d02/go/tosca/processor.go#L22), [Interpreter](https://github.com/Fantom-foundation/Tosca/blob/31731009a10dfe16344549cafbd26d25c0763d02/go/tosca/interpreter.go#L32), and [RunContext](https://github.com/Fantom-foundation/Tosca/blob/31731009a10dfe16344549cafbd26d25c0763d02/go/tosca/interpreter.go#L77) interfaces should be found and documented.

HerbertJordan (approved):

facuMH (approved):

PR #607, opened 2024-07-12
Add offending corpus file.

LuisPH3:
This file was found during fuzzing tests and it triggers a bug yet to be solved.


facuMH:
in the last week a few bug fixes were merged. could you run this corpuses against the updated main ?

LuisPH3:
These copus files had a very large gas value. New checks will skip them.

PR #609, opened 2024-07-12
Add Go Vet pass to build action.

LuisPH3:
Vet is bundled with the go installation, no need for extra setup. This pass will run fast static checks for code sanity.

More about [Vet](https://pkg.go.dev/cmd/vet)

Vet is fast and can be run along the normal build script. Although it founds some trivial errors already, there is a second linter project that finds more interesting errors, and I would like to integrate it as an action as well: #608 

The bug found is a race condition: more about it here https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/
The choice of naming for the new variable is deliberate, this seems to be a convention to address this issue.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -28,7 +28,7 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Thanks for this addition!

PR #455, opened 2024-05-06
CT: Add CALLCODE

facuMH:
This PR adds the instruction CALLCODE to our specification

This PR fixes #381 
This PR depends on #450

PR #612, opened 2024-07-12
Handle ignored error results discovered by static linter

LuisPH3:
The PR here either ignores or does error handling for the previously ignored functions, potentially returning errors.

- Rand (and hash), implement the read (and write) interfaces which are error typed, nevertheless these operations do not fail.
- Error handling is done in unit tests and benchmarks, benchmarks add a negligible constant overhead, and it is better to not benchmark failing functions.

### DIFF
--- a/go/ct/spc/enumeration_test.go
+++ b/go/ct/spc/enumeration_test.go
@@ -239,7 +239,7 @@ func TestEnumeration_RightNumberOfGoroutinesIsStarted(t *testing.T) {
 	printFunction := func(time time.Duration, rate float64, current int64) {}
 	err := ForEachState(Spec.GetRules(), opFunction, printFunction, numJobs, uint64(seed), fullMode)
 	if err != nil {
-		t.Errorf("Unexpected in ForEachState %v", err)
+		t.Errorf("Unexpected error in ForEachState %v", err)
 	}
 
 	if activeJobs != numJobs {
--- a/go/ct/spc/enumeration_test.go
+++ b/go/ct/spc/enumeration_test.go
@@ -237,7 +237,10 @@ func TestEnumeration_RightNumberOfGoroutinesIsStarted(t *testing.T) {
 	}
 
 	printFunction := func(time time.Duration, rate float64, current int64) {}
-	ForEachState(Spec.GetRules(), opFunction, printFunction, numJobs, uint64(seed), fullMode)
+	err := ForEachState(Spec.GetRules(), opFunction, printFunction, numJobs, uint64(seed), fullMode)
+	if err != nil {
+		t.Errorf("Unexpected in ForEachState %v", err)
+	}
 
 	if activeJobs != numJobs {
 		t.Errorf("unexpected number of active jobs, wanted %d, got %d", numJobs, activeJobs)
--- a/go/ct/spc/specification_test.go
+++ b/go/ct/spc/specification_test.go
@@ -76,7 +76,7 @@ func TestSpecification_EachRuleProducesAMatchingTestCase(t *testing.T) {
 			rnd := rand.New(0)
 			hits := 0
 			misses := 0
-			rule.EnumerateTestCases(rnd, func(state *st.State) rlz.ConsumerResult {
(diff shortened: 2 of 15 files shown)


go/ct/spc/enumeration_test.go
```diff
@@ -237,7 +237,10 @@ func TestEnumeration_RightNumberOfGoroutinesIsStarted(t *testing.T) {
 	}
 
 	printFunction := func(time time.Duration, rate float64, current int64) {}
-	ForEachState(Spec.GetRules(), opFunction, printFunction, numJobs, uint64(seed), fullMode)
+	err := ForEachState(Spec.GetRules(), opFunction, printFunction, numJobs, uint64(seed), fullMode)
+	if err != nil {
+		t.Errorf("Unexpected in ForEachState %v", err)
```
HerbertJordan, line 242:
```suggestion
		t.Errorf("Unexpected error in ForEachState: %v", err)
```

go/ct/st/accounts.go
```diff
@@ -32,7 +32,7 @@ func NewAccounts() *Accounts {
 
 func (a *Accounts) GetCodeHash(address tosca.Address) (hash [32]byte) {
 	hasher := sha3.NewLegacyKeccak256()
-	hasher.Write(a.code[address].ToBytes())
+	_, _ = hasher.Write(a.code[address].ToBytes())
```
HerbertJordan, line 35: Like for the random number generator, we should comment that the hasher is never producing an error (assuming this is true):
```suggestion
	_, _ = hasher.Write(a.code[address].ToBytes()) // hasher never produces an error
```

HerbertJordan (approved):

go/ct/spc/specification_test.go
```diff
@@ -240,7 +244,7 @@ func TestSpecification_AtMostOneCodeAtPC(t *testing.T) {
 }
 
 // TODO: re-enable this test when runtime is not an issue anymore.
-// note: before commiting run it locally where timeout is not an issue.
+// note: before committing run it locally where timeout is not an issue.
```
facuMH, line 247: has an issue been created to do this ? if so please write the number

facuMH (approved):
looks good. thanks for the work !! but I'm wondering if it would be cleaner/better practice to have wrapper functions for all these unused error returns ?

simonlechner (approved):

go/ct/spc/specification_test.go
LuisPH3: This is a lint of a misspelled comment.

go/ct/st/accounts.go
LuisPH3: yes, 
from the Hash interface
```go
type Hash interface {
	// Write (via the embedded io.Writer interface) adds more data to the running hash.
	// It never returns an error.
	io.Writer
```
the Read method is a little more complicated, we declare an interface in our own namespace 
```go
// keccakState wraps sha3.state. In addition to the usual hash methods, it also supports
// Read to get a variable amount of data from the hash state. Read is faster than Sum
// because it doesn't copy the internal state, but also modifies the internal state.
type keccakState interface {
	hash.Hash
	Read([]byte) (int, error)
}
```
This interface "discovers" a method Read implemented in crypto@xxx/sha3, the body never returns an error.

go/ct/spc/specification_test.go
LuisPH3: @facuMH you introduced this comment, do you have a follow-up ticket? https://github.com/Fantom-foundation/Tosca/pull/455

facuMH: in my local machine it takes 22.9 seconds. I think this test can be re-enabled and the comment deleted.

facuMH (approved):

go/ct/spc/specification_test.go
LuisPH3: [#615](https://github.com/Fantom-foundation/Tosca/issues/615)

simonlechner (approved):

PR #613, opened 2024-07-12
CT/LFVM: Memory/Gas overflow checks.

### LINKED ISSUE #524: LFVM: Memory expansion overflow checks

facuMH:
Thanks to the fuzzer continously finding more overflow checks, we digged dipper into the issue and found a few missing overflow checks. As well as defining a max gas that would allow for a big enough memory that would trigger the overflows.

This PR includes:
- move `SizeInWords` function to `tosca/utils` package so that it can be used by both ct and lfvm
- rename of `st.MaxGas` to `st.MaxGasUsedByCt`
- limit use `st.MaxGasUsedByCt` as upper bound in `gasDomain`.
- add `st.MaxMemoryExpansionSize` to memory parameters. also remove non interesting parameters to reduce number of generated cases.
- add specification test, testing rules sensitive to this overflow
- add usages of `tosca.SizeInWords` and `sumWithOverflow` to specification in vulnerable effects.
- add usages of `tosca.SizeInWords` in lfvm instructions where applicable.
- add test to `lfvm/instructions_tests.go` to check sensitive rules for these overflow issue.
- add gas overflow check to call operations. 

This PR fixes #524


go/ct/st/state.go
```diff
@@ -24,7 +24,8 @@ import (
 // Upper bound for gas, this limit is required since evmc defines a signed type for gas.
 // Limiting gas also solves issue 293 regarding out of memory failures,
 // discussed here: https://github.com/Fantom-foundation/Tosca/issues/293
-const MaxGas = tosca.Gas(1 << 60)
+// This should be enough gas to allow for ~450MB.
```
HerbertJordan, line 27: It would be good to document how we got to this number for future readers of this code.

go/tosca/utils.go
```diff
@@ -64,3 +66,10 @@ func GetStorageStatus(original, current, new Word) StorageStatus {
 	// Default
 	return StorageAssigned
 }
+
+func SizeInWords(size uint64) uint64 {
```
HerbertJordan, line 70: Please add documentation for this function, as it is now at such a prominent place.

HerbertJordan (changes requested):
I see lots of changes in this PR, but no unit or regression tests.

How would we make sure that after those issues have now been fixed we will not re-introduce those over time?

go/ct/spc/specification_test.go
```diff
@@ -373,3 +377,83 @@ func TestSumWithOverflow(t *testing.T) {
...
+				opToRules[instruction] = append(opToRules[instruction], rule)
+			}
+		}
+	}
+
+	rnd := rand.New(0)
+
+	for op, opRules := range opToRules {
+		for _, rule := range opRules {
+			for _, index := range indexOfMemParam[op] {
```
HerbertJordan, line 424: nit: `index` => `memoryParameterPosition` or something similar

go/ct/spc/specification_test.go
```diff
@@ -373,3 +377,83 @@ func TestSumWithOverflow(t *testing.T) {
...
+						}
+
+						state, err := generator.Generate(rnd)
+						if err != nil {
+							t.Fatalf("failed to generate a random state: %v", err)
+						}
+
+						rule.Effect.Apply(state)
+
+						if state.Status != st.Failed && state.Gas != 0 {
```
HerbertJordan, line 451: I think this condition is wrong ... we expect all those rules to lead to a failing state. And then we do not care about the gas.

The way it is formulated here, we would accept if a rule is transitioning to a running state as long as the gas is 0.

go/ct/spc/specification_test.go
```diff
@@ -373,3 +377,83 @@ func TestSumWithOverflow(t *testing.T) {
 		})
 	}
 }
+
+func TestSpecification_OpsWithDynamicCostHandleOverflowSizes(t *testing.T) {
```
HerbertJordan, line 381: I like the general structure of this test, but I am not too confident about the code filtering out rules to be tested. I do not think there is a need for this. We can just test all rules, as it is done here:
```

func TestSpecification_OpsWithDynamicCostHandleOverflowSizes(t *testing.T) {

	indexOfMemParam := map[vm.OpCode][]int{
		vm.MCOPY:          {0, 1, 2},
		vm.CODECOPY:       {0, 2},
		vm.CALLDATACOPY:   {0, 2},
		vm.RETURNDATACOPY: {0, 1, 2},
		vm.REVERT:         {0, 1},
		vm.CREATE:         {1, 2},
		vm.CREATE2:        {1, 2},
		vm.EXTCODECOPY:    {1, 3},
		vm.CALLCODE:       {3, 4, 5, 6},
		vm.SHA3:           {0, 1},
		vm.STATICCALL:     {2, 3, 4, 5},
	}
	testValues := []uint64{st.MaxMemoryExpansionSize, st.MaxMemoryExpansionSize + 1, math.MaxUint64}

	rnd := rand.New(0)

	allRules := getAllRules()
	for op := range indexOfMemParam {
		for revision := common.MinRevision; revision <= common.NewestSupportedRevision; revision++ {
			for _, overflowPosition := range indexOfMemParam[op] {
				for _, value := range testValues {
					t.Run(fmt.Sprintf("%v_%v_%v_%v", op, revision, overflowPosition, value), func(t *testing.T) {

						condition := rlz.And(
							rlz.IsRevision(revision),
							rlz.Eq(rlz.Status(), st.Running),
							rlz.Eq(rlz.Op(rlz.Pc()), vm.OpCode(op)),
							rlz.Eq(rlz.Param(overflowPosition), common.NewU256(value)),
						)

						for _, i := range indexOfMemParam[op] {
							if i != overflowPosition {
								condition = rlz.And(condition, rlz.Eq(rlz.Param(i), common.NewU256(1)))
							}
						}

						generator := gen.NewStateGenerator()
						condition.Restrict(generator)

						state, err := generator.Generate(rnd)
						if err != nil {
							t.Fatalf("failed to generate a random state: %v", err)
						}

						for _, rule := range allRules {
							match, err := rule.Condition.Check(state)
							if err != nil {
								t.Fatalf("failed to check rule condition for %v: %v", rule.Name, err)
							}
							if match {
								result := state.Clone()
								rule.Effect.Apply(result)
								if want, got := st.Failed, result.Status; want != got {
									t.Fatalf("Rule %s does not fail with overflow parameters, got %v", rule.Name, got)
								}
								result.Release()
							}
						}
					})
				}
			}
		}
	}
}
```
Wdyt?

go/ct/st/state.go
```diff
@@ -25,7 +25,9 @@ import (
...
+// This value comes from inverting the closure calcMemoryCost.
```
HerbertJordan, line 29:
```suggestion
// The costs for buying memory in an EVM instance is defined by
//                              C(W) = W^2/512 + 3*W
// where W is the number of 32-byte words to be allocated. A common
// issue encountered in implementations is that the W^2 term is not
// guarded against overflows, causing resulting costs to be far to low.
// This, in turn, would allow operations with overflowing memory requirements
// to trigger memory expansions for low gas costs that would exceed real
// world memory requirements.
// Such overflows happen whenever more than MaxMemoryExapnsionSize
// words of memory are to be allocated. To be able to detect those in the CT,
// there must be sufficient Gas provided to cover the 3*W term in the gas cost
// equation, assuming that the W^2/512 will overflow to a near zero value.
// Thus,
//   3 * 0x1FFFFFFFE0 =  3 * 137_438_953_440 =  412_316_860_320
// which we rounded up to 500*10^9 as defined below. This amount of gas allows
// for the allocation of up to ~450MB of memory in a correct implementation, which
// is far beyond any real-world budget but acceptable for a test environment.
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -792,3 +793,85 @@ func TestTransientStorageOperations(t *testing.T) {
...
+		maxMemoryExpansionSize,
+		maxMemoryExpansionSize + 1,
+		math.MaxUint64,
+	}
+
+	tests := map[string]struct {
+		op         func(*context)
+		stackSize  int
+		memIndexes []int
+		revision   tosca.Revision
```
HerbertJordan, line 808: nit: revision is always the same value; no need to define it in the test case spec;

HerbertJordan (changes requested):

go/ct/spc/specification_test.go
facuMH: I didn't take this approach originally because I wanted to reduce the number of `rule.Condition.Check`, but I agree it looks cleaner this way.

go/ct/st/state.go
```diff
@@ -25,7 +25,28 @@ import (
...
 // Limiting gas also solves issue 293 regarding out of memory failures,
 // discussed here: https://github.com/Fantom-foundation/Tosca/issues/293
-const MaxGas = tosca.Gas(1 << 60)
+// The costs for buying memory in an EVM instance is defined by
+//
+//	C(W) = W^2/512 + 3*W
+//
+// where W is the number of 32-byte words to be allocated. A common
+// issue encountered in implementations is that the W^2 term is not
+// guarded against overflows, causing resulting costs to be far to low.
```
simonlechner, line 34:
```suggestion
// guarded against overflows, causing resulting costs to be far too low.
```

go/ct/spc/specification_test.go
```diff
@@ -371,3 +374,71 @@ func TestSumWithOverflow(t *testing.T) {
...
+						for _, rule := range allRules {
+							match, err := rule.Condition.Check(state)
+							if err != nil {
+								t.Fatalf("failed to check rule condition for %v: %v", rule.Name, err)
+							}
+							if match {
+								result := state.Clone()
+								rule.Effect.Apply(result)
+								if result.Status != st.Failed {
+									t.Fatalf("Rule %s fails with overflow parameters", rule.Name)
```
simonlechner, line 434: Since you are checking that the status is not equal st.Failed the error message should say that it did NOT fail with overflow parameters. Or am I misunderstanding something?

simonlechner (approved):
Just some minor remarks, nice that it has been found and fixed.

go/ct/st/state.go
```diff
@@ -25,7 +25,28 @@ import (
...
+// The costs for buying memory in an EVM instance is defined by
+//
+//	C(W) = W^2/512 + 3*W
+//
+// where W is the number of 32-byte words to be allocated. A common
+// issue encountered in implementations is that the W^2 term is not
+// guarded against overflows, causing resulting costs to be far too low.
+// This, in turn, would allow operations with overflowing memory requirements
+// to trigger memory expansions for low gas costs that would exceed real
+// world memory requirements.
```
HerbertJordan, line 37:
```suggestion
// world memory capacities.
```

HerbertJordan (dismissed):

LuisPH3 (approved):

simonlechner (approved):

PR #614, opened 2024-07-15
Handle errors discovered by static linter, staticcheck and ineffassign

LuisPH3:
The goal is still to enable #608

Three kinds:
- staticcheck detects range checks that are equal to type range. This is superfluous as the compiler already did that 
- deprecated calls, these are bigger than I. Are they really deprecated?  The code and the uses seem quite critical to me.

### DIFF
--- a/go/ct/gen/interval_solver_test.go
+++ b/go/ct/gen/interval_solver_test.go
@@ -279,7 +279,7 @@ func TestInterval_isEmpty(t *testing.T) {
 	}
 }
 
-func TestIntervalSolver_uint64fullrangeAndEdges(t *testing.T) {
+func TestIntervalSolver_uint64FullRangeAndEdges(t *testing.T) {
 
 	tests := map[string]struct {
 		setup func(*IntervalSolver[uint64])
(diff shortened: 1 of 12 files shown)


go/ct/gen/interval_solver_test.go
```diff
@@ -285,15 +285,12 @@ func TestIntervalSolver_uint64fullrangeAndEdges(t *testing.T) {
 		setup func(*IntervalSolver[uint64])
 		check func(uint64) bool
 	}{
-		"full-range": {
```
facuMH, line 288: I think there is still value in checking that the default is the whole range. Do you think we can test this in a different manner instead of just deleting the test ?

facuMH (dismissed):
LGTM

go/ct/gen/interval_solver_test.go
```diff
@@ -285,15 +285,12 @@ func TestIntervalSolver_uint64fullrangeAndEdges(t *testing.T) {
...
-			setup: func(s *IntervalSolver[uint64]) {},
```
simonlechner, line 289: Although the `check` does not add any value, running the solver without any constraints is still useful to ensure no error is produced.

simonlechner (commented):
Looks good, just one remark.

go/ct/gen/interval_solver_test.go
LuisPH3: You cannot check if it is not in rage, since the test returns uint64 and the value is uint64, any out of range would overflow and therefore be in range. 
Nevertheless, testing that this code does not return an err

go/ct/gen/interval_solver_test.go
LuisPH3: right.

go/ct/utils/adapter_test.go
```diff
@@ -163,7 +163,7 @@ func TestAdapter_ParameterConversion(t *testing.T) {
 			},
 			func(p tosca.Parameters) (any, any) {
 				ctxt := p.Context
-				return tosca.Word{}, ctxt.GetCommittedStorage(tosca.Address{}, tosca.Key{})
+				return tosca.Word{}, ctxt.GetCommittedStorage(tosca.Address{}, tosca.Key{}) //nolint:staticcheck
```
HerbertJordan, line 166: It would be nice to comment what issues should be ignored here, that apparently can not be fixed.

HerbertJordan (changes requested):
I do not like those seemingly unmotivated `nolint:staticcheck` comments. What are the issues there and why can't they be fixed?

go/ct/utils/adapter_test.go
LuisPH3: So far, all ignores are deprecated calls. Fixing these deprecations is quite some work. 
Let me annotate a fix up issue.

LuisPH3: - added issue #616, so that we can enable lint checks in PR asap, fixing these deprecations is delicate work.
- commented ignored lint lines

LuisPH3:
I changed the linter from golangci-lint to staticcheck, these are different projects:

- staticcheck does not report ineffectual assignment, I do prefer the code initializing variables to meaningful values. 
- It has a more expressive ignore comment, fine grain, and with reason comment.

facuMH (approved):
descriptive comments detailing the issue and a number of for the issue to fix it are much appreciated !! thanks

PR #618, opened 2024-07-16
CT: fix makefile geth coverage command

facuMH:
This pr adds the `github.com/ethereum/go-ethereum/core/vm/...` packages to the coverage report of geth. This is the correction of an oversight. 
The resulting report accounts for coverage of all packages in both `./go/...` as well as `github.com/ethereum/go-ethereum/core/vm/...`, which would be created in an analogous place as before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,9 @@ evmone:
 	cmake --build build --parallel -t evmone
 
 coverage-go: DATE=$(shell date +"%Y-%m-%d-%T")
-coverage-go: TOSCA_GO_COVERAGE_DIR=./go/build/coverage/${TOSCA_GO_COVERAGE_EVM}/${DATE}/
+coverage-go: TOSCA_GO_COVERAGE_DIR=./go/build/coverage/${TOSCA_GO_COVERAGE_EVM}/${DATE}
 coverage-go:
-	go build -cover -o ${TOSCA_GO_COVERAGE_DIR}/driver ./go/ct/driver/ ; \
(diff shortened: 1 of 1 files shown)


HerbertJordan (approved):

simonlechner (approved):

Makefile
```diff
@@ -49,6 +49,7 @@ ct-coverage-lfvm: TOSCA_GO_COVERAGE_EVM=lfvm
 ct-coverage-lfvm: coverage-go
 
 ct-coverage-geth: TOSCA_GO_COVERAGE_EVM=geth
+ct-coverage-geth: TOSCA_GO_COVERAGE_DEPENDENCY_PACKAGES=github.com/ethereum/go-ethereum/core/vm/...
```
LuisPH3, line 52: Shall we add more packages here? 
I think we use vm, core, common, params, and types. 

I am not sure if adding github.com/ethereum/go-ethereum would be a good idea, as we may no use the most of it.

LuisPH3 (changes requested):
Can you have a look at the geth packages?

Makefile
LuisPH3: As discussed, other packages are not relevant. important stuff is in vm.

LuisPH3 (approved):

Makefile
facuMH: I think that all the relevant parts for us, are just in core/vm, and this should be sufficient. 
As discussed offline, the gas expantion cost and instructions are both in vm package, so this is fine.

PR #620, opened 2024-07-17
LFVM: unify calls

facuMH:
Since the code for all 4 different kinds if call operations is mostly the same, whenever a bug is found in one, it most likely also applies for all of them, it makes it more maintainable to have a single implementation with the few parts that are specific for each kind guarded. 

This PR does exactly that, unifies the `opCall`, `opCallCode`, `opStaticCall` and `opDelegateCall`. To verify that these changes do not modify the expected behaviour, all operations ran against the CT and Geth.

This PR depends on #613

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -1050,14 +1050,26 @@ func neededMemorySize(c *context, offset, size *uint256.Int) (uint64, error) {
 	return offset.Uint64() + size.Uint64(), nil
 }
 
-func opCall(c *context) {
+func genericCall(c *context, kind tosca.CallKind) {
 	warmAccess, coldCost, err := addressInAccessList(c)
 	if err != nil {
 		return
 	}
 	stack := c.stack
+	value := uint256.NewInt(0)
+
 	// Pop call parameters.
-	provided_gas, addr, value, inOffset, inSize, retOffset, retSize := stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop()
+	provided_gas, addr := stack.pop(), stack.pop()
+	if kind == tosca.Call || kind == tosca.CallCode {
+		value = stack.pop()
+	}
+	inOffset, inSize, retOffset, retSize := stack.pop(), stack.pop(), stack.pop(), stack.pop()
+
+	// We need to check the existence of the target account before removing
+	// the gas price for the other cost factors to make sure that the read
+	// in the state DB is always happening. This is the current EVM behavior,
+	// and not doing it would be identified by the replay tool as an error.
+	toAddr := tosca.Address(addr.Bytes20())
 
 	// Compute and charge gas price for call
 	arg_memory_size, err := neededMemorySize(c, inOffset, inSize)
@@ -1068,48 +1080,52 @@ func opCall(c *context) {
 	if err != nil {
 		return
 	}
+
 	needed_memory_size := arg_memory_size
 	if ret_memory_size > arg_memory_size {
 		needed_memory_size = ret_memory_size
 	}
-	base_gas := c.memory.ExpansionCosts(needed_memory_size)
-	if base_gas < 0 || c.gas < base_gas {
+
+	baseGas := c.memory.ExpansionCosts(needed_memory_size)
+	checkGas := func(cost tosca.Gas) bool {
+		return 0 <= cost && cost <= c.gas
+	}
+	if !checkGas(baseGas) {
 		c.status = OUT_OF_GAS
 		return
 	}
 
-	// We need to check the existence of the target account before removing
-	// the gas price for the other cost factors to make sure that the read
-	// in the state DB is always happening. This is the current EVM behavior,
-	// and not doing it would be identified by the replay tool as an error.
-	toAddr := tosca.Address(addr.Bytes20())
-
+	// for static and delegate calls, the following value checks will always be zero.
 	// Charge for transferring value to a new address
 	if !value.IsZero() {
-		base_gas += tosca.Gas(params.CallValueTransferGas)
+		baseGas += tosca.Gas(params.CallValueTransferGas)
+	}
+	if !checkGas(baseGas) {
+		c.status = OUT_OF_GAS
+		return
 	}
 
-	// if evm.chainRules.IsEIP158 according to GETH it is EIP158 since 2016
-	// !!!! but need to touch stateDB for the address to have it in the substate record key/value
-	if !value.IsZero() && !c.context.AccountExists(toAddr) {
-		base_gas += tosca.Gas(params.CallNewAccountGas)
+	// EIP158 states that non-zero value calls that create a new account should
+	// be charged an additional gas fee.
+	if kind == tosca.Call && !value.IsZero() && !c.context.AccountExists(toAddr) {
+		baseGas += tosca.Gas(params.CallNewAccountGas)
+	}
+	if !checkGas(baseGas) {
+		c.status = OUT_OF_GAS
+		return
 	}
 
-	cost := callGas(c.gas, base_gas, provided_gas)
-
-	if warmAccess {
-		if !c.UseGas(base_gas + cost) {
-			return
-		}
-	} else {
+	cost := callGas(c.gas, baseGas, provided_gas)
+	if !warmAccess {
 		// In case of a cold access, we temporarily add the cold charge back, and also
 		// add it to the returned gas. By adding it to the return, it will be charged
 		// outside of this function, as part of the dynamic gas, and that will make it
 		// also become correctly reported to tracers.
 		c.gas += coldCost
-		if !c.UseGas(base_gas + cost + coldCost) {
-			return
-		}
+		baseGas += coldCost
+	}
+	if !c.UseGas(baseGas + cost) {
+		return
 	}
 
 	// first use static and dynamic gas cost and then resize the memory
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/instructions.go
```diff
@@ -1119,44 +1127,64 @@ func opCall(c *context) {
...
-		balance := c.context.GetBalance(c.params.Recipient)
-		balanceU256 := new(uint256.Int).SetBytes32(balance[:])
-		if balanceU256.Lt(value) {
-			c.stack.pushEmpty().Clear()
-			c.return_data = nil
-			c.gas += cost // the gas send to the nested contract is returned
-			return
+	callParams := tosca.CallParameters{}
+	subKind := kind
+	switch kind {
```
LuisPH3, line 1148: nit:
- 'c.params.Static' check can be done in the opCall function, no need for subkind in this function.
- Input and gas are common to all cases, do once. 
- Value can be common, and set to c.params.Value for DelegateCall

This reduces decision-making in the switch, and it is easier to read, sender, recipient and code address.

go/interpreter/lfvm/instructions.go
```diff
@@ -1175,264 +1203,20 @@ func opCall(c *context) {
...
-		Gas:         cost,
-	})
-
-	if err == nil {
-		if memSetErr := c.memory.Set(retOffset.Uint64(), retSize.Uint64(), ret.Output); memSetErr != nil {
-			c.SignalError(memSetErr)
-		}
-	}
+func opCall(c *context) {
+	genericCall(c, tosca.Call)
```
LuisPH3, line 1207: nit: as mentioned before, static check here and then call with StaticCall if needed.

go/interpreter/lfvm/instructions.go
```diff
@@ -1050,14 +1050,19 @@ func neededMemorySize(c *context, offset, size *uint256.Int) (uint64, error) {
...
 }
 
-func opCall(c *context) {
+func genericCall(c *context, kind tosca.CallKind) {
 	warmAccess, coldCost, err := addressInAccessList(c)
 	if err != nil {
 		return
 	}
 	stack := c.stack
+	value := uint256.NewInt(0)
```
LuisPH3, line 1059:
```suggestion
	value := uint256.NewInt(0)
	
```

go/interpreter/lfvm/instructions.go
```diff
@@ -1119,44 +1127,64 @@ func opCall(c *context) {
 		cost += tosca.Gas(params.CallStipend)
 	}
 
-	// If we are in static mode, recursive calls are to be treated like
-	// static calls. This is a consequence of the unification of the
```
LuisPH3, line 1123: Iff you move the static check somewhere else, please take this comment there as well.

go/interpreter/lfvm/instructions.go
```diff
@@ -1084,19 +1091,20 @@ func opCall(c *context) {
 	// and not doing it would be identified by the replay tool as an error.
 	toAddr := tosca.Address(addr.Bytes20())
```
LuisPH3, line 1092: Can you move these two lines somewhere else? The current code section is computing gas costs, and then this conversion is declared in between.  

IMO this belongs between stack pops & gas cost or after gas costs

go/interpreter/lfvm/instructions.go
```diff
@@ -1119,44 +1127,64 @@ func opCall(c *context) {
...
-	// If we are in static mode, recursive calls are to be treated like
-	// static calls. This is a consequence of the unification of the
-	// interpreter interfaces of EVMC and Geth.
-	// This problem was encountered in block 58413779, transaction 7.
-	kind := tosca.Call
-	if c.params.Static {
-		kind = tosca.StaticCall
-
-		// In a static call, no value must be transferred.
+	if kind == tosca.Call || kind == tosca.CallCode {
```
LuisPH3, line 1130: Because static calls must have value 0. Could we drop this check and keep `if !value.IsZero() {` only?

go/interpreter/lfvm/instructions.go
```diff
@@ -1084,19 +1091,20 @@ func opCall(c *context) {
...
+
 	// Charge for transferring value to a new address
 	if !value.IsZero() {
 		base_gas += tosca.Gas(params.CallValueTransferGas)
 	}
 
 	// if evm.chainRules.IsEIP158 according to GETH it is EIP158 since 2016
 	// !!!! but need to touch stateDB for the address to have it in the substate record key/value
-	if !value.IsZero() && !c.context.AccountExists(toAddr) {
+	if !value.IsZero() && !c.context.AccountExists(toAddr) && kind == tosca.Call {
```
LuisPH3, line 1103:
```suggestion
	if kind == tosca.Call && !value.IsZero() && !c.context.AccountExists(toAddr)  {
```

Avoid triggering AccountExists with lazy evaluation easier to read as well. more generic -> more specific

go/interpreter/lfvm/instructions.go
```diff
@@ -1084,19 +1091,20 @@ func opCall(c *context) {
...
 		base_gas += tosca.Gas(params.CallNewAccountGas)
 	}
 
 	cost := callGas(c.gas, base_gas, provided_gas)
-
 	if warmAccess {
```
LuisPH3, line 1108: This whole if could be refactor: 

```
    if !warmAccess {
                // In case of a cold access, we temporarily add the cold charge back, and also
		// add it to the returned gas. By adding it to the return, it will be charged
		// outside of this function, as part of the dynamic gas, and that will make it
		// also become correctly reported to tracers.
		c.gas += coldCost
                base_gas += coldCost
     }

     if !c.UseGas(base_gas + cost) {
	return
     }
```

- base_gas is never used afterwards
- could you rename base_gas to baseGas?

LuisPH3 (changes requested):
It seems complete to me, nevertheless I see potential for a leaner function.
question: are Create and Create2 are too different to add to this refactor?

facuMH:
> It seems complete to me, nevertheless I see potential for a leaner function. question: are Create and Create2 are too different to add to this refactor?

I think they are. but if we (as a team) decide they should be integrated, it should also be a different PR

go/interpreter/lfvm/instructions.go
```diff
@@ -1119,44 +1127,64 @@ func opCall(c *context) {
...
+	case tosca.DelegateCall:
+		callParams.Sender = c.params.Sender
+		callParams.Recipient = c.params.Recipient
+		callParams.CodeAddress = toAddr
+		callParams.Input = args
+		callParams.Value = c.params.Value
+		callParams.Gas = cost
+	}
+
+	ret, err := c.context.Call(subKind, callParams)
```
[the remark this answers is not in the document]
facuMH: @simonlechner probably has an answer for this.

go/interpreter/lfvm/instructions.go
facuMH: this check is to filter out delegate calls, but I will join it with if right under.

go/interpreter/lfvm/instructions.go
```diff
@@ -1250,94 +1151,47 @@ func opCallCode(c *context) {
...
-	}
+	case tosca.CallCode:
+		callParams.Sender = c.params.Recipient
+		callParams.Recipient = c.params.Recipient
+		callParams.CodeAddress = toAddr
 
-	gas := callGas(c.gas, base_gas, provided_gas)
+	case tosca.StaticCall:
+		callParams.Sender = c.params.Recipient
+		callParams.Recipient = toAddr
```
simonlechner, line 1184: Call and StaticCall could be combined.

simonlechner (commented):
Nice way getting rid of duplicated code. Just some small remarks.

go/interpreter/lfvm/instructions.go
```diff
@@ -1193,42 +1080,56 @@ func opCallCode(c *context) {
...
-	if base_gas < 0 || c.gas < base_gas {
+
+	baseGas := c.memory.ExpansionCosts(needed_memory_size)
+	checkGas := func(cost tosca.Gas) bool {
+		if cost < 0 || c.gas < cost {
+			// c.status = OUT_OF_GAS
+			return false
+		}
+		return true
+	}
```
HerbertJordan, line 1096:
```suggestion
	checkGas := func(cost tosca.Gas) bool {
	        return 0 <= cost && cost <= c.gas
	}
```

go/interpreter/lfvm/instructions.go
```diff
@@ -1193,42 +1080,56 @@ func opCallCode(c *context) {
...
+		baseGas += tosca.Gas(params.CallValueTransferGas)
+	}
+	if !checkGas(baseGas) {
+		c.status = OUT_OF_GAS
+		return
 	}
 
-	cost := callGas(c.gas, base_gas, provided_gas)
+	// if evm.chainRules.IsEIP158 according to GETH it is EIP158 since 2016
+	// !!!! but need to touch stateDB for the address to have it in the substate record key/value
```
HerbertJordan, line 1113: nit: I'd say this comment needs improvement ...

HerbertJordan (approved):
Looks good, thanks!

go/interpreter/lfvm/instructions.go
facuMH: this comment is prior to this PR and was not modified, just relocated. That said, I also found it strange, so I rephrased it.

go/interpreter/lfvm/instructions.go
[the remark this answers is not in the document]
LuisPH3: ping @simonlechner

simonlechner (approved):

PR #622, opened 2024-07-17
Add support for collecting C++ coverage data from Go processes

LuisPH3:
When running a C++ library from go, the coverage instrumentation is never dump to file and is lost. This process needs to be done manually in these cross-language environments.

The solution proposed loads a C library in the Go program, which exports the functionality to dump coverage data into files. This dumps coverage data for any loaded library in the program namespace with coverage enabled.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -221,15 +221,15 @@ Ideally, pull requests targeting performance improvements should include such a 
 
 ## Code Coverage
 
-Code coverage of the Tosca project is a work in progress. As the different systems get integrated, more features will be added and documented here.
-Go has native support for coverage, which makes getting this metric for `lfvm` or `geth` driver runs quite simple.
+The Tosca project allows to collect coverage reports for unit tests and CT runs. 
+More about coverage in [the docs folder](docs/coverage.md)
 
 ### CT Driver Code coverage
 
 To run the the CT driver code coverage simply run `make ct-coverage-lfvm` or `make ct-coverage-geth`, these commands will create the folder `Tosca/go/build/coverage/` and a sub folder with the corresponding interpreter's name. In this directory an instrumented version of the driver will be built and then run. The reports of the coverage of this run will also be in this directory, where lastly an HTML version of the report will be added as well. As more interpreter implementations are added to the Tosca project, they should also be added to the makefile.
 
 
-### C++ Code coverage
+### C++ Code Unit Test Coverage
 
 Current code coverage infrastructure uses GNU gcov system. This is supported in both Gcc and Clang, although Clang target gcov version may not be the installed one. For this reason the current infrastructure is currently enabled for Gcc only. 
 
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ ct-coverage-geth: coverage-go
 
 ct-coverage-evmzero: tosca-cpp-coverage
 ct-coverage-evmzero: 
-	go run ./go/ct/driver run -f push evmzero ; \
+	go run ./go/ct/driver run evmzero ; \
 	echo "Coverage report generated in cpp/build/coverage/index.html"
 	@cd cpp/build ; \
 	cmake --build .  --target coverage 
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,10 @@ tosca-cpp:
 		-DTOSCA_ASAN="$(TOSCA_CPP_ASAN)"; \
 	cmake --build build --parallel
 
+tosca-cpp-coverage: TOSCA_CPP_BUILD = Debug
+tosca-cpp-coverage: TOSCA_CPP_COVERAGE = ON
+tosca-cpp-coverage: tosca-cpp
+
 evmone:
 	@cd third_party/evmone ; \
 	cmake -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_SHARED_LIBRARY_SUFFIX_CXX=.so ; \
@@ -52,6 +56,13 @@ ct-coverage-geth: TOSCA_GO_COVERAGE_EVM=geth
 ct-coverage-geth: TOSCA_GO_COVERAGE_DEPENDENCY_PACKAGES=github.com/ethereum/go-ethereum/core/vm/...
 ct-coverage-geth: coverage-go
 
+ct-coverage-evmzero: tosca-cpp-coverage
+ct-coverage-evmzero: 
+	go run ./go/ct/driver run -f push evmzero ; \
+	echo "Coverage report generated in cpp/build/coverage/index.html"
+	@cd cpp/build ; \
+	cmake --build .  --target coverage 
+
 test: test-go test-cpp
 
 test-go: tosca-go
@@ -64,6 +75,10 @@ test-cpp: tosca-cpp
 test-cpp-asan: TOSCA_CPP_BUILD = Debug
 test-cpp-asan: TOSCA_CPP_ASAN = ON
 test-cpp-asan: test-cpp
+
+cpp-coverage-report: 
+	@cd cpp/build ; \
+	cmake --build .  --target coverage 
 
 test-cpp-coverage: TOSCA_CPP_BUILD = Debug
 test-cpp-coverage: TOSCA_CPP_COVERAGE = ON
--- a/cpp/cmake/tosca_coverage.cmake
+++ b/cpp/cmake/tosca_coverage.cmake
@@ -14,10 +14,6 @@ if(TOSCA_COVERAGE)
   add_link_options(--coverage)
   add_definitions(-DTOSCA_COVERAGE=1)
 
-  find_program(LCOV lcov REQUIRED)
-  find_program(GCOVR gcovr REQUIRED)
-  find_program(GENHTML genhtml REQUIRED)
-
   add_custom_target(coverage
     COMMENT "Generating coverage report."
 
--- a/cpp/cmake/tosca_coverage.cmake
+++ b/cpp/cmake/tosca_coverage.cmake
@@ -1,36 +1,57 @@
 include_guard(GLOBAL)
 
-
 option(TOSCA_COVERAGE "Enable coverage report for evmzero." OFF)
 
 if(TOSCA_COVERAGE)
+ 
   if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     # gcc uses gcov, clang should use it as well, but it has version compatibility issues
     # between compiler generate code and the gcov tool.
-    message(FATAL_ERROR "Coverage is only supported with GCC.")
+    message(FATAL_ERROR "Coverage build currently is only supported with GCC.")
   endif()
 
-  add_compile_options($<$<CXX_COMPILER_ID:GNU>:--coverage>)
-  add_link_options($<$<CXX_COMPILER_ID:GNU>:--coverage>)
+  add_compile_options(--coverage)
+  add_link_options(--coverage)
+  add_definitions(-DTOSCA_COVERAGE=1)
 
   find_program(LCOV lcov REQUIRED)
+  find_program(GCOVR gcovr REQUIRED)
   find_program(GENHTML genhtml REQUIRED)
 
   add_custom_target(coverage
     COMMENT "Generating coverage report."
+
+    # Capture coverage data
     COMMAND ${LCOV} 
       --capture 
       --directory .  
       "$<$<VERSION_GREATER_EQUAL:${CMAKE_CXX_COMPILER_VERSION},13.2.0>:--ignore-errors>" 
       "$<$<VERSION_GREATER_EQUAL:${CMAKE_CXX_COMPILER_VERSION},13.2.0>:mismatch,mismatch,gcov>"
       --output-file coverage.info
+
+    # filter coverage data
     COMMAND ${LCOV} 
       --remove coverage.info 
       '/usr/include/*' 
       '/usr/lib/*' 
       '*/third_party/*'
       --output-file filtered.info
-    COMMAND ${GENHTML} filtered.info --output-directory coverage
+
+    # build and compress HTML report
+    COMMAND ${GENHTML} filtered.info --output-directory coverage --parallel
+    COMMAND tar czvf coverage.tar.gz coverage/ 
+
+    # Text report
+    COMMAND ${GCOVR} -r .. | tee coverage.txt
+
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
   )
+
+  # Add a target to clean coverage data (will print found data coverage report)
+  # executed on demand
+  add_custom_target(clean_coverage_data
+    COMMAND ${GCOVR} -r .. 
+    COMMAND ${CMAKE_COMMAND} -E remove -f coverage.txt coverage coverage.tar.gz
+  )
+
 endif()
--- a/cpp/cmake/tosca_coverage.cmake
+++ b/cpp/cmake/tosca_coverage.cmake
@@ -7,7 +7,7 @@ if(TOSCA_COVERAGE)
   if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     # Gcc uses Gcov, clang should use it as well, but it has version compatibility issues
     # between compiler generate code and the gcov tool. 
-    # Feel free to experiment Clang Gcov # support, unfortunatelly, it requires too much 
+    # Feel free to experiment with Clang Gcov # support, unfortunately, it requires too much 
     # manual tinkering to enable it in an automated way. 
     # Gcc and Gcov are released together, so they should be compatible.
     message(FATAL_ERROR "Coverage build only enabled for GCC.")
--- a/cpp/cmake/tosca_coverage.cmake
+++ b/cpp/cmake/tosca_coverage.cmake
@@ -5,9 +5,12 @@ option(TOSCA_COVERAGE "Enable coverage report for evmzero." OFF)
 if(TOSCA_COVERAGE)
  
   if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    # gcc uses gcov, clang should use it as well, but it has version compatibility issues
-    # between compiler generate code and the gcov tool.
-    message(FATAL_ERROR "Coverage build currently is only supported with GCC.")
+    # Gcc uses Gcov, clang should use it as well, but it has version compatibility issues
+    # between compiler generate code and the gcov tool. 
+    # Feel free to experiment Clang Gcov # support, unfortunatelly, it requires too much 
+    # manual tinkering to enable it in an automated way. 
+    # Gcc and Gcov are released together, so they should be compatible.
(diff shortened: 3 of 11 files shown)


facuMH (dismissed):

LuisPH3:
The more I think about this, the less I like it. 
As @facuMH already mentioned, modifying the SteppableEvmcInterpreter interface is not good.

This is a good interface, but the problem I see with it is that we use it as the return type of the factory method `NewConformanceTestingTarget`, after that we lose the chance to call the destruct function if needed. 

- [  ] I will see how can we de-interface the factories, and remove the Destroy method from SteppableEvmcInterpreter.

Do you guys have any other idea?

cpp/common/coverage/collect_coverage.c
```diff
@@ -0,0 +1,27 @@
+
+#if defined(TOSCA_COVERAGE)
+extern void __gcov_dump();
+#endif
+
+/// Reports whenever the library (and therefore the complete c++ project)
```
HerbertJordan, line 6: Just a question: what's with the three `///` here instead of the usual two? Does this have any particular meaning?

cpp/common/coverage/collect_coverage.c
```diff
@@ -0,0 +1,27 @@
...
+  return 0;
+#endif
+}
+
+/// When using GCov, Shared libraries do not dump coverage data
+/// coverage dump is delegated to the exit of the main program.
+/// When runtime linking is used, the host program can trigger the
+/// coverage dump by calling this function.
+/// This function will trigger dump of all loaded libraries compiled
+/// with coverage flags.
```
HerbertJordan, line 21:
```suggestion
/// When using GCov, Shared libraries do collect coverage data,
/// but it is not automatically written into a file. Usually, GCC would
/// add a corresponding call to the end of a `main` function, but it
/// can not do this for a Go application. Thus, the dumping of
/// coverage data of C++ library code needs to be triggered explicitly
/// by calling this function at the end of an application.
/// Calling this function will dump coverage data for all loaded C++ libraries
/// compiled with coverage flags. If coverage data collection is disabled,
/// this function is a no-op.
```

go/ct/common/cpp_coverage.go
```diff
@@ -0,0 +1,23 @@
...
+#cgo LDFLAGS: -L${SRCDIR}/../../../cpp/build/common/coverage -ltosca_collect_coverage
+#cgo LDFLAGS: -Wl,-rpath,${SRCDIR}/../../../cpp/build/common/coverage
+
+int IsCoverageEnabled();
+void DumpCoverageData();
+*/
+import "C"
+
+// isCppCoverageEnabled returns true if C++ has been compiled with coverage enabled.
+// this assumes that every C++ library loaded at runtime has been compiled with coverage enabled.
```
HerbertJordan, line 13:
```suggestion
// This assumes that every C++ library loaded at runtime for which coverage data should be collected has been compiled with coverage enabled.
```

go/ct/common/cpp_coverage.go
```diff
@@ -0,0 +1,23 @@
...
+func isCppCoverageEnabled() bool {
+	return C.IsCoverageEnabled() != 0
+}
+
+// DumpCppCoverageData triggers the C++ code to dump coverage data.
+// Not calling this function will result in no coverage data being reported
+// for runtime loaded C and C++ libraries.
```
HerbertJordan, line 20:
```suggestion
// for runtime loaded C and C++ libraries. If coverage data collection is not enabled,
// this function is a no-op.
```

go/ct/common/cpp_coverage_test.go
```diff
@@ -0,0 +1,40 @@
+package common
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings" // Add this line to import the "strings" package
```
HerbertJordan, line 7:
```suggestion
	"strings"
```

go/ct/common/cpp_coverage_test.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+		found = strings.HasSuffix(s, ".gcda")
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("Failed to walk directory %s: %v", tempDir, err)
+	}
+
+	if !found {
+		t.Fatalf("Failed to find gcda files")
```
HerbertJordan, line 38:
```suggestion
		t.Fatalf("Failed to collect C++ code coverage data")
```

cpp/cmake/tosca_coverage.cmake
```diff
@@ -1,36 +1,60 @@
 include_guard(GLOBAL)
 
-
 option(TOSCA_COVERAGE "Enable coverage report for evmzero." OFF)
 
 if(TOSCA_COVERAGE)
-  if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+ 
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
HerbertJordan, line 7: The old version checking explicitly for the GNU compiler seems to be the more accurate condition for hat is reported by the error message below. So why would this condition be changed?

HerbertJordan (approved):

HerbertJordan (approved):
The description of this PR should be updated to match the new approach on how code coverage data is collected now.

cpp/common/coverage/collect_coverage.c
LuisPH3: It is one of the methods to document functions using doxygen. https://www.doxygen.nl/manual/docblocks.html
Do we use some particular notation for documentation comments?  I have noticed `// <` in some places.

cpp/cmake/tosca_coverage.cmake
LuisPH3: yes, this has been reverted.

cpp/common/coverage/collect_coverage.c
LuisPH3: this was mostly muscle memory, shall we use something else?

LuisPH3:
Sorry about changes, I had a second PR, that would have part of the responsibilities that ended up merged into this one. 
I just avoid creating a second PR and added those changes here. 

Next time I am in a redesign process, like it happen in this PR. I will close the original PR and start fresh.

docs/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+```bash
+lcov --remove coverage.info /usr/include/* /usr/lib/*' */third_party/* --output-file filtered.info
+```
+- Text report is generated can can be found in `cpp/build/coverage.txt`
+- Html report is generated and can be found in `cpp/build/coverage/index.html` or compressed as `cpp/build/coverage.tar.gz`
+
+
+## Collect C++ coverage when running CT evmzero 
+
+Gcov collects coverage data during runtime and only at the end of the process is saved into a file. This instrumentation is added to the main function of the C++ process, and since CT is Go, be added.
```
simonlechner, line 32:
```suggestion
Gcov collects coverage data during runtime and only at the end of the process it is saved into a file. This mechanism would be called automatically on the C++ main process, but has to be called manually when using evmzero from a Go process.
```

simonlechner (dismissed):
Nice solution with good documentation.

docs/coverage.md
LuisPH3: Thanks

BUILD.md
```diff
@@ -221,15 +221,15 @@ Ideally, pull requests targeting performance improvements should include such a
...
+The Tosca project allows to collect coverage reports for unit tests and CT runs. 
+More about coverage in [the docs folder](docs/coverage.md)
 
 ### CT Driver Code coverage
 
 To run the the CT driver code coverage simply run `make ct-coverage-lfvm` or `make ct-coverage-geth`, these commands will create the folder `Tosca/go/build/coverage/` and a sub folder with the corresponding interpreter's name. In this directory an instrumented version of the driver will be built and then run. The reports of the coverage of this run will also be in this directory, where lastly an HTML version of the report will be added as well. As more interpreter implementations are added to the Tosca project, they should also be added to the makefile.
 
 
-### C++ Code coverage
+### C++ Code Unit Test Coverage
```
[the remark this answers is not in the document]
LuisPH3: Long story short. The coverage thing is too complicated not to explain some details. I did not find the right place to write this down for future me or other people. 

The idea behind splitting this documentation is to differentiate between how to run things from some text explaining how does it work internally. I thought it was just too much text to keep it in the same place. 
future changes or improvements in the usage of the feature would lead to `doc/coverage.md` being extended, but the section in `build.md` would need to be rewritten. 

wdyt?

Makefile
```diff
@@ -52,6 +56,13 @@ ct-coverage-geth: TOSCA_GO_COVERAGE_EVM=geth
 ct-coverage-geth: TOSCA_GO_COVERAGE_DEPENDENCY_PACKAGES=github.com/ethereum/go-ethereum/core/vm/...
 ct-coverage-geth: coverage-go
 
+ct-coverage-evmzero: tosca-cpp-coverage
+ct-coverage-evmzero: 
+	go run ./go/ct/driver run -f push evmzero ; \
```
HerbertJordan, line 61: This seems to use a filter to limit the range of covered tests. I guess this is a development artifact, right?

cpp/cmake/tosca_coverage.cmake
```diff
@@ -1,36 +1,56 @@
...
 
 if(TOSCA_COVERAGE)
+ 
   if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    # gcc uses gcov, clang should use it as well, but it has version compatibility issues
-    # between compiler generate code and the gcov tool.
-    message(FATAL_ERROR "Coverage is only supported with GCC.")
+    # Gcc uses Gcov, clang should use it as well, but it has version compatibility issues
+    # between compiler generate code and the gcov tool. 
+    # Feel free to experiment Clang Gcov # support, unfortunatelly, it requires too much
```
HerbertJordan, line 10:
```suggestion
    # Feel free to experiment with Clang Gcov # support, unfortunately, it requires too much 
```

cpp/cmake/tosca_coverage.cmake
```diff
@@ -1,36 +1,56 @@
...
+
+    # Text report
+    COMMAND ${GCOVR} -r .. | tee coverage.txt
+
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
   )
+
+  # Add a target to clean coverage data (will print found data coverage report)
+  # executed on demand
+  add_custom_target(clean_coverage_data
```
HerbertJordan, line 51: It does not seem like this is included in the `make clean` command. Is it covered implicitly?

cpp/common/coverage/collect_coverage.c
HerbertJordan: Seeing a `///` instead of a `//` makes me think that there is some special meaning for it, which, based on your feedback, there is not. So it should be rewritten to use `//`.

About the `// <` I use some times: I do this for fields, if I do not want to break the formatting.

When you have a struct like
```
type myData struct {
   short                 T
   longFieldName S
}
```
Go-format is nicely aligning the types. However, when adding documentation between those fields, this is lost:
```
type myData struct {
   // short is ...
   short T
   // longFieldName is ...
   longFieldName S
}
```
and it quickly looks messy. Thus, I use end-of-line comments like this:
```
type myData struct {
   short                 T  // < this is ..
   longFieldName S  // < this is ...
}
```
and the formatter makes it look good. Here, the `<` is to point to the thing it is documenting.

I saw this supported once in some doc generation tool, but I never checked what Go would make of it. I see those comments to be consumed mainly in-code.

cpp/common/coverage/collect_coverage.c
```diff
@@ -0,0 +1,30 @@
...
+/// was compiled with coverage flags.
+int IsCoverageEnabled() {
+#if defined(TOSCA_COVERAGE)
+  return 1;
+#else
+  return 0;
+#endif
+}
+
+/// When using Gcov, Shared libraries do collect coverage data,
```
HerbertJordan, line 16:
```suggestion
/// When using Gcov, shared libraries do collect coverage data,
```

docs/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+
+The official [documentation](https://go.dev/doc/build-cover) explains how to use it.
+Both Unit tests and CT driver can produce coverage reports.
+
+By default Go coverage will only include the source of included files, not dependencies. 
+The argument `-coverpkg=./go/...,github.com/project/dependency-name/` will include coverage reports from included code.
+
+## Coverage in C++
+
+C and C++ can be instrumented for coverage reports using both Gcc and Clang, nevertheless procedure is slightly different:
```
HerbertJordan, line 15:
```suggestion
C and C++ can be instrumented for coverage reports using both Gcc and Clang. However, the actual procedure to do so is slightly different:
```

docs/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+- Gcc uses [gcov](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html), there is an ecosystem of tools around gcov, to filter, modify and produce reports: [lcov](https://wiki.documentfoundation.org/Development/Lcov)
```
HerbertJordan, line 16:
```suggestion
- Gcc uses [gcov](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html); there is an ecosystem of tools around gcov, to filter, modify and produce reports: [lcov](https://wiki.documentfoundation.org/Development/Lcov)
```

docs/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+- [Clang](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) implements its [own coverage technology](https://clang.llvm.org/docs/SanitizerCoverage.html), but in theory they provide support for gcov. This later follows a different paradigm, as the coverage engine will only record data for the module indicated at runtime. 
+
+Tosca uses *gcov* to collect coverage data from the C++ tests.
+At the time this document was written, Clang compilation support for Gcov would require some extra work at the compilation level.
+
+By default all available source files collect coverage data. Third party and system headers coverage is filtered out after collection using lcov.
+```bash
+lcov --remove coverage.info /usr/include/* /usr/lib/*' */third_party/* --output-file filtered.info
+```
+- Text report is generated can can be found in `cpp/build/coverage.txt`
```
HerbertJordan, line 26:
```suggestion
- Text report is generated and can be found in `cpp/build/coverage.txt`
```

docs/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+- Html report is generated and can be found in `cpp/build/coverage/index.html` or compressed as `cpp/build/coverage.tar.gz`
+
+
+## Collect C++ coverage when running CT evmzero 
+
+Gcov collects coverage data during runtime and only at the end of the process it is saved into a file. This mechanism would be called automatically on the C++ main process, but has to be called manually when using evmzero from a Go process.
+It is possible to call the Gcov dump routine in runtime, by calling __gcov_dump. A mechanism to invoke such routine from Go has been implemented, so that the CT can generate the required data files.
```
HerbertJordan, line 33:
```suggestion
It is possible to call the Gcov dump routine in runtime, by calling __gcov_dump. A mechanism to invoke this routine from Go has been implemented, so that the CT can generate the required data files. 
```

docs/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+- Text report is generated can can be found in `cpp/build/coverage.txt`
+- Html report is generated and can be found in `cpp/build/coverage/index.html` or compressed as `cpp/build/coverage.tar.gz`
+
+
+## Collect C++ coverage when running CT evmzero 
+
+Gcov collects coverage data during runtime and only at the end of the process it is saved into a file. This mechanism would be called automatically on the C++ main process, but has to be called manually when using evmzero from a Go process.
+It is possible to call the Gcov dump routine in runtime, by calling __gcov_dump. A mechanism to invoke such routine from Go has been implemented, so that the CT can generate the required data files. 
+
+To retrieve the coverage report execute in the following order:
```
HerbertJordan, line 35:
```suggestion
To retrieve the coverage report execute the following commands:
```

docs/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+```bash
+make tosca-cpp-coverage
+go run ./go/ct/driver run evmzero
+make cpp-coverage-report
+```
+
+When replacing `make tosca-cpp-coverage` with `make cpp-test-coverage`, unit test coverage will be combined into the report.
+
+Results can be found as in the same folder as described before.
```
HerbertJordan, line 44:
```suggestion
Results can be found in the same folder as described before.
```

go/ct/common/cpp_coverage_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// check that at least one file is generated
+	found := false
+	err := filepath.WalkDir(tempDir, func(s string, _ fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		// .gcno are generated at compile time, with source code locations and other meta
+		// .gcda are generated at runtime with the actual coverage
+		found = strings.HasSuffix(s, ".gcda")
```
HerbertJordan, line 54: should this be changed to this?
```suggestion
		found = found || strings.HasSuffix(s, ".gcda")
```
Otherwise it would require the last file that is passed to have this extension.

go/ct/common/cpp_coverage_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	for _, arg := range args {
+		if foundDash {
+			afterDash = append(afterDash, arg)
+		} else if arg == "--" {
+			foundDash = true
+		}
+	}
+	return afterDash
+}
```
HerbertJordan, line 111: There is a standard library for parsing flags: [flag](https://pkg.go.dev/flag). Maybe this library should be used instead.

We have been using this in the past. An example can be seen [here](https://github.com/Fantom-foundation/Carmen/blob/0fcc59226f6a50509a0ee7d1b7fa04d3623cd23b/go/state/state_test.go#L913-L921).

HerbertJordan (changes requested):
Looks good, just a few minor suggestions on improving details.

Makefile
LuisPH3: Definitely an artifact. sorry about that

cpp/cmake/tosca_coverage.cmake
LuisPH3: Make clean deletes the build folder, coverage data will be gone.

Coverage data is incremental, any number of processes executions will add to the counters. This is good, because you can accumulate counters of unit tests.
Nevertheless, you may need a method to clean previous executions to measure coverage, but you may not want to compile the complete project again.

cpp/common/coverage/collect_coverage.c
LuisPH3: Well, there is a meaning: // is a comment, /// is documentation. Any tool processing documentation will ignore // comments. I see no use for this now, I'll remove them.

HerbertJordan (approved):

Issue #617, opened 2024-07-16
investigate ways to expand the code coverage to C++ code sections


LuisPH3:

LuisPH3:
My bad, I clicked in the wrong place and an issue was issued.

PR #623, opened 2024-07-17
Document and Script how to use C++ coverage in CT.

LuisPH3:
This PR is mostly documentation, should explain itself.

But there are some script changes
- enable c++ unit tests coverage in PR CI, because it is pretty fast.
- modify scripts to allow generating reports without compiling or running anything. Useful to compile-run-report arbitrary commands. 

fixes #623

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,13 @@ ct-coverage-geth: TOSCA_GO_COVERAGE_EVM=geth
 ct-coverage-geth: TOSCA_GO_COVERAGE_DEPENDENCY_PACKAGES=github.com/ethereum/go-ethereum/core/vm/...
 ct-coverage-geth: coverage-go
 
+ct-coverage-evmzero: tosca-cpp-coverage
+ct-coverage-evmzero: 
+	go run ./go/ct/driver run -f push evmzero ; \
+	echo "Coverage report generated in cpp/build/coverage/index.html"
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -32,6 +32,10 @@ tosca-cpp:
 		-DTOSCA_ASAN="$(TOSCA_CPP_ASAN)"; \
 	cmake --build build --parallel
 
+test-cpp-coverage: TOSCA_CPP_BUILD = Debug
```
facuMH, line 35: should this not also be `tosca-cpp-coverage` ?? 

also, should we try to keep it consistent with the other coverage commands, like `ct-coverage-lfvm`. wdyt?

facuMH (approved):
looks good. just a small naming suggestion

Makefile
LuisPH3: I don't think so:

- tosca-cpp: job compiles c++ code
- test-cpp: job compiles and runs c++ unit tests
- test-cpp-coverage: same but with coverage

I will introduce ct-coverage-evmzero in the next pr.

LuisPH3:
fixes  #617

simonlechner (approved):

HerbertJordan (commented):
Waiting for this to be synced with other changes.

LuisPH3:
contents in this file were added to #622

PR #624, opened 2024-07-17
Add gas specific integration test cases

simonlechner:
This PR introduces gas specific integration test cases. It extends the integration test scenario with an expected opera error, since the opera reference implementation handles insufficient gas with errors.

### DIFF
--- a/go/integration_test/processor/processor_gas_test.go
+++ b/go/integration_test/processor/processor_gas_test.go
@@ -12,6 +12,7 @@ package processor
 
 import (
 	"fmt"
+	"maps"
 	"strings"
 	"testing"
 
@@ -21,13 +22,11 @@ import (
 )
 
 func getGasTestScenarios() map[string]Scenario {
-	const (
-		codePrice = 3 + 3
-		excessGas = 100
-	)
+	// cost for 2 PUSH1 operations
+	const executionGasCost = 3 + 3
 
-	allTestCases := map[string]Scenario{
-		"ValueTransferExact": {
+	exactTestCases := map[string]Scenario{
+		"ValueTransfer": {
 			Before: WorldState{
 				{1}: Account{Balance: tosca.NewValue(100), Nonce: 4},
 			},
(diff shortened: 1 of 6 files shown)


go/integration_test/processor/scenario.go
```diff
@@ -29,15 +29,19 @@ type Scenario struct {
 	Parameters  tosca.BlockParameters
 	Transaction tosca.Transaction
 	Receipt     tosca.Receipt
+	OperaError  error
```
facuMH, line 32: this should an implementation agnostic name, maybe `ProcessorError` or `TxError`, but not `OperaError`, wdyt?

go/integration_test/processor/processor_gas_test.go
```diff
@@ -0,0 +1,503 @@
...
+				Sender:    tosca.Address{1},
+				Recipient: &tosca.Address{2},
+				GasLimit:  floria.TxGas + codePrice,
+				Nonce:     4,
+			},
+			After: WorldState{
+				{1}: Account{Balance: tosca.NewValue(100), Nonce: 5},
+				{2}: Account{Balance: tosca.NewValue(0),
+					Code: tosca.Code{
+						byte(op.PUSH1), byte(0), // < PUSH 0
```
facuMH, line 111: why not use `PUSH0` since it should be implemented in all interpreters ?

go/integration_test/processor/processor_gas_test.go
```diff
@@ -0,0 +1,503 @@
...
+						byte(op.RETURN),
+					},
+				},
+			},
+			Receipt: tosca.Receipt{
+				Success: false,
+				GasUsed: floria.TxGas + codePrice - 1,
+			},
+			OperaError: fmt.Errorf("gas too low"),
```
facuMH, line 185: this error is used multiple times, maybe it's worth predeclaring as a variable or const.

facuMH (approved):
Looks good, just a couple of style suggestions. Also, the gas limit  formulas are repeated (or patially repeated) for multiple test cases. do you think it would make sense (or possible) to give them significant variable names and reuse them throughout the tests ? this way each formula would only be declared once, less error prone. But I can also see this as a personal style preference, wdyt  ?

go/integration_test/processor/processor_gas_test.go
```diff
@@ -0,0 +1,503 @@
...
+				GasLimit:  floria.TxGas - 1,
+				Value:     tosca.NewValue(3),
+				Nonce:     4,
+			},
+			After: WorldState{
+				{1}: Account{Balance: tosca.NewValue(100), Nonce: 4},
+			},
+			Receipt: tosca.Receipt{
+				Success: false,
+				GasUsed: 0,
```
HerbertJordan, line 86: We once started discussing this but did not get to a conclusion; I would assume that in such a case the GasLimit might have to be returned, since this corresponds to the actual gas cost, right? After all, I would assume that the gas bought by the user at the begin of the transaction is not returned.

This would also make this case handled identically to a case where one is running out of gas during a contract execution.

go/integration_test/processor/processor_gas_test.go
```diff
@@ -0,0 +1,503 @@
...
+						byte(op.PUSH1), byte(0), // < PUSH 0
+						byte(op.PUSH1), byte(0), // < PUSH 0
+						byte(op.RETURN),
+					},
+				},
+			},
+			Transaction: tosca.Transaction{
+				Sender:    tosca.Address{1},
+				Recipient: &tosca.Address{2},
+				GasLimit:  floria.TxGas + codePrice,
```
HerbertJordan, line 104: nit: "codePrice" => "executionGas" or "executionGasCosts"

go/integration_test/processor/processor_gas_test.go
HerbertJordan: I would prefer the `PUSH1` version as it is here, since it would make those tests independent of the revision.

go/integration_test/processor/processor_gas_test.go
```diff
@@ -0,0 +1,503 @@
...
+			After: WorldState{
+				{1}: Account{Balance: tosca.NewValue(100), Nonce: 4},
+				{2}: Account{Balance: tosca.NewValue(0)},
+			},
+			Receipt: tosca.Receipt{
+				Success: false,
+				GasUsed: 0,
+			},
+			OperaError: fmt.Errorf("gas too low"),
+		},
```
HerbertJordan, line 250: Maybe a helper function producing test cases "<name>Exact", "<name>Sufficient", and "<name>Insufficient" for a given scenario would be a nice to have. Wdyt?

go/integration_test/processor/processor_gas_test.go
```diff
@@ -0,0 +1,503 @@
...
+				Nonce:     4,
+				Input:     []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+			},
+			After: WorldState{
+				{1}: Account{Balance: tosca.NewValue(100), Nonce: 4},
+				{2}: Account{Balance: tosca.NewValue(0)},
+			},
+			Receipt: tosca.Receipt{
+				Success: false,
+				GasUsed: 0,
```
HerbertJordan, line 311: As argued above, I find it surprising that the reported `GasUsed` here is 0 instead of `GasLimit`. We need to check this.

go/integration_test/processor/processor_gas_test.go
```diff
@@ -0,0 +1,503 @@
...
+
+	allTestCases := map[string]Scenario{
+		"ValueTransferExact": {
+			Before: WorldState{
+				{1}: Account{Balance: tosca.NewValue(100), Nonce: 4},
+			},
+			Transaction: tosca.Transaction{
+				Sender:    tosca.Address{1},
+				Recipient: &tosca.Address{2},
+				GasLimit:  floria.TxGas,
```
HerbertJordan, line 37: nit: it is a bit odd to use constants from a concrete implementation to define implementation agnostic tests. It might be nice to have those constants defined at some implementation-independent place, but I am not sure what a good place might be. So I guess we leave this for a potential future cleanup pass ...

HerbertJordan (changes requested):
We should re-check the reported `GasUsed` values for failed operations. This might simply also be wrong in the Opera implementation. Reporting values of 0 seem to be off.

go/integration_test/processor/processor_gas_test.go
facuMH: I like the idea of implementation agnostic tests, I would say that if we want these values to also be called in the implementations, then a nice place would be on the same level as the processor interface definition. But if we just want these values to be test constants, then they could be defined in `go/integration_tests/processor/gas_values.go` . wdyt ?

LuisPH3 (commented):
I looks good to me, please don't wait for my approval. The missing details may be a little too domain specific for my current understanding.

go/processor/floria/processor.go
```diff
@@ -16,6 +16,15 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 )
 
+const (
+	TxGas                     = 21_000
```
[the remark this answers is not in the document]
simonlechner: This has been taken from the reference implementation.

go/integration_test/processor/scenario.go
simonlechner: Our implementation will not return error for the tested cases, that's why I specifically named the field `OperaError`.

go/integration_test/processor/processor_gas_test.go
simonlechner: The code has been refactored to only use this error in one place.

go/integration_test/processor/processor_gas_test.go
simonlechner: I will keep this as a cleanup task once the implementation is fixed. I think all these constants should be in the same place eventually.

go/integration_test/processor/processor_gas_test.go
simonlechner: I also thought this is very surprising in the current reference implementation, but I thought it has to be correct because the whole history passed. I changed the reference implementation on this branch, it still has been confirmed to be compatible with the history.

go/integration_test/processor/scenario.go
```diff
@@ -29,15 +29,19 @@ type Scenario struct {
...
 }
 
 func (s *Scenario) Run(t *testing.T, processor tosca.Processor) {
 
 	context := newScenarioContext(s.Before)
 	receipt, err := processor.Run(s.Parameters, s.Transaction, context)
-	if err != nil {
+	if err != nil && s.OperaError == nil {
```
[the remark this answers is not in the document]
simonlechner: I don't want to change major parts of the reference implementation in this stage of the implementation, but I will keep this as a follow up task.

go/integration_test/processor/processor_gas_test.go
HerbertJordan: It might very likely be the case that in case of a failure the `GasUsed` field is ignored in Aida and the full gas consumption is just hard-coded.

It might also be that `receipts` are not checked at all in `aida-vm`.

HerbertJordan (approved):
Looks good, thanks!

PR #625, opened 2024-07-17
Floria: Implement gas billing

simonlechner:
This PR adds gas computations and billing to Floria. A number of integration test cases stressing the setup gas computation have been added in #624.

### DIFF
--- a/go/processor/floria/processor.go
+++ b/go/processor/floria/processor.go
@@ -99,7 +99,7 @@ func gasUsed(transaction tosca.Transaction, gasLeft tosca.Gas) tosca.Gas {
 		gasLeft -= gasLeft / 10
 	}
 
-	return transaction.GasLimit - tosca.Gas(gasLeft)
+	return transaction.GasLimit - gasLeft
 }
 
 func setupGasBilling(transaction tosca.Transaction) tosca.Gas {
--- a/go/processor/floria/processor.go
+++ b/go/processor/floria/processor.go
@@ -44,15 +44,21 @@ func (p *processor) Run(
 	transaction tosca.Transaction,
 	context tosca.TransactionContext,
 ) (tosca.Receipt, error) {
-
 	errorReceipt := tosca.Receipt{
 		Success: false,
 		GasUsed: transaction.GasLimit,
 	}
+	gas := transaction.GasLimit
 
 	if err := buyGas(transaction, context); err != nil {
 		return errorReceipt, nil
 	}
+
+	intrinsicGas := setupGasBilling(transaction)
+	if gas < intrinsicGas {
+		return errorReceipt, nil
+	}
+	gas -= intrinsicGas
 
 	if err := handleNonce(transaction, context); err != nil {
 		return errorReceipt, nil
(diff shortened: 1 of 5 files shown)


go/processor/floria/processor.go
```diff
@@ -70,31 +73,74 @@ func (p *processor) Run(
 		// Create new contract
 	} else {
 		// Call existing contract
-		result, err = call(p.interpreter, transaction, context)
+		result, err = call(p.interpreter, transaction, context, gas)
 		if err != nil {
-			return errorReceipt, nil
+			return tosca.Receipt{GasUsed: transaction.GasLimit}, nil
```
facuMH, line 78: why is this no longer an error ?

go/processor/floria/processor.go
```diff
@@ -70,31 +73,74 @@ func (p *processor) Run(
...
 		Logs:            nil,
 	}, nil
 }
 
+func gasUsed(transaction tosca.Transaction, gasLeft tosca.Gas) tosca.Gas {
+	if transaction.Sender != (tosca.Address{}) {
+		gasLeft -= gasLeft / 10
+	}
+
+	return transaction.GasLimit - tosca.Gas(gasLeft)
```
facuMH, line 98: could `GasLimit` ever be less than `gasLeft` ? should check for underflow ?

go/processor/floria/processor.go
```diff
@@ -70,31 +73,74 @@ func (p *processor) Run(
...
+}
+
+// No overflow check for the gas computation is required although it is performed in the
+// opera version. The overflow check would be triggered in a worst case with an input
+// greater than 2^64 / 16 - 53000 = ~10^18, which is not possible with real world hardware
```
facuMH, line 103: not possible because the code would have to be longer than 10^18 bytes of ram or why ?

facuMH: also, this code looks like a comment for the operations in line 121, not function documentation

go/processor/floria/processor_test.go
```diff
@@ -113,3 +114,99 @@ func TestProcessor_BuyGasInsufficientBalance(t *testing.T) {
...
+				t.Errorf("gasUsed returned incorrect result, got: %d, want: %d", actualGasUsed, test.expectedGasUsed)
+			}
+		})
+	}
+}
+
+func TestProcessor_SetupGasBilling(t *testing.T) {
+	tests := map[string]struct {
+		recipient       *tosca.Address
+		input           []byte
```
facuMH, line 157: dont't we have a `tosca.Code` for this types ?

facuMH (commented):
looks good, just a couple of questions.

go/processor/floria/processor.go
HerbertJordan, line 78: The potential `err` produced by the `call` is ignored. Should this be like this?
```suggestion
			return tosca.Receipt{GasUsed: transaction.GasLimit}, err
```

go/processor/floria/processor.go
```diff
@@ -70,31 +73,74 @@ func (p *processor) Run(
...
-		GasUsed:         transaction.GasLimit,
+		GasUsed:         gasUsed,
 		ContractAddress: nil,
 		Output:          result.Output,
 		Logs:            nil,
 	}, nil
 }
 
+func gasUsed(transaction tosca.Transaction, gasLeft tosca.Gas) tosca.Gas {
+	if transaction.Sender != (tosca.Address{}) {
```
HerbertJordan, line 94: This special case handling for the zero-sender is surprising and at least note-worthy. Why would this be done?

If we confirm this, we should have an integration making sure that the surplus-gas charge is indeed not charged when a transaction was send by Address zero.

go/processor/floria/processor.go
```diff
@@ -70,31 +73,74 @@ func (p *processor) Run(
...
+	}
+
+	return transaction.GasLimit - tosca.Gas(gasLeft)
+}
+
+// No overflow check for the gas computation is required although it is performed in the
+// opera version. The overflow check would be triggered in a worst case with an input
+// greater than 2^64 / 16 - 53000 = ~10^18, which is not possible with real world hardware
+func setupGasBilling(transaction tosca.Transaction) tosca.Gas {
+	var gas uint64
```
HerbertJordan, line 105: nit: should this use the `tosca.Gas` type?

HerbertJordan (approved):

go/processor/floria/processor.go
```diff
@@ -70,31 +73,74 @@ func (p *processor) Run(
 		// Create new contract
 	} else {
```
LuisPH3, line 74: no else, if error you returned already.

LuisPH3 (dismissed):

go/processor/floria/processor.go
simonlechner: This `else` is not from an error check but from an if checking whether it is a create (not yet implemented) or a normal call.

go/processor/floria/processor.go
simonlechner: Not in this spot, gas checks are performed before the execution.

go/processor/floria/processor_test.go
simonlechner: This is the input and not code, we handle it as a byte slice everywhere.

go/processor/floria/processor.go
LuisPH3: Question: 
If we return `value, err` and the caller of this function does the usual:
 ```go 
 v, err :=call()
 if err {
       // pontentially break flow
}
 ```

Aren't we dropping the value in that case? 
wouldn't be nil, err more solid? 

The way I see it, there are two options: 

- processor takes the decision, this error is handled as `transaction.GasLimit`, log error is needed, after which it can be dropped
- Processor honours the error, returning `nil, err`, and error handling is applied in the caller side. 

The proposed change will lead to code ignoring the `transaction.GasLimit` resolution and end up having a new resolution for the error applied.

go/processor/floria/processor.go
simonlechner: The comment in the reference implementation says "For whatever reason". I added a comment stating that this is indented behavior, as well as an integration test.

HerbertJordan: I remember writing the "whatever reason" comment :)
It was and still is a surprise.

go/processor/floria/processor.go
```diff
@@ -70,31 +76,74 @@ func (p *processor) Run(
...
 	}, nil
 }
 
+func gasUsed(transaction tosca.Transaction, gasLeft tosca.Gas) tosca.Gas {
+	// 10% of remaining gas is charged for non-internal transactions
+	if transaction.Sender != (tosca.Address{}) {
+		gasLeft -= gasLeft / 10
+	}
+
+	return transaction.GasLimit - tosca.Gas(gasLeft)
```
HerbertJordan, line 102: nit: it should be fine like this, since `gasLeft` is already of type `tosca.Gas`
```suggestion
	return transaction.GasLimit - gasLeft
```

HerbertJordan (approved):

HerbertJordan (approved):

LuisPH3 (approved):
lgtm

PR #626, opened 2024-07-23
Run examples through processor

simonlechner:
To ensure transaction code is handled correctly by the processor and passed to the interpreter, the example codes are now also executed on a processor level. Output handling is also checked with these test runs. 
Examples with recursive calls should be added in the future.

Most of the changes in this PR are due to renaming to make fields and functions accessible outside of the examples package.

### DIFF
--- a/go/integration_test/processor/examples_test.go
+++ b/go/integration_test/processor/examples_test.go
@@ -1,0 +1,75 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package processor
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/examples"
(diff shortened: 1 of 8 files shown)


LuisPH3 (approved):
The changes add test routines that test all the examples.

go/integration_test/processor/examples_test.go
```diff
@@ -0,0 +1,99 @@
...
+					if want != got.Result {
+						t.Fatalf("incorrect result, wanted %d, got %d", want, got.Result)
+					}
+				})
+			}
+		}
+	}
+}
+
+func RunOnProcessor(example examples.Example, processor tosca.Processor, argument int) (examples.Result, error) {
```
HerbertJordan, line 54: The `RunOn` for the interpreter is located [here](https://github.com/Fantom-foundation/Tosca/blob/3bd91c86e2aaacf4bd136c2aa4d05b3d7d356a47/go/examples/example.go#L52), inside the example package. Maybe the one for the processor should be placed next to it. This would eliminate the need for making example-package internal fields and functions public and allow the examples -- if this would ever be needed -- to be in control of e.g. block parameters.

The `TransactionContext` derived from the `Scenario` could be passed as an argument to the `RunOnProcessor` function.

Wdyt?

HerbertJordan (approved):
Looks good, just wondering whether the execution of examples should be moved to the examples package. Your call.

HerbertJordan (approved):

PR #627, opened 2024-07-24
Add max depth integration test

simonlechner:
This PR adds an integration test stressing the maximal recursive call depth in a single transaction.
This test can also also be used to implement recursive calls in floria without the need of implementing different types of calls.

### DIFF
--- a/go/integration_test/processor/calls_test.go
+++ b/go/integration_test/processor/calls_test.go
@@ -1,0 +1,104 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package processor
+
+import (
+	"fmt"
+	"math/big"
+	"strings"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
+)
+
+func TestProcessor_MaximalCallDepthIsEnforced(t *testing.T) {
+	gasLimit := tosca.Gas(1000000000000)
+	for processorName, processor := range getProcessors() {
+		// todo implement recurisve calls and depth checks in floria
+		if strings.Contains(processorName, "floria") {
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


go/integration_test/processor/calls_test.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		t.Run(fmt.Sprintf("%s-MaxCallDepth", processorName), func(t *testing.T) {
+			sender := tosca.Address{1}
+			receiver := &tosca.Address{2}
+
+			values := []*big.Int{big.NewInt(32), big.NewInt(0),
+				big.NewInt(32), big.NewInt(0), big.NewInt(0),
+				new(big.Int).SetBytes(receiver[:]), big.NewInt(int64(gasLimit))}
+			callStack := pushToStack(values)
```
HerbertJordan, line 28: nit: the variable name `callStack` is a bit misleading ... since in the end it is code establishing the stack listed above.
Also, it is only used at one position below. Creating it there with something like
```
code = append(code, pushToStack([]*big.Int{
    big.NewInt(32),                                // result size
    big.NewInt(0),                                  // result offset
    big.NewInt(32),                                // argument size
    big.NewInt(0),                                  // argument offset
    big.NewInt(0),                                  // transfered value
    new(big.Int).SetBytes(receiver[:]),   // call target
    big.NewInt(int64(gasLimit)),             // gas send to nested call
})
```
would clean this up. Since this is the reverse order of parameters as usually used when looking at the CALL operation specification, it might also be a nice feature to invert the parameter order.

Wdyt?

go/integration_test/processor/calls_test.go
```diff
@@ -0,0 +1,91 @@
...
+			// add stack values for call instruction
+			code = append(code, callStack...)
+
+			// make inner call and return 32byte value with 0 offset from memory
+			codeReturn := []byte{
+				byte(vm.CALL),
+				byte(vm.PUSH1), byte(32),
+				byte(vm.PUSH1), byte(0),
+				byte(vm.RETURN)}
+			code = append(code, codeReturn...)
```
HerbertJordan, line 49: nit:
```suggestion
			code = append(code, []byte{
				byte(vm.CALL),
				byte(vm.PUSH1), byte(32),
				byte(vm.PUSH1), byte(0),
				byte(vm.RETURN),
			}...)
```

HerbertJordan (dismissed):

go/integration_test/processor/calls_test.go
simonlechner: Nice suggestion, I changed the argument order and appended it directly to the `code`.

HerbertJordan (approved):

PR #628, opened 2024-07-25
Floria: Implement run context and enable recursive calls

simonlechner:
This PR depends on #627 and implements the run context which allows for recursive calls from the interpreter.


go/processor/floria/processor.go
```diff
@@ -23,6 +23,8 @@ const (
 	TxDataZeroGasEIP2028      = 4
 	TxAccessListAddressGas    = 2400
 	TxAccessListStorageKeyGas = 1900
+
+	CallCreateDepth = 1024 // Maximum depth of call/create stack.
```
HerbertJordan, line 27: nit: `CallCreateDepth` => `MaxRecursiveDepth` or `MaxCallNestingDepth`

go/processor/floria/processor.go
```diff
@@ -93,6 +112,14 @@ func (p *processor) Run(
 	}, nil
 }
 
+type runContext struct {
+	tosca.TransactionContext
+	interpreter           tosca.Interpreter
+	blockParams           tosca.BlockParameters
+	transactionParameters tosca.TransactionParameters
```
HerbertJordan, line 119: nit: would be nice to be consistent ... either both are `Params` or `Parameters`

go/processor/floria/calls.go
```diff
@@ -16,62 +16,70 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 )
 
-func call(interpreter tosca.Interpreter, transaction tosca.Transaction, context tosca.TransactionContext, gas tosca.Gas) (tosca.Result, error) {
+func (r runContext) Call(kind tosca.CallKind, parameters tosca.CallParameters) (tosca.CallResult, error) {
```
HerbertJordan, line 19: It is surprising that the `runContext` is defined in the `processor.go` but its main only member function `Call` is defined here. I think the type definition should be moved here and the file may be renamed from `call.go` to `run.go` or `run_context.go`. Wdyt?

go/processor/floria/calls.go
```diff
@@ -16,62 +16,70 @@ import (
...
 		return fmt.Errorf("insufficient balance: %v < %v", senderBalance, transferValue)
 	}
 
 	senderBalance = tosca.Sub(senderBalance, transferValue)
-	context.SetBalance(transaction.Sender, senderBalance)
+	context.SetBalance(sender, senderBalance)
 
-	receiverBalance := context.GetBalance(*transaction.Recipient)
+	receiverBalance := context.GetBalance(recipient)
 	receiverBalance = tosca.Add(receiverBalance, transferValue)
```
HerbertJordan, line 81: If we want to play it super-safe we would have to check that adding the transferred value is not leading to an overflow at the receiver. In such a case, we would also let the transfer fail.

go/processor/floria/calls.go
```diff
@@ -16,62 +16,70 @@ import (
...
-		Recipient: *transaction.Recipient,
-		Sender:    transaction.Sender,
-		Input:     transaction.Input,
-		Value:     transaction.Value,
-		CodeHash:  &codeHash,
-		Code:      code,
+		BlockParameters:       r.blockParams,
+		TransactionParameters: r.transactionParameters,
+		Context:               r,
+		Kind:                  tosca.Call,
```
HerbertJordan, line 34: I guess the `Kind` should be `kind` ... but this will get unearthed when testing different call types.

HerbertJordan (changes requested):
The architecture looks good, just a few requests regarding code organization. Please have a look.

go/processor/floria/calls.go
```diff
@@ -16,62 +16,70 @@ import (
...
+) error {
+	senderBalance := context.GetBalance(sender)
+	transferValue := value
 	if senderBalance.Cmp(transferValue) < 0 {
 		return fmt.Errorf("insufficient balance: %v < %v", senderBalance, transferValue)
 	}
 
 	senderBalance = tosca.Sub(senderBalance, transferValue)
-	context.SetBalance(transaction.Sender, senderBalance)
+	context.SetBalance(sender, senderBalance)
```
LuisPH3, line 78: is negative balance allowed?

LuisPH3 (dismissed):

go/processor/floria/calls.go
simonlechner: No, there is a check whether the sender has sufficient balance.

go/processor/floria/calls.go
simonlechner: Right, I moved it to the calls file and renamed it to run_context.go

go/processor/floria/calls.go
simonlechner: Right, this has been missed in the last PR were this function was introduced. I restructured the whole function and added more unit tests.

Issue #603, opened 2024-07-11
Floria: Gas computation and billing


simonlechner:
To support a simple value transfer, the functionality of buying gas has been added in #601. Gas has to be handled and refunded correctly. Extensive unit testing and integration test cases should be added to close this issue.

PR #629, opened 2024-07-29
Floria: Implement gas refunding

### LINKED ISSUE #575: Floria: Add gas refund after fail call

simonlechner:
This PR implements gas refunding in Floria. Further, the gas integration tests are split into multiple functions and an end to end test for gas billing/handling using a mock interpreter is introduced. 

Fixes #575 and #603.

### DIFF
--- a/go/integration_test/processor/gas_billing_test.go
+++ b/go/integration_test/processor/gas_billing_test.go
@@ -1,0 +1,118 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package processor
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
+	op "github.com/ethereum/go-ethereum/core/vm"
+	"go.uber.org/mock/gomock"
+
+	_ "github.com/Fantom-foundation/Tosca/go/processor/opera" // < registers opera processor for testing
+)
+
+func TestProcessor_GasBillingEndToEnd(t *testing.T) {
+	senderBalance := tosca.NewValue(1000000)
+	gasLimit := tosca.Gas(100000)
+	gasRefund := tosca.Gas(3000)
+	gasPrice := tosca.NewValue(5)
+	gasLeftSuccess := tosca.Gas(5000)
+
+	tests := map[string]struct {
+		result  tosca.Result
+		gasUsed tosca.Gas
+		success bool
+	}{
+		"success": {
+			result: tosca.Result{
+				GasLeft:   gasLeftSuccess,
+				Success:   true,
+				GasRefund: gasRefund,
+			},
+			gasUsed: gasLimit - (gasLeftSuccess - gasLeftSuccess/10 + gasRefund),
+			success: true,
+		},
+		"failed": {
+			result: tosca.Result{
+				GasLeft:   0,
+				Success:   false,
+				GasRefund: gasRefund,
+			},
+			gasUsed: gasLimit,
+			success: false,
+		},
+	}
+
+	ctrl := gomock.NewController(t)
+	interpreter := tosca.NewMockInterpreter(ctrl)
+
+	sender := tosca.Address{1}
+	recipient := tosca.Address{2}
+	before := WorldState{
+		sender: Account{Balance: senderBalance, Nonce: 4},
+		recipient: Account{Balance: tosca.NewValue(0),
+			Code: tosca.Code{
+				byte(vm.PUSH1), byte(0), // < push 0
+				byte(vm.PUSH1), byte(0), // < push 0
+				byte(op.RETURN),
+			},
+		},
+	}
+
+	transaction := tosca.Transaction{
+		Sender:    sender,
+		Recipient: &recipient,
+		GasLimit:  gasLimit,
+		GasPrice:  gasPrice,
+		Nonce:     4,
+	}
+
+	for name, test := range tests {
+		for processorName, processor := range processorsWithInterpreter("mockInterpreter", interpreter) {
+			t.Run(fmt.Sprintf("%s/%s", processorName, name), func(t *testing.T) {
+
+				after := before.Clone()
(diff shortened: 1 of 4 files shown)


go/integration_test/processor/gas_billing_test.go
```diff
@@ -0,0 +1,126 @@
...
+			},
+			result: tosca.Result{
+				GasLeft:   5000,
+				Success:   true,
+				GasRefund: 300,
+			},
+		},
+		"failed": {
+			after: WorldState{
+				{1}: Account{Balance: tosca.NewValue(522500), Nonce: 5},
```
HerbertJordan, line 55: The only difference between the two after states in these tests seems to be the balance of account 1. It should reduce the code complexity a lot if you only put the balance in the test case specification and build the rest in the test run function below.

go/integration_test/processor/gas_billing_test.go
```diff
@@ -0,0 +1,126 @@
...
+			},
+			receipt: tosca.Receipt{
+				Success: true,
+				GasUsed: 95200,
+			},
+			result: tosca.Result{
+				GasLeft:   5000,
+				Success:   true,
+				GasRefund: 300,
+			},
```
HerbertJordan, line 51: nit: I think definitions are in general easier to read if they are causally ordered; Input parameters should be followed by output vlaues.

In this test case, the `result` leads to the `after` state and `receipt`. So I would argue that `result` should be defined first, followed by `after` and `receipt`. Also, `result` as a name for an input parameter makes reading this code hard.

go/integration_test/processor/gas_billing_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func TestProcessor_GasBillingEndToEnd(t *testing.T) {
+	tests := map[string]struct {
+		after   WorldState
+		receipt tosca.Receipt
+		result  tosca.Result
+	}{
+		"success": {
+			after: WorldState{
+				{1}: Account{Balance: tosca.NewValue(524000), Nonce: 5},
```
HerbertJordan, line 34: I am having a hard time understanding where this final balance is coming from ... so it is difficult to understand the specification covered by this test.

go/integration_test/processor/processor_gas_test.go
```diff
@@ -145,132 +143,225 @@ func getGasTestScenarios() map[string]Scenario {
...
+		"GasPriceCalculation": {
+			Before: WorldState{
+				{1}: Account{Balance: tosca.NewValue(floria.TxGas * gasPrice), Nonce: 4},
+				{2}: Account{},
+			},
+			Transaction: tosca.Transaction{
+				Sender:    tosca.Address{1},
+				Recipient: &tosca.Address{2},
+				GasLimit:  floria.TxGas,
+				GasPrice:  tosca.ValueFromUint256(uint256.NewInt(gasPrice)),
```
HerbertJordan, line 293: Here and at other places:
```suggestion
				GasPrice:  tosca.NewValue(gasPrice),
```

go/integration_test/processor/processor_gas_test.go
```diff
@@ -145,132 +143,225 @@ func getGasTestScenarios() map[string]Scenario {
...
-			Sender:    tosca.Address{},
-			Recipient: &tosca.Address{2},
-			GasLimit:  floria.TxGas + 100,
-			Nonce:     4,
+	return testCases
+}
+
+func gasPricingTestCases() map[string]Scenario {
+	gasPrice := uint64(10)
+	testCases := map[string]Scenario{
```
HerbertJordan, line 283: I would try to minimize the details that need to be specified in those cases. Everything that is the same in all of those should be handled in the Run function below.

Here, it looks like all you need is the before and after value for the account's balance and the receipt.

go/integration_test/processor/processor_gas_test.go
```diff
@@ -145,132 +143,225 @@ func getGasTestScenarios() map[string]Scenario {
...
+			},
+			Receipt: tosca.Receipt{
+				Success: true,
+				GasUsed: floria.TxGas,
+			},
 		},
-		After: WorldState{
-			{}:  Account{Balance: tosca.NewValue(100), Nonce: 5},
-			{2}: Account{Balance: tosca.NewValue(0)},
+		"GasPriceCalculationExcessGas": {
```
HerbertJordan, line 306: This label is wrong. It is not "excess gas" but "excess balance".

HerbertJordan (changes requested):

HerbertJordan (commented):
You should try to minimize test cases to make them more readable and verifiable.

go/integration_test/processor/gas_billing_test.go
simonlechner: I don't think it makes sense to write down the whole gas computation in the unit tests, the idea of this end to end test is to make sure both the reference implementation as well as floria deliver the same results. I added a comment.

go/integration_test/processor/processor_gas_test.go
simonlechner: I was able to reduce the unit test length and also restructured parts of the other gas specific tests.

go/integration_test/processor/gas_billing_test.go
HerbertJordan: Something to consider though: unlike in the interpreter case, the reference version of the processor may be buggy.

go/processor/floria/processor_test.go
```diff
@@ -116,41 +115,138 @@ func TestProcessor_BuyGasInsufficientBalance(t *testing.T) {
...
+				Sender:   tosca.Address{},
+				GasLimit: 1000,
+			},
+			result: tosca.CallResult{
+				GasLeft:   500,
+				Success:   false,
+				GasRefund: 500,
+			},
+			revision:        tosca.R10_London,
+			expectedGasLeft: 500,
```
HerbertJordan, line 213: In case of an unsuccessful execution all gas should be consumed .. so this value should be `0`, right?

The implementation of `calculateGasLeft` seems to assume that in case of a unsuccessful execution `result.GasLeft` is `0`. This is probably the case for all interpreter implementations, and so the implementation will likely produce the correct result. If so, this test case describes an impossible input.

HerbertJordan (approved):

go/processor/floria/processor_test.go
simonlechner: Right, I mainly wanted to test that the refund is not added back in an unsuccessful case. I changed the gasLeft to 0 to avoid any confusion.

HerbertJordan (approved):

HerbertJordan (approved):
Thanks for documenting the expectations using code!

PR #630, opened 2024-07-31
Add integration test for different call types

simonlechner:
One of the biggest differences between the different call types is whether they access the same storage. An integration test writing into storage and reading from the same slot in an inner call is added in this PR.
Some minor restructuring as preparation for the implementation in Floria is also performed.


HerbertJordan (approved):
Looks good, thanks!

HerbertJordan (approved):

PR #631, opened 2024-08-01
Add commit validation stage

wsodsong:
This PR updates the Jenkinsfile such that it first validates whether a commit is from trustable source before starting PR tests.


HerbertJordan (approved):
Thanks for adding this check!

Issue #402, opened 2024-04-10
VM: integrate Tosca EVM interface into Aida


HerbertJordan:
To facilitate the testing of the Tosca EVM in Aida's integration test framework, the interface needs to be integrated into Aida.

PR #632, opened 2024-08-01
Suppress export of geth reference interpreter to go-ethereum

HerbertJordan:
This PR fixes the support for running the `geth` interpreter with the `aida` or `floria` processor in Aida.

The change in this PR prevents the `geth` reference interpreter maintained by Tosca to replace the `geth` interpreter implementation in a host system like Sonic or Aida. Before this change, the geth version included in the host system got replaced by a version with the following structure:
```
gethToToscaAdapter(toscaToGethAdapter(gethReferenceInterpreter)))
```
This is a surprising setup, that is not the expected configuration when users are selecting the `geth` VM option in Aida or Sonic.

The issue was not discovered earlier since the [RegisterInterpreterFactory](https://github.com/Fantom-foundation/go-ethereum-sonic/blob/2400937cc3b1c370ea8735d7af86eac621e4eb70/core/vm/tosca_integration.go#L72) function in `go-ethereum-sonic` is not checking for name collisions. This should be fixed eventually, but even if it would be checked, the correct response would have been a panic. Thus, the call needs to be prevented in the first place anyway.

This is part of #402

### DIFF
--- a/go/geth_adapter/adapter.go
+++ b/go/geth_adapter/adapter.go
@@ -38,7 +38,15 @@ const adapterDebug = false
 
 func init() {
 	for name, interpreter := range tosca.GetAllRegisteredInterpreters() {
-		RegisterGethInterpreter(name, interpreter)
+		// We register all tosca interpreters except the geth reference interpreter.
+		// Tosca's geth reference implementation in combination with the adapter
+		// implemented would lead to unexpected behavior since geth would no longer
+		// use its own interpreter when requesting the `geth` interpreter. Instead
+		// Tosca's geth reference wrapped in the adapter would be used -- a combination
+		// that is not well tested.
+		if name == "" || name != "geth" {
+			RegisterGethInterpreter(name, interpreter)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


simonlechner (approved):
Thank you for fixing.

LuisPH3 (approved):
It may fix the bug, nevertheless the change feels more like a patch for a problem than a solution.

It is strangely odd to have a module being called geth_adapter which includes all VMs except for geth.

IMO: Unexpected behavior and global out-of-order initialization go hand by hand.

LuisPH3:
> It may fix the bug, nevertheless the change feels more like a patch for a problem than a solution.
I see the problem now:

The processor includes geth [here](https://github.com/Fantom-foundation/Tosca/blob/29d864860cbc5083431fa2c91c7d5ef8f859fa44/go/processor/opera/processor.go#L28). This will initialize the register with geth for an instance of Tosca. This violates the register intention to have optional modules and forces the dependency on the geth version. 

The problem is indeed hard to resolve.

PR #633, opened 2024-08-02
Processor calls integration tests

simonlechner:
To ensure the different call types are implemented correctly more integration tests are added checking the call value, sender and whether it is in static context or not. All tests return the stressed property as the return value of the transaction.

### DIFF
--- a/go/integration_test/processor/calls_test.go
+++ b/go/integration_test/processor/calls_test.go
@@ -22,8 +22,50 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
 )
 
+const sufficientGas = tosca.Gas(1000000000000)
+
+type callProperties struct {
+	callType    vm.OpCode
+	hasValue    bool
+	sameStorage bool
+	sameValue   bool
+	sameSender  bool
+}
+
+func callTypesAndProperties() map[string]callProperties {
+	return map[string]callProperties{
+		"call": {
+			callType:    vm.CALL,
+			hasValue:    true,
+			sameStorage: false,
+			sameValue:   false,
+			sameSender:  false,
+		},
+		"callCode": {
+			callType:    vm.CALLCODE,
+			hasValue:    true,
+			sameStorage: true,
+			sameValue:   false,
+			sameSender:  false,
+		},
+		"delegateCall": {
+			callType:    vm.DELEGATECALL,
+			hasValue:    false,
+			sameStorage: true,
+			sameValue:   true,
+			sameSender:  true,
+		},
+		"staticCall": {
+			callType:    vm.STATICCALL,
+			hasValue:    false,
+			sameStorage: false,
+			sameValue:   false,
+			sameSender:  false,
+		},
+	}
+}
+
 func TestProcessor_MaximalCallDepthIsEnforced(t *testing.T) {
-	gasLimit := tosca.Gas(1000000000000)
 	for processorName, processor := range getProcessors() {
 		t.Run(fmt.Sprintf("%s-MaxCallDepth", processorName), func(t *testing.T) {
 			sender := tosca.Address{1}
@@ -41,7 +83,7 @@ func TestProcessor_MaximalCallDepthIsEnforced(t *testing.T) {
 
 			// add stack values for call instruction
 			code = append(code, pushToStack([]*big.Int{
-				big.NewInt(int64(gasLimit)),        // gas send to nested call
+				big.NewInt(int64(sufficientGas)),   // gas send to nested call
 				new(big.Int).SetBytes(receiver[:]), // call target
 				big.NewInt(0),                      // value to transfer
 				big.NewInt(0),                      // argument offset
@@ -62,11 +104,11 @@ func TestProcessor_MaximalCallDepthIsEnforced(t *testing.T) {
 			transaction := tosca.Transaction{
 				Sender:    sender,
 				Recipient: receiver,
-				GasLimit:  gasLimit,
+				GasLimit:  sufficientGas,
 				Nonce:     0,
 				Input:     tosca.Data{},
 			}
-			scenario := getScenarioContext(sender, *receiver, code, gasLimit)
+			scenario := getScenarioContext(sender, *receiver, code, sufficientGas)
 			transactionContext := newScenarioContext(scenario.Before)
 
 			// Run the processor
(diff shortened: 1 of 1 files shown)


go/integration_test/processor/calls_test.go
```diff
@@ -177,13 +172,11 @@ func TestProcessor_DifferentCallTypesAccessStorage(t *testing.T) {
 					receiver1: Account{Code: code1},
 				}
 				scenario := Scenario{
```
facuMH, line 174: is `scenario` needed if you only use the `Transaction` ?

go/integration_test/processor/calls_test.go
```diff
@@ -193,20 +186,266 @@ func TestProcessor_DifferentCallTypesAccessStorage(t *testing.T) {
...
+		for callName, call := range callTypesAndProperties() {
+			t.Run(fmt.Sprintf("%s-%s", processorName, callName), func(t *testing.T) {
+				sender0 := tosca.Address{1}
+				receiver0 := tosca.Address{2}
+				receiver1 := tosca.Address{3}
+				senderBalance := tosca.NewValue(100)
+				transferValue := tosca.NewValue(42)
+
+				// set call arguments
+				code0 := pushCallArguments(call, gasLimit, tosca.Value{}, receiver1)
```
HerbertJordan, line 218: An idea for a small improvement: you could set a different value here to also check for those call types that do have a value parameter that the value passed as an argument is reaching the inner call

HerbertJordan (dismissed):
Looks like a good set of tests :+1:

go/integration_test/processor/calls_test.go
```diff
@@ -193,20 +186,266 @@ func TestProcessor_DifferentCallTypesAccessStorage(t *testing.T) {
...
+						t.Errorf("%s did access the same storage, got unexpected output: %v", callName, result.Output)
 					}
 				}
 			})
 		}
 	}
 }
 
+func TestProcessor_DifferentCallTypesHandleValueCorrectly(t *testing.T) {
+	gasLimit := tosca.Gas(1000000)
```
facuMH, line 204: should we have a const for this amount of gas with a meaninful name ?

go/integration_test/processor/calls_test.go
```diff
@@ -193,20 +186,266 @@ func TestProcessor_DifferentCallTypesAccessStorage(t *testing.T) {
...
+					byte(vm.PUSH1), byte(0),
+					byte(vm.RETURN),
+				}
+
+				state := WorldState{
+					sender0:   Account{Balance: senderBalance},
+					receiver0: Account{Code: code0},
+					receiver1: Account{Code: code1},
+				}
+				scenario := Scenario{
```
facuMH, line 242: same as before,  `scenario` is not needed, just the `Transaction`

go/integration_test/processor/calls_test.go
```diff
@@ -193,20 +186,266 @@ func TestProcessor_DifferentCallTypesAccessStorage(t *testing.T) {
...
+
+func pushCallArguments(
+	call callProperties,
+	gasLimit tosca.Gas,
+	transferValue tosca.Value,
+	receiver tosca.Address,
+) []byte {
+	var code []byte
+	if call.hasValue {
+		code = pushToStack([]*big.Int{
```
facuMH, line 427: would it make sense to use `tosca.Word` instead of `*big.Int` ? there shouldn't be any impact, I just thought "why is big int needed here?"

facuMH (dismissed):
nice test structure !! just a couple of minor suggestions

go/integration_test/processor/calls_test.go
simonlechner: We had a discussion about this when I implemented the `pushToStack` function, due to the variable length of big int it is very easy to calculate the right `PUSH` function. I tried to change the type to `tosca.Word` but things got a little messy and unreadable.

facuMH (approved):

HerbertJordan (approved):

go/integration_test/processor/calls_test.go
```diff
@@ -176,37 +171,274 @@ func TestProcessor_DifferentCallTypesAccessStorage(t *testing.T) {
...
+				transactionContext := newScenarioContext(state)
+
+				// Run the processor
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil || !result.Success {
+					t.Errorf("execution was not successful or failed with error %v", err)
+				}
+				if !slices.Equal(result.Output, bytes.Repeat([]byte{0}, 32)) {
+					t.Errorf("%s did not enforce static mode and returned %v", callName, result.Output)
+				}
```
LuisPH3, line 406: Aren't we missing a check here for the reverted storage mentioned above?

go/integration_test/processor/calls_test.go
```diff
@@ -176,37 +171,274 @@ func TestProcessor_DifferentCallTypesAccessStorage(t *testing.T) {
...
+		}
+		calls := callTypesAndProperties()
+		for callName, call := range calls {
+			t.Run(fmt.Sprintf("%s-%s", processorName, callName), func(t *testing.T) {
+
+				// Test structure:
+				// sender0 --CALL--> receiver0 --STATICCALL--> receiver1
+				// --test.CALL--> receiver2 -> SSTORE
+				// after the first static calls all following calls have to be static,
+				// therefore the SSTORE should revert. The success boolean of the test.CALL
```
LuisPH3, line 347: A thing about writing: should is conditional, this means that not doing it is acceptable.
- Use shall when you want to be specific about intention. 
- Use may when it is intentionally conditional. 

 (requirements texts avoid using must, will, could, and should because of different semantics)

go/integration_test/processor/calls_test.go
```diff
@@ -22,8 +22,50 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
 )
 
+const sufficientGas = tosca.Gas(1000000000000)
```
LuisPH3, line 25:
```suggestion
const sufficientGas = tosca.Gas(1000_000_000_000)
```

This gas is double as the maximum allowed by the CT (to prevent memory expansions) Therefore using this
amount of gas may explore untested code regions.

https://github.com/Fantom-foundation/Tosca/blob/main/go/ct/st/state.go#L49

LuisPH3 (approved):
The tests stress calls and seem to cover all scenarios I can think of. 
please check comments, and apply on your own criteria.

go/integration_test/processor/calls_test.go
simonlechner: I don't want to enforce a certain storage state in a failed case, this is handled by the interpreter not the processor.

go/integration_test/processor/calls_test.go
simonlechner: Good catch, I changed it to the same value as `MaxGasUsedByCt`.

PR #636, opened 2024-08-06
Github report coverage

facuMH:
This PR adds a github workflow, a makefile target and a config file. These combined produce a report on the coverage and (when enabled) post this report on the PR.

This is currently disabled because the unit-test coverage of some packages does not meet the current requirements. This needs to be fixed in a separate PR or lowered thresholds. 

NOTE: 
a series of follow-up PRs will add unit tests to increase coverage and eventually empty the list of overrides

### DIFF
--- a/.testcoverage.yml
+++ b/.testcoverage.yml
@@ -1,0 +1,59 @@
+# (mandatory) 
+# Path to coverprofile file (output of `go test -coverprofile` command).
+#
+# For cases where there are many coverage profiles, such as when running 
+# unit tests and integration tests separately, you can combine all those
+# profiles into one. In this case, the profile should have a comma-separated list 
+# of profile files, e.g., 'cover_unit.out,cover_integration.out'.
+profile: cover.out
+
+# (optional; but recommended to set) 
+# When specified reported file paths will not contain local prefix in the output
+local-prefix: "github.com/org/project"
+
+# Holds coverage thresholds percentages, values should be in range [0-100]
+threshold:
+  # (optional; default 0) 
+  # The minimum coverage that each file should have
+  file: 40
+
+  # (optional; default 0) 
+  # The minimum coverage that each package should have
+  package: 80
+
+  # (optional; default 0) 
+  # The minimum total coverage project should have
+  total: 95
+
+# Holds regexp rules which will override thresholds for matched files or packages 
+# using their paths.
+#
+# First rule from this list that matches file or package is going to apply 
+# new threshold to it. If project has multiple rules that match same path, 
+# override rules should be listed in order from specific to more general rules.
+override:
+  # Increase coverage threshold to 100% for `foo` package 
+  # (default is 80, as configured above in this example)
+  # - threshold: 100
+  #   path: ^pkg/lib/foo$
+
+# Holds regexp rules which will exclude matched files or packages 
+# from coverage statistics
+exclude:
+  # Exclude files or packages matching their paths
+  paths:
+    - ^(.+)mock(.+)         # mock files should not be tested
+    - go/ct/common/error.go # error file only declares error type without any logic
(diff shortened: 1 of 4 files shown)


.testcoverage.yml
```diff
@@ -0,0 +1,153 @@
...
+# from coverage statistics
+exclude:
+  # Exclude files or packages matching their paths
+  paths:
+    - ^(.+)mock(.+)         # mock files should not be tested
+    - go/ct/common/error.go # error file only declares error type without any logic
+    - go/ct/driver/         # entry points, logic should be tested elsewhere.
+    - go/examples           # examples are not tested.
+    - go/interpreter/lfvm/errors.go # error file only declares error type without any logic
+    - go/interpreter/lfvm/instructions.go # these are tested by driver
```
simonlechner, line 146: I think eventually there should also be unit tests for instructions in lfvm.

.testcoverage.yml
```diff
@@ -0,0 +1,153 @@
...
+  # The minimum total coverage project should have
+  total: 68
+
+# Holds regexp rules which will override thresholds for matched files or packages 
+# using their paths.
+#
+# First rule from this list that matches file or package is going to apply 
+# new threshold to it. If project has multiple rules that match same path, 
+# override rules should be listed in order from specific to more general rules.
+override:
```
simonlechner, line 44: Why override all these thresholds and not just lower the threshold for files and packages?

simonlechner (commented):
Is it required to specify thresholds and not just get the current coverage? I don't think it makes sense to set thresholds now, while we still have to override them for a substantial amount of the packages.

.testcoverage.yml
facuMH: I opted for this path because it forces us to list what needs fixing rather than accepting everything.

.testcoverage.yml
facuMH: like the list of overrides, this can also be an issue with its own pr and tests.

facuMH:
> Is it required to specify thresholds and not just get the current coverage? I don't think it makes sense to set thresholds now, while we still have to override them for a substantial amount of the packages.

it is not required. And I agree, partly, that setting the thresholds now is a little pointless. That said, I still see value in having a value there, and forcing us to list what needs fixing.

LuisPH3 (approved):

go/ct/common/hash.go
```diff
@@ -15,7 +15,7 @@ import (
 	"pgregory.net/rand"
 )
 
-func GetRandomHash(rnd *rand.Rand) tosca.Hash {
+func GetRandomHash(rnd *rand.Rand) tosca.Hash { // coverage-ignore nothing to test
```
LuisPH3, line 18: two things, why do we need this? 
second: I tried it locally, and it does not ignore my code, is it correct?

facuMH: this comment was the first try to exclude/ignore files/pkg/blocks. Also this file should not be considered for testing since it has only one function that calls to `rnd.Read`, so it would be like testing that function.

HerbertJordan: I do not think this is a good argument for why this should not be tested. Please remove the `coverage-ignore` comment.

A test just running it to make sure it does not crash would be sufficient in this case.

HerbertJordan (dismissed):

go/ct/common/hash.go
facuMH: a test has been added for this in the follow-up PR #639

HerbertJordan (dismissed):

simonlechner (dismissed):

LuisPH3:
- [ ] fix main branch issue.

LuisPH3 (approved):
let see if this fix the main branch bug.

simonlechner (approved):

PR #638, opened 2024-08-06
Integration tests for CREATE and CREATE2

simonlechner:
This PR adds integration tests for CREATE and CREATE2 instructions. 
Due to the required setup of `initCode` and `toBeCreatedCode` the tests are rather long, to still keep them readable small numbers of lines is group and commented.

### DIFF
--- a/go/integration_test/processor/create_test.go
+++ b/go/integration_test/processor/create_test.go
@@ -1,0 +1,349 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package processor
+
+import (
+	"bytes"
+	"fmt"
+	"math/big"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
+)
+
+func TestProcessor_CreateAndCallContract(t *testing.T) {
+	gasLimit := big.NewInt(int64(sufficientGas))
+	gasPush := vm.PUSH1 + vm.OpCode(len(gasLimit.Bytes())-1)
+
+	for processorName, processor := range getProcessors() {
+		if strings.Contains(processorName, "floria") {
+			continue // todo implement different create types
+		}
+		for _, create := range []vm.OpCode{vm.CREATE, vm.CREATE2} {
+			t.Run(fmt.Sprintf("%s-%s", processorName, create.String()), func(t *testing.T) {
+				sender0 := tosca.Address{1}
+				receiver0 := tosca.Address{2}
+				toBeCreatedCodeHolder := tosca.Address{3}
+				initCodeHolder := tosca.Address{4}
+
+				initCodeOffset := 32
+
+				input := byte(42)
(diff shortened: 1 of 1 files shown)


go/integration_test/processor/create_test.go
```diff
@@ -0,0 +1,349 @@
...
+func TestProcessor_CreateAndCallContract(t *testing.T) {
+	gasLimit := big.NewInt(int64(sufficientGas))
+	gasPush := vm.PUSH1 + vm.OpCode(len(gasLimit.Bytes())-1)
+
+	for processorName, processor := range getProcessors() {
+		if strings.Contains(processorName, "floria") {
+			continue // todo implement different create types
+		}
+		for _, create := range []vm.OpCode{vm.CREATE, vm.CREATE2} {
+			t.Run(fmt.Sprintf("%s-%s", processorName, create.String()), func(t *testing.T) {
```
LuisPH3, line 34: Note: I realized that you can nest `t.Run` calls, and the testing infrastructure namespaces the tests for you.
this way you could t.Skip instead of continue in the processor loop 

It would be quite the same, but a little more expressive regarding tests.

LuisPH3 (approved):
I don't think these tests are too long, There are clean and straight forward, doing what they should do.

go/integration_test/processor/create_test.go
simonlechner: Thank you for the suggestion, I will keep this in mind for the next one. I am already working on a PR removing those lines, as they are only temporary I think they can stay in this PR.

LuisPH3: yes they can, you don't have to change anything.

facuMH (dismissed):

go/integration_test/processor/create_test.go
```diff
@@ -0,0 +1,349 @@
...
+		transactionContext := newScenarioContext(state)
+
+		// Run the processor
+		result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+		if err != nil || !result.Success {
+			t.Errorf("execution was not successful or failed with error %v", err)
+		}
+		if !slices.Equal(result.Output, append(bytes.Repeat([]byte{0}, 31), checkValue)) {
+			t.Errorf("creation of account was not successful, returned %v", result.Output)
+		}
```
HerbertJordan, line 331: The result should contain a `ContractAddress` field that should also be checked here whether it contains the address of the created account.

HerbertJordan (approved):

go/integration_test/processor/create_test.go
simonlechner: Added it to this integration tests but I mainly want to stress the created account address with unit tests during the implementation in Floria.

go/integration_test/processor/create_test.go
```diff
@@ -0,0 +1,355 @@
...
+	"bytes"
+	"fmt"
+	"math/big"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
+	"github.com/ethereum/go-ethereum/common"
```
facuMH, line 23: do we want to add more dependencies from geth or would it be too much work to implement our own address generator ?

simonlechner: I think inside the tests it is fine to have dependencies to geth, we need to ensure that the result is correct somehow. We also have this in other integration test files.

facuMH: I noticed them, and some of them seem easy to remove. but perhaps it should be another task

facuMH (approved):

PR #639, opened 2024-08-07
Improve coverage CT/common

facuMH:
This PR improves UT coverage of the ct/common package and removes the relevant files from the override list of the coverage test config file.

This PR depends on #636


go/ct/common/bytes_test.go
```diff
@@ -99,3 +101,42 @@ func TestBytes_InvalidJsonFails(t *testing.T) {
...
+
+func TestBytes_Length(t *testing.T) {
+	b := NewBytes([]byte{1, 2, 3})
+	if b.Length() != 3 {
+		t.Errorf("unexpected length, got %v", b.Length())
+	}
+}
+
+func TestBytes_Get(t *testing.T) {
+	b := NewBytes([]byte{1, 2, 3})
```
LuisPH3, line 138: same same, not sure if dead code.

go/ct/common/bytes_test.go
```diff
@@ -99,3 +101,42 @@ func TestBytes_InvalidJsonFails(t *testing.T) {
 		})
 	}
 }
+
+func TestBytes_RandomBytesProducesRandom(t *testing.T) {
+	rnd := rand.New()
+	for i := 0; i < 10000; i++ {
```
HerbertJordan, line 107: nit: This seems a bit excessive ...

go/ct/common/bytes_test.go
```diff
@@ -99,3 +101,42 @@ func TestBytes_InvalidJsonFails(t *testing.T) {
...
+		for _, b := range b1.data {
+			if b != 0 {
+				allZeros = false
+				break
+			}
+		}
+
+		if len(b1.data) > 1 && allZeros {
+			t.Errorf("RandomBytes produced all zeros: %v, len:v %v", b1.data, len(b1.data))
+		}
```
HerbertJordan, line 119: There is a certain probability that randomly something like `[]byte{0,0}` is generated, and it would be valid.
Also, this test would pass if `RandomBytes` would always return `[]byte{0,1}`.

Suggestion for an alternative testing strategy: generate 5 random byte sequences and check that there are no collisions.

go/ct/common/bytes_test.go
```diff
@@ -99,3 +101,42 @@ func TestBytes_InvalidJsonFails(t *testing.T) {
...
+func TestBytes_Length(t *testing.T) {
+	b := NewBytes([]byte{1, 2, 3})
+	if b.Length() != 3 {
+		t.Errorf("unexpected length, got %v", b.Length())
+	}
+}
+
+func TestBytes_Get(t *testing.T) {
+	b := NewBytes([]byte{1, 2, 3})
+	if got := b.Get(1, 2); string(got) != "\x02" {
```
HerbertJordan, line 139: This kinda depends on how byte slices are converted to strings. Maybe better:
```suggestion
	if got := b.Get(1, 2); bytes.Equal(got, []byte{2}) {
```

go/ct/common/immutable_hash_array.go
```diff
@@ -64,12 +64,12 @@ func (h *ImmutableHashArray) UnmarshalJSON(data []byte) error {
 	// Unmarshal the JSON array into a slice of tosca.Hash
 	var hashes [256]tosca.Hash
 	err := json.Unmarshal(data, &hashes)
-	if err != nil {
+	if err != nil { // coverage-ignore
```
HerbertJordan, line 67: We should use `// coverage-ignore` very sparsely, and if we do, with a proper explanation.

go/ct/common/cpp_coverage.go
```diff
@@ -30,6 +30,6 @@ func isCppCoverageEnabled() bool {
 // Not calling this function will result in no coverage data being reported
 // for runtime loaded C and C++ libraries.
 // If coverage data collection is not enabled, this function is a no-op.
-func DumpCppCoverageData() {
+func DumpCppCoverageData() { // coverage-ignore nothing to test
```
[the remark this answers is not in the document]
facuMH: as I understood it, the only way to make a test for this function would be to start it with the cpp coverage flag enabled. And this is not something we will do when running UT/coverage.

go/ct/common/immutable_hash_array_test.go
```diff
@@ -135,3 +136,36 @@ func TestImmutableHashArray_Get(t *testing.T) {
 		})
 	}
 }
+
+func TestINmmutableHashArray_NewRandomImmutableHashArray(t *testing.T) {
+	rnd := rand.New()
+
+	for i := 0; i < 100; i++ {
```
HerbertJordan, line 143: Same here as for the random hash: 
 - repetition is a bit excessive
 - an all-zero value is not really a problem
 - the test would not detect if the function would always return the same non-zero value

Idea: generate N samples and check that there are no duplicates.

go/ct/common/revisions_test.go
```diff
@@ -85,3 +86,27 @@ func TestRevision_GetRevisionForBlock(t *testing.T) {
 	}
 
 }
+
+func TestRevisions_GetForkTime(t *testing.T) {
+	tests := map[string]struct {
```
HerbertJordan, line 91: nit: the revision could be the key of this map

HerbertJordan (changes requested):
Please reduce the utilization of coverage-ignore

go/ct/common/immutable_hash_array.go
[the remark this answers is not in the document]
facuMH: I think that making a unit test for this error, would be testing that `json.Unmarshal` is working as expected, and we should not be testing that.

go/ct/common/bytes_test.go
```diff
@@ -99,3 +101,42 @@ func TestBytes_InvalidJsonFails(t *testing.T) {
...
+
+func TestBytes_String(t *testing.T) {
+	b := NewBytes([]byte{1, 2, 3})
+	if b.String() != "0x010203" {
+		t.Errorf("unexpected string representation, got %v", b.String())
+	}
+}
+
+func TestBytes_Length(t *testing.T) {
+	b := NewBytes([]byte{1, 2, 3})
```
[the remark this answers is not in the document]
facuMH: it is called, all the `XData` types in the state are of type `Bytes`, and it's `Lenght` is queried in multiple places. I will add the empty case.

go/ct/common/cpp_coverage.go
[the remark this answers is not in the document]
HerbertJordan: A simple call to this function making sure it does not panic of crash the test would be sufficient to get the coverage done.

The explanation "nothing to test" is a not acceptable, since "nothing to test" means also that this function has no purpose and should not exist.

As far as I remember, when @LuisPH3 added this function there have already been tests that make sure that if compiled with a special flag, dump data is actually produced. So this does not have to be tested again.

go/ct/common/hasd_test.go
HerbertJordan, line 1: there is a typo in the file name

go/ct/common/hasd_test.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"pgregory.net/rand"
+)
+
+func TestHash_GetRandomHash(t *testing.T) {
+	rnd := rand.New()
+	hash := GetRandomHash(rnd)
```
HerbertJordan, line 12: nit: maybe also use the 5-random value idea here?

go/ct/common/hash.go
```diff
@@ -17,6 +17,6 @@ import (
 
 func GetRandomHash(rnd *rand.Rand) tosca.Hash {
 	var res tosca.Hash
-	_, _ = rnd.Read(res[:]) // rnd.Read never returns an error
+	_, _ = rnd.Read(res[:])
```
HerbertJordan, line 20: Why would you remove this comment?

go/ct/common/revisions_test.go
```diff
@@ -85,3 +86,27 @@ func TestRevision_GetRevisionForBlock(t *testing.T) {
 	}
 
 }
+
+func TestRevisions_GetForkTime(t *testing.T) {
+	tests := map[tosca.Revision]struct {
+		revision tosca.Revision
+		forkTime uint64
+	}{
```
HerbertJordan, line 94: what I ment with using the revision as the key of the map is that you can also remove it from the struct. Why would you list it twice.

All you need is a `map[tosca.Revision]uint64` mapping revisions to fork times.

go/ct/common/revisions_test.go
```diff
@@ -85,3 +86,27 @@ func TestRevision_GetRevisionForBlock(t *testing.T) {
...
+		tosca.R10_London:        {tosca.R10_London, 2000},
+		tosca.R11_Paris:         {tosca.R11_Paris, 3000},
+		tosca.R12_Shanghai:      {tosca.R12_Shanghai, 4000},
+		tosca.R13_Cancun:        {tosca.R13_Cancun, 5000},
+		R99_UnknownNextRevision: {R99_UnknownNextRevision, 6000},
+	}
+
+	for name, test := range tests {
+		t.Run(name.String(), func(t *testing.T) {
+			got := GetForkTime(test.revision)
```
HerbertJordan, line 106: Since we are already here, why is `GetForkTime` not returning a `time.Time`? Fixing this may be a small follow-up cleanup.

HerbertJordan (changes requested):

go/ct/common/revisions_test.go
facuMH: I think uint64 was used because in the reference implementation, uint64 is also used. Maybe it could be turned into `time.Time`, but that would be a separate PR

go/ct/common/bytes_test.go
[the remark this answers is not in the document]
LuisPH3: if it was called, why was coverage not accounting for it?

go/ct/common/cpp_coverage.go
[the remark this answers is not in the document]
LuisPH3: I think this could be solved in a more elegant way by changing the order of the unit test.
```
       // write coverage data into tempDir directory
	tempDir := t.TempDir()
	os.Setenv("GCOV_PREFIX", tempDir)

	// run dump routine
	DumpCppCoverageData()
	
        expectEnabled := *stateImpl
	enabled := isCppCoverageEnabled()

	if !enabled {
		if expectEnabled {
			t.Fatalf("Failed, cpp coverage is not enabled")
		} else {
			t.Skip("Skipping test, cpp coverage disabled and not expected to be enabled")
		}
	} else if !expectEnabled {
		t.Fatalf("Failed, cpp coverage is enabled, but it was expected disabled")
	}
```

With this change, the function is always called, even when side-effect is to be ignored.

go/ct/common/bytes_test.go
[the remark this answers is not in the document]
facuMH: package coverage did not account for it, because this method was not called in any other test in this package. However it is used in many other packages and tests (serialization_test, for example), but since these calls are outside of the package scope they are not considered for package coverage.

go/ct/common/immutable_hash_array.go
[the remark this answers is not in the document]
LuisPH3: Changes in the implementation of the type will lead to runtime errors when unmarshalling. This is a thing that may happen in the future when refactoring/optimizing/adding features, and we will silently drop it. I think we really need to test it. 
We can ignore the error in this location, if we provide guarantees that json.Marshal will not fail for our type. (an explicit test that proves that it cannot return an error)

go/ct/common/immutable_hash_array_test.go
```diff
@@ -107,6 +108,17 @@ func TestImmutableHashArray_CanBeJsonEncoded(t *testing.T) {
...
 	}
+
+	t.Run("short-encoded", func(t *testing.T) {
+		encoded, err := json.Marshal(zeroHash[:len(zeroHash)-1])
+		if err != nil {
+			t.Fatalf("failed to encode into JSON: %v", err)
+		}
+		var restored ImmutableHashArray
+		if err := json.Unmarshal(encoded, &restored); err == nil {
+			t.Fatalf("expected error, got %v", err)
```
LuisPH3, line 119:
```suggestion
			t.Fatalf("expected no error, got %v", err)
```

go/ct/common/cpp_coverage_test.go
```diff
@@ -62,3 +62,8 @@ func TestDumpCppCoverageData(t *testing.T) {
 		t.Fatalf("Failed, test generated no coverage data files")
 	}
 }
+
+func TestDumpCppCoverageData_enabled(t *testing.T) {
```
LuisPH3, line 66: please see my comment. would be better to re-order test

go/ct/common/immutable_hash_array.go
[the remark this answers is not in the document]
facuMH: a test has been added to test this error.

go/ct/common/u256_test.go
HerbertJordan, line 1: If I remember correctly, this file was not part of this PR the last time I reviewed it. You should not keep adding no material to a PR that is under review. This adds confusion and complexity. The next time, please make such unrelated additions a distinct PR or only submit the initial PR once your work on the objective you try to achieve (in this case increase test coverage of the common package) is complete.

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,178 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+	}{
+		{U256{}, U256{}},
+		{NewU256(1), NewU256(1)},
+		{NewU256(0xff), NewU256(0xff)},
+		{NewU256(0xffffffffffffffff), NewU256(0xffffffffffffffff)},
+		{NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff), NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff)},
+	}
+
+	for _, test := range tests {
+		if want, got := test.want, test.value.SignExtend(NewU256(0xfff)); !want.Eq(got) {
```
HerbertJordan, line 551: This test does not cover the case where the argument to `SignExtend` is less than <= 31, in which case the semantic of the function changes.

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,178 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+	tests := []struct {
+		a    U256
+		b    U256
+		want U256
+	}{
+		{U256{}, U256{}, U256{}},
+		{NewU256(1), NewU256(1), NewU256(1)},
+		{NewU256(0xff), NewU256(0xff), NewU256(0xff)},
+		{NewU256(0xffffffffffffffff), NewU256(0xffffffffffffffff), NewU256(0xffffffffffffffff)},
+		{NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff), NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff), NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff)},
```
HerbertJordan, line 567: These are rather pure test inputs for the `And` function. If it would be implemented as a no-op, all these tests would pass.

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,178 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
```
HerbertJordan, line 587: Same as for the `And` function test inputs.

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,178 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+	tests := []struct {
+		a    U256
+		b    U256
+		want U256
+	}{
+		{U256{}, U256{}, U256{}},
+		{NewU256(1), NewU256(1), U256{}},
+		{NewU256(0xff), NewU256(0xff), U256{}},
+		{NewU256(0xffffffffffffffff), NewU256(0xffffffffffffffff), U256{}},
+		{NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff), NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff), U256{}},
```
HerbertJordan, line 607: This test would be passed by an `Xor` that just always resets the content of the value to zero.

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,178 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+	}{
+		{U256{}, U256{}, U256{}},
+		{NewU256(1), NewU256(1), U256{}},
+		{NewU256(0xff), NewU256(0xff), U256{}},
+		{NewU256(0xffffffffffffffff), NewU256(0xffffffffffffffff), U256{}},
+		{
+			NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff),
+			NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff),
+			NewU256(0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff),
+		},
```
HerbertJordan, line 631: Not great test inputs either ...

HerbertJordan (changes requested):
Please take a bit more care when creating tests on what you are testing. The goal should not be to just add tests to make the code coverage happy.

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,213 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+	tests := []struct {
+		a    U256
+		b    U256
+		want U256
+	}{
+		{U256{}, U256{}, U256{}},
+		{NewU256(1), NewU256(1), NewU256(1)},
+		{NewU256(0xff), NewU256(0xff), NewU256(0xff)},
+		{NewU256(maxu64), NewU256(maxu64), NewU256(maxu64)},
+		{NewU256(0x12345678), NewU256(0x87654321), NewU256(0x02244220)},
```
HerbertJordan, line 579: nit: make the intend explicit
```suggestion
		{NewU256(0x12345678), NewU256(0x87654321), NewU256(0x12345678 & 0x87654321)},
```

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,213 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+	tests := []struct {
+		a    U256
+		b    U256
+		want U256
+	}{
+		{U256{}, U256{}, U256{}},
+		{NewU256(1), NewU256(), NewU256(1)},
+		{NewU256(0xff), NewU256(0xff), NewU256(0xff)},
+		{NewU256(maxu64), NewU256(0), NewU256(maxu64)},
+		{NewU256(0x12345678), NewU256(0x87654321), NewU256(0x97755779)},
```
HerbertJordan, line 615:
```suggestion
		{NewU256(0x12345678), NewU256(0x87654321), NewU256(0x12345678 | 0x87654321)},
```

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,213 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+	tests := []struct {
+		a    U256
+		b    U256
+		want U256
+	}{
+		{U256{}, U256{}, U256{}},
+		{NewU256(1), NewU256(0), NewU256(1)},
+		{NewU256(0), NewU256(0xff), NewU256(0xff)},
+		{NewU256(maxu64), NewU256(maxu64), U256{}},
+		{NewU256(0x12345678), NewU256(0x87654321), NewU256(0x95511559)},
```
HerbertJordan, line 641:
```suggestion
		{NewU256(0x12345678), NewU256(0x87654321), NewU256(0x12345678 ^ 0x87654321)},
```

go/ct/common/u256_test.go
```diff
@@ -461,3 +462,213 @@ func TestU256_MarshallingRoundTrip(t *testing.T) {
...
+func TestU256_Srsh(t *testing.T) {
+	tests := []struct {
+		a    U256
+		b    U256
+		want U256
+	}{
+		{NewU256(0x12345678, 0x9abcdef0, 0x12345678, 0x9abcdef0), NewU256(0), NewU256(0x12345678, 0x9abcdef0, 0x12345678, 0x9abcdef0)},
+		{NewU256(0x12345678, 0x9abcdef0, 0x12345678, 0x9abcdef0), NewU256(64), NewU256(0, 0x12345678, 0x9abcdef0, 0x12345678)},
+		{NewU256(0x12345678, 0x9abcdef0, 0x12345678, 0x9abcdef0), NewU256(128), NewU256(0, 0, 0x12345678, 0x9abcdef0)},
+		{NewU256(0x12345678, 0x9abcdef0, 0x12345678, 0x9abcdef0), NewU256(256), NewU256(0, 0, 0, 0)},
```
HerbertJordan, line 665: A non-multiple of 64 like 47 would be nice ... to make sure non-word aligned shifts are also covered.

HerbertJordan (approved):

LuisPH3 (approved):

HerbertJordan (approved):

PR #640, opened 2024-08-08
Improve coverage gen package

facuMH:
This PR increases coverage of gen package.
in particular it adds a test for the string of `Assigment` and then a series of tests checking that calling the different methods that add different type of constraints to the state generator, are indeed adding the constraints.

### DIFF
--- a/go/ct/gen/assignment_test.go
+++ b/go/ct/gen/assignment_test.go
@@ -1,0 +1,46 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package gen
+
+import (
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/ct/common"
+)
+
+func TestAssignment_String(t *testing.T) {
+
+	tests := map[string]struct {
+		value common.U256
+	}{
+		"a": {common.NewU256(1)},
+		"b": {common.NewU256(2, 2)},
+		"c": {common.NewU256(3, 3, 3)},
+		"d": {common.NewU256(4, 4, 4, 4)},
+	}
+
+	for variable, test := range tests {
+		a := make(Assignment)
+		a[Variable(variable)] = test.value
+
+		got := a.String()
+		want := "{" + variable + "->" + test.value.String() + "}"
+
+		if got != want {
+			t.Errorf("Assignment.String() = %v, want %v", got, want)
+		}
+	}
+
+	if a := Assignment(nil); a.String() != "{}" {
+		t.Errorf("Assignment.String() = %v, want {}", a.String())
+	}
+
+}
--- a/go/ct/gen/state_test.go
+++ b/go/ct/gen/state_test.go
@@ -260,7 +260,7 @@ func TestStateGenerator_SetGasRefundIsEnforced(t *testing.T) {
 		generator.SetGasRefund(gasRefund)
 		state, err := generator.Generate(rnd)
 		if err != nil {
-			t.Fatalf("unexpected error during build: %v", err)
+			t.Fatalf("failed to generate state, unexpected error: %v", err)
 		}
 		if want, got := gasRefund, state.GasRefund; want != got {
 			t.Errorf("unexpected amount of gas refund, wanted %d, got %d", want, got)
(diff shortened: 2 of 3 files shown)


go/ct/gen/assignment_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/ct/common"
+)
+
+func TestAssignment_String(t *testing.T) {
+
+	tests := map[string]struct {
+		value common.U256
+		want  string
```
HerbertJordan, line 23: nit: since this test does not need to test the `String` function of U256 as well, all you would need  is the `value`. The `want` field in the struct can be derived int the test loop by calling `String()` on the value.

HerbertJordan (approved):

simonlechner (dismissed):

go/ct/gen/state_test.go
```diff
@@ -234,6 +234,20 @@ func TestStateGenerator_NonConflictingGasAreAccepted(t *testing.T) {
...
 }
 
+func TestStateGenerator_AddGasLowerUpperBoundIsEnforced(t *testing.T) {
+	generator := NewStateGenerator()
+	generator.AddGasLowerBound(42)
+	generator.AddGasUpperBound(44)
+
+	state, err := generator.Generate(rand.New(0))
+	if err != nil {
+		t.Fatalf("unexpected error during build: %v", err)
```
LuisPH3, line 244: for the future:
error strings are better written "failed to do something, the reason is...." 
This case: `failed to generate state, unexpected error ....`

go/ct/gen/state_test.go
```diff
@@ -478,3 +506,149 @@ func TestStateGenerator_BlockNumberHashes(t *testing.T) {
...
+
+// //////////////////////////////////////////////////////////
+// blob hash index
+func TestStateGenerator_IsAbsentBlobHashIndex(t *testing.T) {
+	generator := NewStateGenerator()
+	generator.IsAbsentBlobHashIndex(Variable("a"))
+	if generator.transactionContextGen.blobHashVariables[Variable("a")] {
+		t.Errorf("unexpected transaction context configuration")
+	}
+}
```
LuisPH3, line 654: After reading all these generators, I have two suggestions:
- It may be possible to factor all of them in one or two groups, with table tests. 
- I am not very familiar with the generation process itself, but I think it may be possible to add constraints and generate a state. Where the constraints can be observed. I am a little surprised that that kind of test does not exist already, and if it does, it could be easy to upgrade it to cover these missing parts. 
   -  something like [this](https://github.com/Fantom-foundation/Tosca/blob/2972bf5e4475318ed368c39493a2e80befbe0d1f/go/ct/gen/code_test.go#L118)  or  [this](https://github.com/Fantom-foundation/Tosca/blob/2972bf5e4475318ed368c39493a2e80befbe0d1f/go/ct/gen/transaction_context_test.go#L98)

go/ct/gen/state_test.go
```diff
@@ -478,3 +506,149 @@ func TestStateGenerator_BlockNumberHashes(t *testing.T) {
...
 }
+
+// //////////////////////////////////////////////////////////
+// code constraints
+func TestStateGenerator_AddCodeOperation(t *testing.T) {
+	generator := NewStateGenerator()
+	var1 := Variable("a")
+	generator.AddCodeOperation(var1, vm.ADD)
+	varOpConstraint := varOpConstraint{var1, vm.ADD}
+	if len(generator.codeGen.varOps) != 1 || generator.codeGen.varOps[0] != varOpConstraint {
```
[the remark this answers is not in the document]
facuMH: I wanted to keep the tests as minimal and unitarian as possible, since Generate is such a complex function and it has a lot of other tests addressing it, I thought here aiming for just the explicit effect of the function being tested is enough.

go/ct/gen/state_test.go
facuMH: done

go/ct/gen/state_test.go
```diff
@@ -478,3 +506,110 @@ func TestStateGenerator_BlockNumberHashes(t *testing.T) {
 		newHashes = append(newHashes, hashi)
 	}
 }
+
+func TestStateGenerator_VariousContraints(t *testing.T) {
```
HerbertJordan, line 510: These tests seem to focus on the effect of the functions registering conditions on the generator -- which is, on its own fine. What I am a bit surprised / worried is that these functions have not been called as part of the test making sure that those constraints can be satisfied. Are there no such tests?

HerbertJordan (approved):

go/ct/gen/state_test.go
facuMH: there are such a tests, like the one in line 296, but since some conditions clash, not all of them are called. I could modify this test to generate a state and check that the condition is also met afterwards, but I feel that would also be testing the generate function, not really the constraining functions

LuisPH3 (approved):
This mentioned before: I am shocked about how much code did not have a test.

simonlechner (approved):

PR #641, opened 2024-08-09
Reorder makefile commands

facuMH:
After merging the report coverage we observed that the tests were running 2 times for each commit and there was also a problem with the coverage report trying to publish in a PR when a commit was done into main (which has no PR). Also keeping into consideration that when developing it's easier to run just one command to get the report and not having to mentally record whether the `cover.out` is updated or not, we came up with this solution:
- for development one runs `make test-coverage` which will run the tests and then report the coverage on the updated code.
- for commits in PRs (aka branches that are not main) this is separated into 2 steps, `make tests-go` and `make coverage-report`
- for commits in main, this will only run `make tests-go`. 

Hence preventing commits in PRs to run tests twice while still facilitating a single command to run both things locally

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -56,19 +56,13 @@ jobs:
     - name: Test
       run: make test-go
 
-    - name: CT regression tests LFVM
-      run: go run ./go/ct/driver regressions lfvm
-
-    - name: CT regression tests evmzero
-      run: go run ./go/ct/driver regressions evmzero
-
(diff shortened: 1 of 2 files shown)


LuisPH3 (approved):
This Change avoids execution of tests in CI twice, improving turnaround time.

HerbertJordan (approved):
Nice improvement, thanks!

simonlechner (approved):

PR #642, opened 2024-08-12
Implement create calls

simonlechner:
This PR adds support for CREATE and CREATE2 instructions to Floria. To ensure the right address is calculated another integration test is added, stressing this calculation.

### DIFF
--- a/go/integration_test/processor/create_test.go
+++ b/go/integration_test/processor/create_test.go
@@ -15,7 +15,6 @@ import (
 	"fmt"
 	"math/big"
 	"slices"
-	"strings"
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
@@ -29,9 +28,6 @@ func TestProcessor_CreateAndCallContract(t *testing.T) {
 	gasPush := vm.PUSH1 + vm.OpCode(len(gasLimit.Bytes())-1)
 
 	for processorName, processor := range getProcessors() {
-		if strings.Contains(processorName, "floria") {
-			continue // todo implement different create types
-		}
 		for _, create := range []vm.OpCode{vm.CREATE, vm.CREATE2} {
 			t.Run(fmt.Sprintf("%s-%s", processorName, create.String()), func(t *testing.T) {
 				sender0 := tosca.Address{1}
@@ -153,9 +149,6 @@ func TestProcessor_CreateInitCodeIsExecutedInRightContext(t *testing.T) {
 	gasPush := vm.PUSH1 + vm.OpCode(len(gasLimit.Bytes())-1)
 
 	for processorName, processor := range getProcessors() {
-		if strings.Contains(processorName, "floria") {
-			continue // todo implement different create types
-		}
 		for _, create := range []vm.OpCode{vm.CREATE, vm.CREATE2} {
 			t.Run(fmt.Sprintf("%s-%s", processorName, create.String()), func(t *testing.T) {
 				sender0 := tosca.Address{1}
@@ -294,14 +287,10 @@ func TestProcessor_CreateInitCodeIsExecutedInRightContext(t *testing.T) {
 }
 
 func TestProcessor_EmptyReceiverCreatesAccount(t *testing.T) {
-	addressToBeCreated := tosca.Address(crypto.CreateAddress(common.Address{1}, 0))
-	for processorName, processor := range getProcessors() {
-		if strings.Contains(processorName, "floria") {
-			continue // todo implement different create types
-		}
-
-		sender0 := tosca.Address{1}
+	for _, processor := range getProcessors() {
 		checkValue := byte(42)
+		sender := tosca.Address{1}
+		addressToBeCreated := tosca.Address(crypto.CreateAddress(common.Address(sender), 0))
 
 		initCode := []byte{
 			byte(vm.PUSH1), checkValue,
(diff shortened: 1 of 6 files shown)


go/processor/floria/run_context.go
```diff
@@ -73,14 +95,36 @@ func (r runContext) Call(kind tosca.CallKind, parameters tosca.CallParameters) (
 		r.RestoreSnapshot(snapshot)
 	}
 
+	if kind == tosca.Create || kind == tosca.Create2 {
+		r.SetCode(createdAddress, tosca.Code(result.Output))
```
HerbertJordan, line 99: This should probably be skipped (or reverted by the snapshot above) if the execution produced an error, right?

go/processor/floria/run_context.go
```diff
@@ -73,14 +95,36 @@ func (r runContext) Call(kind tosca.CallKind, parameters tosca.CallParameters) (
...
-		Success:   result.Success,
+		Output:         result.Output,
+		GasLeft:        result.GasLeft,
+		GasRefund:      result.GasRefund,
+		Success:        result.Success,
+		CreatedAddress: createdAddress,
 	}, err
 }
 
+func hashFromCode(code tosca.Code) tosca.Hash {
```
HerbertJordan, line 111: just `hashCode` maybe? `hash` is also a verb ... I think ...

go/processor/floria/run_context.go
```diff
@@ -26,42 +30,60 @@ type runContext struct {
...
 		code = r.GetCode(parameters.CodeAddress)
 		codeHash = r.GetCodeHash(parameters.CodeAddress)
 	}
 
+	recipient := parameters.Recipient
+	var createdAddress tosca.Address
+	if kind == tosca.Create || kind == tosca.Create2 {
+		if parameters.Recipient == (tosca.Address{}) {
+			code = tosca.Code(parameters.Input)
+			codeHash = hashFromCode(code)
```
HerbertJordan, line 52: I think in the interpreter implementations we accept `nil` pointer for code hashes in such cases. This should prevent the need to hash and cache codes that are only used once.

We can keep the hashing here now, it should not be a problem. But maybe, if we see performance issues with hashing of codes, this may be something to be removed.

HerbertJordan (dismissed):
Looks good. How is the code coverage of the unit tests?

facuMH (dismissed):
I am not super happy about introducing dependencies to geth in floria, but I acknowledge that when it comes to cryptographic libraries, it's usually better to use an existing validated library, rather than making your own

go/processor/floria/run_context.go
simonlechner: I will focus on correctness and keeping it as simple as possible first. We can also profile and tune the implementation when we are satisfied with the correctness.

go/processor/floria/run_context.go
```diff
@@ -26,42 +30,60 @@ type runContext struct {
...
+		r.SetNonce(parameters.Sender, r.GetNonce(parameters.Sender)+1)
+		r.SetNonce(createdAddress, 1)
+		recipient = createdAddress
+	}
+
 	if kind == tosca.StaticCall {
 		r.static = true
 	}
 
+	snapshot := r.CreateSnapshot()
```
LuisPH3, line 71: It makes sense to me to move the "undo" functionality as close as possible to the run call.
Would it make sense to use functions to scope the region of the code that it is transactional? 

```go
func (r runcontext)runTransaction(....)  error
     snapshot := r.CreateSnapshot()
     var err error
     defer func () {
         if err != nil{
             r.RestoreSnapshot(snapshot)
          }
     }()

    err = transferValue(....)
    if err != nil{
        return err
    }
    
    result, err := r.interpreter.Run(interpreterParameters)
	if err != nil || !result.Success {
	     return fmt.Errorf("failed blah blah blah")
	}
```

This code could even have an evolution where you have something like
```go
func (r runcontext) runTransaction (parts ...func ()error) error
``` 
which will automatically revert the state if any of the lambdas returns an error.

The reason for all these ideas is that call is rather large and accumulates quite some logic. The transactional behavior could be abstracted into some concept that reads more like: if this fails, the state is reverted. Instead of having to read through (specially when you begin to have multiple restore snapshot down the line. 

Please consider this in future changes.

LuisPH3 (approved):

facuMH (approved):

HerbertJordan (approved):

go/processor/floria/run_context.go
simonlechner: I will keep this idea in my head but I want to focus on a simple implementation that is hopefully tested well enough to be correct. Once all the functionality is there, I think there will be some restructuring of the call.

go/integration_test/processor/create_test.go
```diff
@@ -336,7 +324,131 @@ func TestProcessor_EmptyReceiverCreatesAccount(t *testing.T) {
...
+				salt := append(bytes.Repeat([]byte{0}, 31), saltByte)
+				wantAddress := tosca.Address(crypto.CreateAddress(common.Address(receiver), state[receiver].Nonce))
+				if create == vm.CREATE2 {
+					wantAddress = tosca.Address(crypto.CreateAddress2(common.Address(receiver), [32]byte(salt), codeHash[:]))
+				}
+
+				// Run the processor
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil || !result.Success {
+					t.Errorf("execution was not successful or failed with error %v", err)
```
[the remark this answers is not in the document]
simonlechner: That is why there is an "or" in the error message, either it was unsuccessful or it failed with the following error.

PR #643, opened 2024-08-12
improve rlz coverage 

facuMH:
This PR adds UT to improve the coverage of the `rlz` package and removes it from the `override` config.

Some test have been deleted and refactored  together to make the most out of table testing.

Noteworthy comment:
coverage is not 100% because there are a couple of checks that can never fail but I think the should still not be removed, such as the inequalities in confitions_tests, or the `Op.Eval`. In both these cases the corresponding checks can not fail because the current implementations do no generate errors or allow values that could lead to the fail of those checks.

### DIFF
--- a/go/ct/rlz/domains_test.go
+++ b/go/ct/rlz/domains_test.go
@@ -25,8 +25,8 @@ func TestRemoveDuplicatesGeneric(t *testing.T) {
 
 	tests := map[string]struct {
 		generic_type reflect.Type
-		input        interface{}
-		expected     interface{}
+		input        any
+		expected     any
 	}{
 		"empty": {
 			generic_type: reflect.TypeOf(1),
@@ -68,7 +68,7 @@ func TestRemoveDuplicatesGeneric(t *testing.T) {
 
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			var result interface{}
+			var result any
 			switch tc.generic_type {
 			case reflect.TypeOf(1):
 				result = removeDuplicatesGeneric[int](tc.input.([]int))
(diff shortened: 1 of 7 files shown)


go/ct/rlz/domains_test.go
```diff
@@ -79,3 +85,296 @@ func TestRemoveDuplicatesGeneric(t *testing.T) {
...
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			defer func() {
+				if r := recover(); r == nil {
+					t.Errorf("The code did not panic")
+				}
+			}()
+			switch tc.generic_type {
```
HerbertJordan, line 244: For situations like this, Go has a [type-switch](https://go.dev/tour/methods/16) statement. With this, this test case could be shortened to:
```

func TestDomain_PredecessorPanic(t *testing.T) {
	tests := map[string]any{
		"bool":       true,
		"statusCode": st.Running,
		"opCode":     vm.ADD,
	}

	for name, value := range tests {
			defer func() {
				if r := recover(); r == nil {
					t.Errorf("The code did not panic")
				}
			}()
			switch value := value.(type) {
			case bool:
				boolDomain{}.Predecessor(value)
			case st.StatusCode:
				statusCodeDomain{}.Predecessor(value)
			case vm.OpCode:
				opCodeDomain{}.Predecessor(value)
			default:
				t.Errorf("Add type to test cases: %v", reflect.TypeOf(value))
			}
		})
	}
}
```

go/ct/rlz/domains_test.go
```diff
@@ -79,3 +85,296 @@ func TestRemoveDuplicatesGeneric(t *testing.T) {
...
+			}
+		})
+	}
+
+}
+
+func TestDomain_Predecessor(t *testing.T) {
+
+	tests := map[string]struct {
+		got  interface{}
```
HerbertJordan, line 195: nit: since Go 1.18 the keyword `any` can be used instead of `interface{}`.

HerbertJordan (dismissed):
Thanks for extending the test coverage. Just a few suggestions to be considered.

In the bigger picture, I think the transition from tests per expression or domain type to one test for each operation on the domains and then testing all domains in those tests is a bit against the open/closed principle. Adding a domain, for instances, requires the modification of existing tests. Deleting it as well.

If you encounter such a situation in the future, we should talk about better ways to structure the tests. In particular, one could set up generic tests for an abstract domain, covering all domain relevant tests, and then instantiate those once for each domain.

LuisPH3 (dismissed):

simonlechner (dismissed):

LuisPH3 (approved):

HerbertJordan (approved):

simonlechner (approved):

PR #646, opened 2024-08-14
remove github wokflows that comments on PRs

facuMH:
The aim of this PR is to make it so that we only a get a msg in a PR if the test has failed, but if the test succeeded, then no comment is needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,4 +121,4 @@ test-coverage: test-go coverage-report
 
 coverage-report:
 	@go install github.com/vladopajic/go-test-coverage/v2@v2.10.1
-	@go-test-coverage --config .testcoverage.yml
+	@go-test-coverage --config .testcoverage.yml && ([ $$? -eq 0 ] && exit 0) || exit 0
--- a/Makefile
+++ b/Makefile
@@ -121,4 +121,4 @@ test-coverage: test-go coverage-report
 
 coverage-report:
 	@go install github.com/vladopajic/go-test-coverage/v2@v2.10.1
-	@go-test-coverage --config .testcoverage.yml && ([ $$? -eq 0 ] && exit 0) || exit 0
+	@go-test-coverage --config .testcoverage.yml
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -63,12 +63,17 @@ jobs:
        script: |
           const execSync = require('child_process').execSync;
           const coverageResult = execSync('make coverage-report', { encoding: 'utf-8' });
-          github.rest.issues.createComment({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            body: `Test results:\n\n${coverageResult}`
-          });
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -121,4 +121,4 @@ test-coverage: test-go coverage-report
 
 coverage-report:
 	@go install github.com/vladopajic/go-test-coverage/v2@v2.10.1
-	@go-test-coverage --config .testcoverage.yml
\ No newline at end of file
+	@go-test-coverage --config .testcoverage.yml && ([ $$? -eq 0 ] && exit 0) || exit 0
```
LuisPH3, line 124: I don't think this is a good idea, you are polluting the makefile with side effects to make github actions happy. This line returning a non 0 value is the expected behavior of any faulty test, any other CI script consuming this command would expect so.

LuisPH3 (changes requested):
I am sorry, but I don't think this is a good solution. 
This PR shifts the responsibility of a failing command from returning a non 0 value into grepping for a string in the output. The ad-hoc behavior is not intuitive for any other use of the makefile target. 
I believe that github actions have better methods to trigger action after failute than to hack it together. https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f

Makefile
HerbertJordan: I totally agree with Luis. This is not a good solution.

HerbertJordan (changes requested):

Makefile
facuMH: The current problem is that we need to do some exit code handling in `go.yml`.
I will elaborate. We need to:
- run `make coverage-report`, and we want the exit code of this command to dictate whether the step failed or not
- we want to save the output of the previous command to an env var or file, so that the next step can comment it
- we want to comment the output of previous step but only if it failed (this is doable with `if:failure()`)

The problem is in the first step, which should do the first 2 actions, run the comand, save the output, and report whether the command failed or not. But if the make command failed, then it will not execute the following command (which saves the output into an env var or file.

Maybe there is a more elegant manner of doing this, but (as far as I know) this can be done saving the exit code of the first command, and using it after the output text was saved. 

However, this requires more research into github workflows and time I think now should be spent in other tasks, so I will remove it all, just leave the basic, it fails if command fails, pass if success, and we will have to look into the build log to see what went wrong. 

Maybe in the future we can resume this research task and find a more elegant way of doing this.

LuisPH3: I am afraid that the Makefile is more important than the CI scripts: The expected behavior of a failing step is to return non 0. This is very important because any scripting is built upon this principle. 
The problem you are experiencing is that you need to make actions on failure. This needs to be a solution provided in the CI script and not in the makefile. 

I am afraid that if we can not solve this now, we need to give up on posting messages until we find the time to learn how to do it properly.

LuisPH3:
Please rename tittle of the PR for future reference. and I guess we are ready to go

LuisPH3 (approved):

HerbertJordan (approved):

PR #647, opened 2024-08-14
Add Rust EVM interpreter (evmrs)

LorenzSchueler:
emvrs passes CT (`full-mode` except for `pc_on_data_is_ignored`) and all integrations tests.


LorenzSchueler:
@LuisPH3 could you have a quick look at this please?

go/interpreter/evmrs/evmrs.go
```diff
@@ -0,0 +1,117 @@
...
+	if params.Revision > newestSupportedRevision {
+		return tosca.Result{}, &tosca.ErrUnsupportedRevision{Revision: params.Revision}
+	}
+	return e.e.Run(params)
+}
+
+// evmrsInstanceWithProfiler implements the tosca.ProfilingVM interface and is used for all
+// configurations collecting profiling data.
+type evmrsInstanceWithProfiler struct {
+	*evmrsInstance
```
LuisPH3, line 116: have you tested this?

rust/src/types/u256_mod.rs
LuisPH3, line 1: This file is a good candidate to start breaking down the PR.
We can start by adding build environment and this as an initial rust code.  Then we build on top of it.

This file needs a friend: unit testing.

rust/src/types/u256_mod.rs
```diff
@@ -0,0 +1,516 @@
...
+}
+
+impl DerefMut for u256 {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0.bytes
+    }
+}
+
+impl Debug for u256 {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
LuisPH3, line 37: This seems pretty readable to me, why debug? Let's make it a real formatter.

rust/src/types/u256_mod.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+}
+
+impl From<u256> for I256 {
+    fn from(value: u256) -> Self {
+        BInt::from_be_slice(value.deref()).unwrap()
+    }
+}
+
+impl From<U512> for u256 {
```
LuisPH3, line 106: Not sure if we need this at all. I think we can drop 512 and 128 for the time being.

rust/src/types/u256_mod.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+}
+
+impl From<u256> for U256 {
+    fn from(value: u256) -> Self {
+        BUint::from_be_slice(value.deref()).unwrap()
+    }
+}
+
+impl From<I256> for u256 {
```
LuisPH3, line 86: I don't think we need signed conversions, neither.

LorenzSchueler: this is for the signed variants (SDIV, SREM, SLT, SGT)

rust/src/types/u256_mod.rs
LorenzSchueler: this is for addmod and mulmod where the intermediate value (sum or product) may not fit into 256 bits

PR #648, opened 2024-08-16
Fix new linter findings

facuMH:
since we are installing the static check module with `@latest` whenever they release a new version we might get new findings. 
This is party good because we can keep improving our code quality, but it means that occasionally PRs will fail for reasons unrelated its changes.

### DIFF
--- a/go/ct/driver/probe.go
+++ b/go/ct/driver/probe.go
@@ -11,6 +11,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"runtime"
 	"sync"
@@ -193,7 +194,7 @@ func testState(specification spc.Specification, state *st.State, evm ct.Evm) err
 	defer result.Release()
 
 	if !expected.Eq(result) {
-		return fmt.Errorf(formatDiffForUser(state, result, expected, rules[0].Name))
+		return errors.New(formatDiffForUser(state, result, expected, rules[0].Name))
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


go/ct/driver/probe.go
```diff
@@ -193,7 +193,7 @@ func testState(specification spc.Specification, state *st.State, evm ct.Evm) err
 	defer result.Release()
 
 	if !expected.Eq(result) {
-		return fmt.Errorf(formatDiffForUser(state, result, expected, rules[0].Name))
+		return fmt.Errorf("%v", formatDiffForUser(state, result, expected, rules[0].Name))
```
simonlechner, line 196:
```suggestion
		return fmt.Error(formatDiffForUser(state, result, expected, rules[0].Name))
```
I think the linter does not suggested adding `%v` for default formatting but rather calling the error function which does not support formatting. 
Also for the other occurrences.

simonlechner (changes requested):

go/ct/driver/probe.go
facuMH: well, the linter is saying that if we call `Errorf` then we have to use `Errorf("%someverb", somevalue)` and if we call `Error` then we use `Error(somevalue)`. both ways fix the problem. But yes, in this case I think we can also just call `Error`

LuisPH3: would it work with %w ?  for other errors with format?

facuMH: `fmt.Error` does not existe, instead using `errors.New`

LuisPH3 (approved):

simonlechner (approved):

HerbertJordan (approved):

PR #649, opened 2024-08-16
Fix the expected stack element order in steppable evmc interface

LorenzSchueler:
Currently, the (prepared) stack that is passed from Go to C++/Rust in the `StepN` interface of evmc has the lowest element of the stack in the first position of the array and the top of the stack in the last position. However, the stack that C++/Rust returns to Go is expected to have the top of the stack in the first position, which is unintuitive. This PR aligns the behavior so that the top of the stack is always last.


LuisPH3 (approved):

facuMH (approved):

HerbertJordan (approved):
Thanks for fixing this!

LorenzSchueler:
closing in favor of #653

PR #656, opened 2024-08-20
Add evmrs stack

LorenzSchueler:
- add `Stack` type

### DIFF
--- a/rust/src/interpreter/stack.rs
+++ b/rust/src/interpreter/stack.rs
@@ -1,0 +1,116 @@
+use evmc_vm::StatusCode;
+
+use crate::types::u256;
+
+#[derive(Debug)]
+pub struct Stack(Vec<u256>);
+
+impl Stack {
+    pub fn new() -> Self {
+        Self(Vec::new())
+    }
+
+    pub fn from_vec(inner: Vec<u256>) -> Self {
+        Self(inner)
+    }
+
+    pub fn into_inner(self) -> Vec<u256> {
+        self.0
+    }
+
+    pub fn push(&mut self, value: impl Into<u256>) -> Result<(), StatusCode> {
+        self.check_overflow_on_push()?;
+        self.0.push(value.into());
+        Ok(())
+    }
+
+    pub fn swap_with_top(&mut self, nth: usize) -> Result<(), StatusCode> {
+        self.check_underflow(nth + 1)?;
+
+        let len = self.0.len();
+        self.0.swap(len - 1, len - 1 - nth);
+        Ok(())
+    }
+
+    pub fn pop<const N: usize>(&mut self) -> Result<[u256; N], StatusCode> {
+        self.check_underflow(N)?;
+
+        let mut array = [u256::ZERO; N];
+        for element in &mut array {
+            *element = self.0.pop().unwrap();
+        }
+        Ok(array)
+    }
+
+    pub fn nth(&self, nth: usize) -> Result<u256, StatusCode> {
+        self.check_underflow(nth + 1)?;
+        Ok(self.0[self.0.len() - nth - 1])
+    }
+
+    #[inline(always)]
+    fn check_overflow_on_push(&self) -> Result<(), StatusCode> {
+        if self.0.len() >= 1024 {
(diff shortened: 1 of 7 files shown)


rust/src/interpreter/stack.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    pub fn empty() -> Self {
+        Self(Vec::new())
+    }
+
+    pub fn into_inner(self) -> Vec<u256> {
+        self.0
+    }
+
+    pub fn push(&mut self, value: impl Into<u256>) -> Result<(), (StepStatusCode, StatusCode)> {
```
LuisPH3, line 21: The Result type is rather verbose, consider introducing an error type.

rust/src/interpreter/stack.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    pub fn swap_with_top(&mut self, nth: usize) -> Result<(), (StepStatusCode, StatusCode)> {
+        self.check_underflow(nth + 1)?;
+
+        let len = self.0.len();
+        self.0.swap(len - 1, len - 1 - nth);
+        Ok(())
+    }
+
+    pub fn pop<const N: usize>(&mut self) -> Result<[u256; N], (StepStatusCode, StatusCode)> {
```
LuisPH3, line 35: Integral generics is a new feature I never got to use before. Happy that it is here now.

rust/src/types/amount.rs
```diff
@@ -157,6 +170,30 @@ impl From<usize> for u256 {
     }
 }
 
+impl From<Address> for u256 {
```
LuisPH3, line 173: I pressure that these conversions simplify quite a lot popping/pushing addresses in the stack. 
I wonder if the Address type should be a Cow, to avoid copying in read-only situations. You can add this to the list to investigate in the future.

rust/src/interpreter/stack.rs
```diff
@@ -0,0 +1,149 @@
...
+
+#[derive(Debug)]
+pub struct Stack(Vec<u256>);
+
+impl Stack {
+    pub fn from(inner: Vec<u256>) -> Self {
+        Self(inner)
+    }
+
+    pub fn empty() -> Self {
```
LuisPH3, line 13: nit: call it new, make it first function in list.

rust/src/interpreter/stack.rs
```diff
@@ -0,0 +1,149 @@
+use evmc_vm::{StatusCode, StepStatusCode};
+
+use crate::types::u256;
+
+#[derive(Debug)]
+pub struct Stack(Vec<u256>);
+
+impl Stack {
+    pub fn from(inner: Vec<u256>) -> Self {
```
LuisPH3, line 9: nit: call it newFrom, or newWith. Just for coherence with the rust style.

LuisPH3 (approved):

rust/src/types/amount.rs
LorenzSchueler: `Address` is almost always used as an argument for db functions so the conversion needs to happen anyway.

rust/src/interpreter/stack.rs
LorenzSchueler: Agreed.

rust/src/interpreter/stack.rs
LorenzSchueler: Agreed.

rust/src/interpreter/stack.rs
LorenzSchueler: Agreed.

rust/src/types/mod.rs
```diff
@@ -1 +1,3 @@
 mod amount;
+
+pub use amount::u256;
```
LuisPH3, line 3: It does not make much sense neither than the namespace is called amount and the type u256. This comes from my own misunderstanding. please rename this module whenever you find time, to something that makes more sense, like u256.

LuisPH3 (approved):

HerbertJordan (approved):

PR #657, opened 2024-08-20
Evmrs opcodes and codestate

LorenzSchueler:
This PR adds `Opcode` as well as `CodeState`. 
`CodeState` holds the code, pc and jump analysis results and makes sure all accesses are correct.


rust/src/interpreter/code_state.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        bytes.into()
+    }
+
+    pub fn pc(&self) -> usize {
+        self.pc
+    }
+}
+
+fn code_byte_types(code: &[u8]) -> Box<[CodeByteType]> {
```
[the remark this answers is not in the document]
LuisPH3: compute_opcode_byte_types, 
process_code_byte_types,

(can you check how is this called in evmzero? )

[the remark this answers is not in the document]
LorenzSchueler: Done

rust/src/interpreter/code_state.rs
```diff
@@ -0,0 +1,271 @@
...
+            Err((
+                StepStatusCode::EVMC_STEP_FAILED,
+                StatusCode::EVMC_BAD_JUMP_DESTINATION,
+            ))
+        );
+    }
+
+    #[test]
+    fn code_state_get_push_data() {
+        let mut code_state = CodeState::new(&[Opcode::SelfDestruct as u8; 32], 0);
```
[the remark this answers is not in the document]
LorenzSchueler: Done

rust/src/interpreter/code_state.rs
```diff
@@ -0,0 +1,271 @@
...
+            };
+            Ok(op)
+        }
+    }
+
+    pub fn next(&mut self) {
+        self.pc += 1;
+    }
+
+    pub fn try_jump(&mut self, dest: u256) -> Result<(), (StepStatusCode, StatusCode)> {
```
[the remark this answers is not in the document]
LorenzSchueler: Done

rust/src/interpreter/code_state.rs
```diff
@@ -0,0 +1,265 @@
...
+        } else if self.code_byte_type[self.pc] == CodeByteType::DataOrInvalid {
+            Err(GetOpcodeError::Invalid)
+        } else {
+            let op = self.code[self.pc];
+            let op = unsafe {
+                // SAFETY:
+                // [Opcode] has repr(u8) and therefore the same memory layout as u8.
+                // In get_code_byte_types this byte of the code was determined to be a valid opcode.
+                // Therefore the value is a valid [Opcode].
+                mem::transmute::<u8, Opcode>(op)
```
HerbertJordan, line 49: Curiosity Question: is this for efficiency because it is faster than implementing a `From<u8>` on the Opcode or is it mainly for simplicity to not have to implement the `From<u8>`?

I guess the "unsafe" part of it is if there is a `u8` value read for which there would actually not be an Opcode enum value ...

rust/src/types/opcode.rs
```diff
@@ -0,0 +1,364 @@
+#![allow(dead_code)]
+
+const STOP: u8 = 0x00;
```
HerbertJordan, line 3: Also a curiosity question: why are there `u8` constants for the opcodes as well as the unit-only enum constants of type `Opcode`? Is this to ease the definition of test cases because with the Opcodes-only you would have to write

```
[Opcode:ADD as u8, Opcode:PUSH1 as u8, 4, Opcode:STOP as u8]
```
and with these constants you can write
```
[ADD, PUSH1, 4, STOP]
```
or is there another use case for this?

Maybe the latter is also not possible, because it does not seem to be used in test cases ...

LorenzSchueler: Ideally I would like to have only the `Opcode` enum and convert to `u8` when needed. The problem is that in the match in `code_byte_type` I have to match against `u8` but type casts are not possible in match arms. The tests in the `code_state` module use `Opcode` instead of the consts because the consts are private to the `opcode` module whereas `Opcode` is public. I want to keep the consts private because that makes it easier to get rid if them at a later point if possible.

rust/src/interpreter/code_state.rs
LorenzSchueler: There is no safe way to convert `u8` to `Opcode` except for matching for each opcode again. However this check was already done when doing the jump analysis.

rust/src/interpreter/code_state.rs
```diff
@@ -0,0 +1,265 @@
+use std::{cmp::min, mem, ops::Deref};
+
+use evmc_vm::StatusCode;
+
+use crate::types::{code_byte_type, u256, CodeByteType, Opcode};
+
+#[derive(Debug)]
+pub struct CodeState<'a> {
```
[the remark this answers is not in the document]
LorenzSchueler: I like `CodeReader`. Will change this in a refactor PR after the other parts are merged.

rust/src/types/opcode.rs
LorenzSchueler: BTW
I just tried to figure out if it is possible to avoid the consts completely.
```rs
match code_byte {
    ...
    Opcode::Push9 as u8 => (CodeByteType::Opcode, 9), // preferred but not possible
    v if v == Opcode::Push9 as u8 => (CodeByteType::Opcode, 9), // possible but ugly
    const { Opcode::Push9 as u8 } => (CodeByteType::Opcode, 9), // possible on nightly
    ...
}
```

rust/src/interpreter/code_state.rs
LuisPH3: I guess the question was more about implementing `From<u8>`
I guess the decision of not doing so is that such a from function will always succeed, even with invalid input, leading to undefined behaviour. 
On the other hand, this specific location seems to assume that this is a valid opcode value, and therefore forces the cast. 

Another alternative would be to implement TryFrom<u8> which would return an error if not possible. Such implementation would have the overhead of testing all valid codes each time. 

I think it is rather verbose, but unsafe block may be the right choice.

PR #659, opened 2024-08-21
Improve coverage lfvm PART1

facuMH:
This PR aims to improve the coverage of the LFVM package, thought does not complete this tasks, I think this is a good point for an intermediate PR and have multiple smaller PRs rather than one PR with an unreadable amount of tests.

This PR: 
- removes `converter.go`, `hash_cache.go`, `instruction.go` and `interpreter.go` from the list of override, since those files are now over 90% covered by unit tests.
- upgrades the threshold for `lfvm` package up to 80%
- removes returned error from `Convert` function, since it can no longer return errors.
- adds hash cache tests
- adds super instrusctions convertion tests
- moves the `switch ctxt.status` from the end of `Run` function to its own for better testing.
- adds tests to check printed log.

- adds `specification.go` file to exception list, since those are covered by the CT run


go/interpreter/lfvm/converter.go
```diff
@@ -100,16 +98,13 @@ func (b *codeBuilder) toCode() Code {
...
 }
 
-func convert(code []byte, with_super_instructions bool) (Code, error) {
+func convert(code []byte, with_super_instructions bool) Code {
 	res := newCodeBuilder(len(code))
 
 	// Convert each individual instruction.
 	for i := 0; i < len(code); {
 		// Handle jump destinations
 		if code[i] == byte(vm.JUMPDEST) {
```
LuisPH3, line 107: This change removes the check, but since length returns nextPos, nextPos is monotonic, and the relationship between nextPos and i depend on the code structure; I guess that the error can still happen given a malformed program as input. 

In this case, the virtual machine may even panic if nextPos grows > len(code)`.

I don't think the vm should ever crash, even with invalid/bogus/malformed/corrupted/malicious code. Can we guarantee this after removing this check?

LuisPH3 (changes requested):
I am afraid this PR is too complicated. 
Can you please break it down: One pr for the converter error removal and other(s) for the rest.

Can you motivate the removal of the error ? I am concerned that this may hide a silent panic when processing malformed code.

facuMH:
As @LuisPH3 mentioned, this PR has too many changes, I will split it to different smaller PRs.

PR #661, opened 2024-08-22
Add Memory and helper functions to deal with gas and slices

LorenzSchueler:


rust/src/interpreter/utils.rs
```diff
@@ -0,0 +1,127 @@
+use std::cmp::min;
+
+use evmc_vm::{ExecutionMessage, MessageFlags, Revision, StatusCode};
+
+use crate::{interpreter::gas::consume_copy_cost, types::u256};
+
+#[inline(always)]
+pub fn get_mut_slice_within_bounds<T>(data: &[T], offset: u256, len: u64) -> &[T] {
```
LuisPH3, line 8: Is the name correct?  I don't see the mut part. 

Shall this be renamed to something like: subslice? span? 
Otherwise: It could be a trait implement for &[T], that could facilitate naming as calls will be using the dot notation. 
i.e. ` sliceXXX.get_in_bounds(uin256, u64)`

rust/src/interpreter/utils.rs
```diff
@@ -0,0 +1,127 @@
...
+    let (end, end_overflow) = offset.overflowing_add(len);
+    if end_overflow || offset >= data.len() {
+        &[]
+    } else {
+        &data[offset..min(end, data.len())]
+    }
+}
+
+#[inline(always)]
+pub fn zero_slice(data: &mut [u8]) {
```
LuisPH3, line 27:
```suggestion
pub fn set_to_zero(data: &mut [u8]) {
```
Functions should have a verb, a noun would make one think that this function returns the zero value. 
extra mile: could be a trait. 
(exceptions in the language are things like from, into, new.... but those have well defined semantics)

rust/src/interpreter/utils.rs
```diff
@@ -0,0 +1,127 @@
...
+    gas_left: &mut u64,
+) -> Result<(), StatusCode> {
+    consume_copy_cost(gas_left, dest.len() as u64)?;
+    dest[..src.len()].copy_from_slice(src);
+    zero_slice(&mut dest[src.len()..]);
+    Ok(())
+}
+
+#[inline(always)]
+pub fn word_size(bytes: u64) -> Result<u64, StatusCode> {
```
LuisPH3, line 46:
```suggestion
pub fn into_word_size(bytes: u64) -> Result<u64, StatusCode> {
```

rust/src/interpreter/utils.rs
```diff
@@ -0,0 +1,127 @@
...
+    Ok(end / 32)
+}
+
+#[inline(always)]
+pub(super) fn check_min_revision(
+    min_revision: Revision,
+    revision: Revision,
+) -> Result<(), StatusCode> {
+    if revision < min_revision {
+        return Err(StatusCode::EVMC_UNDEFINED_INSTRUCTION);
```
LuisPH3, line 60: This error code is kind of strange in this context. We are talking about revisions, and the error is about instructions. 
I cannot judge until I see how the caller side looks like. I presume this is related to checking for version in certain opcodes, where some sugar will reduce boilerplate. 
Nevertheless, out of context, this symbol is strange.

LuisPH3 (dismissed):
I would really have an iteration on naming. 
Other than that it looks good.

rust/src/interpreter/memory.rs
```diff
@@ -0,0 +1,354 @@
...
+
+#[derive(Debug)]
+pub struct Memory(Vec<u8>);
+
+impl Memory {
+    pub fn new(memory: Vec<u8>) -> Self {
+        Self(memory)
+    }
+
+    pub fn into_inner(self) -> Vec<u8> {
```
LuisPH3, line 21: Is this ever used?  
Shall we provide &[u8] getter and let the caller side decide what to do with it?

rust/src/interpreter/gas.rs
```diff
@@ -0,0 +1,69 @@
...
+    *gas_left -= gas;
+    Ok(())
+}
+
+#[inline(always)]
+pub(super) fn consume_positive_value_cost(
+    value: &u256,
+    gas_left: &mut u64,
+) -> Result<(), StatusCode> {
+    if *value != u256::ZERO {
```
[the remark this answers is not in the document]
LorenzSchueler: What do you mean exactly? The function consumes 9000 gas if value is not null. But the amount of gas does not depend on the value of value.

rust/src/interpreter/memory.rs
LorenzSchueler: Yes, the vector is needed in the ffi module. By providing only `&[u8]` a new allocation would be needed.

rust/src/interpreter/utils.rs
LorenzSchueler: It is used for instructions which have been added in later revisions to fail if the revision is older than the one where the opcode was added.

rust/src/interpreter/utils.rs
LorenzSchueler: I think `byte_len` for the argument makes more sense. Will change that.

rust/src/interpreter/utils.rs
LorenzSchueler: Agreed. I added the `mut` part probably when I was renaming a couple of functions. I will add an extension trait for `[u8]`.

rust/src/interpreter/gas.rs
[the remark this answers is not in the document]
LuisPH3: yes you are right.

rust/src/interpreter/utils.rs
LuisPH3: yes, I understand that. But still. you have a function here comparing revisions in a namespace utils... and then returns a undefined_instruction error out of nowhere. 
Maybe this function should be local to the other file, where it shares more context.

LuisPH3 (approved):

rust/src/interpreter/memory.rs
LuisPH3: OK, I still do not like the name. Into_inner is not a good name, leaks implementation.

LuisPH3 (approved):

rust/src/interpreter/gas.rs
```diff
@@ -0,0 +1,68 @@
...
+    value: &u256,
+    addr: &Address,
+    context: &mut ExecutionContext,
+    gas_left: &mut u64,
+) -> Result<(), StatusCode> {
+    if *value != u256::ZERO && !context.account_exists(addr) {
+        consume_gas(gas_left, 25000)?;
+    }
+    Ok(())
+}
```
HerbertJordan, line 36: I am curious about the code using these two functions ... they seem to be fragments that should be an integral part of some enclosing code -- something computing the cost for a value transfer and only when this is complete, call `consume_gas`.

Could it be that this function, together with the calling code, could be merged into a single `charge_value_transfer_costs` function?

rust/src/interpreter/utils.rs
```diff
@@ -0,0 +1,116 @@
+use std::cmp::min;
+
+use evmc_vm::{ExecutionMessage, MessageFlags, Revision, StatusCode};
+
+use crate::{interpreter::gas::consume_copy_cost, types::u256};
+
+pub trait SliceExt {
```
HerbertJordan, line 7: nit: Please document and, if possible, suggest a better name for this. I would not know what a `SliceExt` is.

rust/src/interpreter/utils.rs
```diff
@@ -0,0 +1,116 @@
...
+    revision: Revision,
+) -> Result<(), StatusCode> {
+    if revision < min_revision {
+        return Err(StatusCode::EVMC_UNDEFINED_INSTRUCTION);
+    }
+    Ok(())
+}
+
+#[inline(always)]
+pub(super) fn check_not_read_only(
```
HerbertJordan, line 72: This is a very surprising utility function. Where would this function be used? And is there any reuse?

LorenzSchueler: This is called for `SStore`, `TStore`, `SelfDestruct`, `Log*`, `Create`, `Create2`, `Call` and `CallCode` to fail if the read-only / static flag was set.

rust/src/interpreter/gas.rs
LorenzSchueler: `consume_positive_value_cost` gets called for `Call` and `CallCode` while `consume_value_to_empty_account_cost` only gets called for the former. When putting both together there would be an additional argument necessary to indicate if the value to empty account cost applies or not. By keeping them as separate functions it is more clear which costs get subtracted and they correspond 1:1 to the names in [evm.codes](https://www.evm.codes/).

rust/src/interpreter/memory.rs
LorenzSchueler: I kind of agree but on the flip side `.into_inner()` is not uncommon. e.g. `Box::into_inner` or `BufReader::into_inner`. And the fact that `Memory` holds a `Vec` could be also inferred from `Memory::new` which takes a `Vec`. But I would also be fine with renaming it.

rust/src/interpreter/utils.rs
LorenzSchueler: Agreed. Will do that if a future PR.

rust/src/interpreter/utils.rs
LorenzSchueler: In my mind `utils` is more or less a collection of stuff that does clearly belong to any particular other module. But I'm fine with moving it.

Issue #644, opened 2024-08-13
Cleanup of LFVM code conversion


HerbertJordan:
The design and implementation of LFVM's code conversion infrastructure needs to be reviewed and refined.

The [implementation](https://github.com/Fantom-foundation/Tosca/blob/438970e2907bf1f4cf69a97efa461ab25179e5e1/go/interpreter/lfvm/converter.go#L40) has grown over time organically and may exhibit more flags and code paths than are actually used / necessary in its current version.

A clean definition of its role and task should be defined, unit tests formalizing this definition implemented, and the implementation adapted to match this specification.

HerbertJordan:
Things that still need to be done:

- [ ] cleanup of conversion code interface
- [ ] document interface and describe intended semantic
- [ ] make tests enforcing the newly defined semantic

PR #662, opened 2024-08-22
LFVM: remove converter err

facuMH:
This PR removes the returned error from `converter` and `Converter` functions from LFVM, and adds testing for the converter file. 

The particular error that is removed (previously in line `converter.go#111 `) can not be triggered (`res.length() > i`) because:
note: `.lenght()` returns the value of `res.nextPos`
- in the case of `VM.JUMPDEST`, `padNoOpsUntil(i)` updates `nextPos` to the value of `i`, then `appendCode` increases `nextPos` by one, lastly `i` is also increased by one. 
- `appendInstructions` makes calls to `appendOp` and `appendData` and returns however many times this functions were called -1,  and this returned value is then used to increment `i`, +1. 
So it can never happen that `nextPos > i` 

Also the check that all valid opcodes are covered is moved to a unit test, even if this panic could only trigger during initialization, now that all opcodes are implemented it should be a unit test.

### DIFF
--- a/go/interpreter/lfvm/converter.go
+++ b/go/interpreter/lfvm/converter.go
@@ -37,7 +37,7 @@ func clearConversionCache() {
 	cache.Purge()
 }
 
-func Convert(code []byte, withSuperInstructions bool, isInitCode bool, noCodeCache bool, codeHash tosca.Hash) (Code, error) {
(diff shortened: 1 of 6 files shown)


LuisPH3 (approved):
Thanks for splitting this change.

LuisPH3:
How does these changes relate to https://github.com/Fantom-foundation/Tosca/issues/644 ?

HerbertJordan (approved):
Thanks for cleaning this up!

PR #664, opened 2024-08-22
LFVM: add hash cache tests

facuMH:
This PR renames the `HashCache` type to `hashCache` to make it private, renames the static variable `hashCache` to `hashVault`, adds tests for the hash cache and removes `hash_cache.go` from the override list.


LuisPH3:
> This PR renames the `HashCache` type to `hashCache` to make it private, renames the static variable `hashCache` to `hashVault`, adds tests for the hash cache and removes `hash_cache.go` from the override list.
> 
> it also removes an `else` case from `getHashXX`. This `else` can never happen since the first time an element is added it would create a `head` and this is always updated, but never set to nil. The only case where this "could" happen is if calling `hashCache{}` instead of calling `newHashCache` function, and even then the program would crash when trying to add the first element, not even reaching the point of getting the hash for an existing entry whose `succ` is nil.

Note, this description is fine, but first you could say what was the problem (someone calls wrong constructor), and afterwards the action taken (make type private).

go/interpreter/lfvm/instructions.go
```diff
@@ -581,7 +581,7 @@ func opExp(c *context) {
 }
 
 // Evaluations show a 96% hit rate of this configuration.
-var hashCache = newHashCache(1<<16, 1<<18)
+var hashVault = newHashCache(1<<16, 1<<18)
```
LuisPH3, line 584: :thinking: just `cache`?  `sha3Cache`?

LuisPH3 (dismissed):

go/interpreter/lfvm/hash_cache.go
```diff
@@ -124,8 +124,6 @@ func (h *HashCache) getHash32(c *context, data []byte) tosca.Hash {
 			entry.pred.succ = entry.succ
 			if entry.succ != nil {
 				entry.succ.pred = entry.pred
-			} else {
-				h.tail32 = entry.pred
```
HerbertJordan, line 128: This line of code is triggered if the very last element in the cache is touched and moved to the front. An example test provoking it is this:
```
func TestHashCache_TouchLastElement(t *testing.T) {
	ctxt := &context{}
	cache := newHashCache(3, 3)
	cache.getHash32(ctxt, []byte{byte(1)})
	cache.getHash32(ctxt, []byte{byte(2)})
	cache.getHash32(ctxt, []byte{byte(3)})
	cache.getHash32(ctxt, []byte{byte(1)})
}
```
Without this line the tail would end up pointing to the head when touching the last element in the LRU list and the LRU order is destroyed.

go/interpreter/lfvm/instructions.go
HerbertJordan: I like `sha3Cache` ...

HerbertJordan (changes requested):
> it also removes an else case from getHashXX. This else can never happen since the first time an element is added it would create a head and this is always updated, but never set to nil. The only case where this "could" happen is if calling hashCache{} instead of calling newHashCache function, and even then the program would crash when trying to add the first element, not even reaching the point of getting the hash for an existing entry whose succ is nil.

This argument is wrong (see comment) and I think removing those `else` branches breaks the cache.

LuisPH3 (approved):

simonlechner (approved):

go/interpreter/lfvm/hash_cache_test.go
```diff
@@ -0,0 +1,210 @@
...
+	if entry3 != hash3 {
+		t.Fatalf("second cal for same data should return same hash. want %v but got %v", hash3, entry3)
+	}
+	entry2 := cache.hash(ctxt, data2)
+	if entry2 != hash2 {
+		t.Fatalf("second cal for same data should return same hash. want %v but got %v", hash2, entry2)
+	}
+}
+
+func TestHashCache_useCacheTail32(t *testing.T) {
```
HerbertJordan, line 132: I think those tests are unnecessarily convoluted and seem to be mainly focused on pushing coverage up. These tests are not sufficient to test the actual requirements on the hash cache.

HerbertJordan (changes requested):
I think the way the hash cache is tested needs to be rethought. Substantial properties are not tested, including its thread safety and the node replacement policy. See, for example, these [tests](https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/node_cache_test.go) for the node cache in Carmen for reference.

We should discuss the general testing approach before moving forward.

facuMH:
the hash cache will undergo some refactoring, and the testing will get a new approach, hence I'm closing this PR

PR #665, opened 2024-08-22
LFVM: instruction & interpreter tests

facuMH:
This PR:
- adds tests for instruction `.String()` and code `.String()`
- removes unused error code `SEGMENTATION_FAULT`
- moves `switch` from `Run` that generates `tosca.Result` into its own function for ease of testing.
- adds tests for the multiple cases of `Run`
- adds tests checking for correct recording of statistics
- adds tests for statistics and logs printing.

### DIFF
--- a/go/interpreter/lfvm/instruction_test.go
+++ b/go/interpreter/lfvm/instruction_test.go
@@ -1,0 +1,35 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package lfvm
+
+import "testing"
+
+func TestInstruction_String(t *testing.T) {
+	instruction := Instruction{opcode: STOP, arg: 0x0000}
+	if got, want := instruction.String(), "STOP"; got != want {
+		t.Errorf("Instruction.String() = %q, want %q", got, want)
+	}
+
+	instruction = Instruction{opcode: PUSH1, arg: 0x0001}
+	if got, want := instruction.String(), "PUSH1 0x0001"; got != want {
+		t.Errorf("Instruction.String() = %q, want %q", got, want)
+	}
+}
+
+func TestCode_String(t *testing.T) {
+	code := Code{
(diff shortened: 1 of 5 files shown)


go/interpreter/lfvm/instruction_test.go
```diff
@@ -0,0 +1,35 @@
...
+		t.Errorf("Instruction.String() = %q, want %q", got, want)
+	}
+
+	instruction = Instruction{opcode: PUSH1, arg: 0x0001}
+	if got, want := instruction.String(), "PUSH1 0x0001"; got != want {
+		t.Errorf("Instruction.String() = %q, want %q", got, want)
+	}
+}
+
+func TestCode_Strign(t *testing.T) {
```
HerbertJordan, line 27:
```suggestion
func TestCode_String(t *testing.T) {
```

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -453,6 +462,427 @@ func TestOKInstructionPath(t *testing.T) {
...
+		Static: true,
+		Gas:    10,
+		Code:   []byte{0x0},
+	}
+	ctxt.code = instructions
+
+	// redirect stdout
+	old := os.Stdout
+	r, w, _ := os.Pipe()
+	os.Stdout = w
```
HerbertJordan, line 534: This is fine for now, to increase the test coverage, but in general the logging mode should get an improvement. The logging target should not be `os.Stdout` but some log consumer with some mock-able interface (just fyi, nothing to be changed here).

HerbertJordan (dismissed):

LuisPH3 (approved):

LuisPH3 (commented):
Can we just avoid the ignore line for now? 
If this code is tested somehow else: it should be ok to lower the threshold for the file.

go/interpreter/lfvm/interpreter.go
```diff
@@ -423,7 +427,7 @@ func steps(c *context, one_step_only bool) {
 		}
 
 		// Execute instruction
-		switch op {
+		switch op { // coverage-ignore opcodes are tested by CT
```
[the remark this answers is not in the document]
facuMH: yes, this can be done. I thought it would best to ignore this `switch` because it is such a large block of code, that we know is tested by another big project of this repo. But yes, an override can be done for this file anyway.

HerbertJordan (dismissed):

simonlechner (approved):

LuisPH3 (approved):

PR #666, opened 2024-08-23
LFVM: memory tests

facuMH:
This PR cleans a little bit the memory functions, in particular those that call `EnsureCapacity` and then other memory methods. 
Many checks are removed, all of them checking that there is space in the memory, but since this is ensured by `EnsureCapacity` which is called just before the checks, these checks cannot fail.

Tests are added to verify expected behaviour and error reporting of memory methods.

### DIFF
--- a/go/interpreter/lfvm/memory.go
+++ b/go/interpreter/lfvm/memory.go
@@ -13,6 +13,7 @@ package lfvm
 import (
 	"fmt"
 	"math"
+	"strings"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	"github.com/holiman/uint256"
@@ -168,15 +169,15 @@ func (m *Memory) Set(offset, size uint64, value []byte) error {
 			return errGasUintOverflow
 		}
 		if offset+size > m.Len() {
-			return errSetMemTooSmall(m.Len(), size, offset)
+			return errInsufficientMemory(m.Len(), size, offset)
 		}
 		copy(m.store[offset:offset+size], value)
 	}
 	return nil
 }
 
-func errSetMemTooSmall(memSize, size, offset uint64) error {
-	return ConstError(fmt.Sprintf("memory too small, size %d, attempted to write %d bytes at %d", memSize, size, offset))
+func errInsufficientMemory(memSize, size, offset uint64) error {
+	return tosca.ConstError(fmt.Sprintf("memory too small, size %d, attempted to write %d bytes at %d", memSize, size, offset))
 }
 
 func (m *Memory) SetWithCapacityAndGasCheck(offset, size uint64, value []byte, c *context) error {
@@ -246,19 +247,21 @@ func (m *Memory) Data() []byte {
 	return m.store
 }
 
-func (m *Memory) Print() {
-	fmt.Printf("### mem %d bytes ###\n", len(m.store))
+func (m *Memory) Print() string {
+	returnString := strings.Builder{}
+	returnString.WriteString(fmt.Sprintf("### mem %d bytes ###\n", len(m.store)))
 	if len(m.store) > 0 {
 		addr := 0
 		for i := 0; i+32 <= len(m.store); i += 32 {
-			fmt.Printf("%03d: % x\n", addr, m.store[i:i+32])
+			returnString.WriteString(fmt.Sprintf("%03d: % x\n", addr, m.store[i:i+32]))
 			addr++
 		}
 		if len(m.store)%32 != 0 {
-			fmt.Printf("%03d: % x\n", addr, m.store[len(m.store)/32*32:])
+			returnString.WriteString(fmt.Sprintf("%03d: % x\n", addr, m.store[len(m.store)/32*32:]))
 		}
 	} else {
-		fmt.Println("-- empty --")
+		returnString.WriteString("-- empty --\n")
 	}
-	fmt.Println("####################")
+	returnString.WriteString("####################\n")
+	return returnString.String()
 }
--- a/go/interpreter/lfvm/memory.go
+++ b/go/interpreter/lfvm/memory.go
@@ -168,11 +168,15 @@ func (m *Memory) Set(offset, size uint64, value []byte) error {
 			return errGasUintOverflow
 		}
 		if offset+size > m.Len() {
-			return fmt.Errorf("memory too small, size %d, attempted to write %d bytes at %d", m.Len(), size, offset)
+			return errSetMemTooSmall(m.Len(), size, offset)
 		}
 		copy(m.store[offset:offset+size], value)
 	}
 	return nil
+}
+
+func errSetMemTooSmall(memSize, size, offset uint64) error {
+	return ConstError(fmt.Sprintf("memory too small, size %d, attempted to write %d bytes at %d", memSize, size, offset))
 }
 
 func (m *Memory) SetWithCapacityAndGasCheck(offset, size uint64, value []byte, c *context) error {
--- a/go/interpreter/lfvm/memory.go
+++ b/go/interpreter/lfvm/memory.go
@@ -197,8 +197,7 @@ func (m *Memory) CopyWord(offset uint64, trg *uint256.Int, c *context) error {
 	if err != nil {
 		return err
 	}
-	toWrite := m.store[offset : offset+32]
-	trg.SetBytes32(toWrite) //m.store[offset : offset+32])
+	trg.SetBytes32(m.store[offset : offset+32])
 	return nil
 }
 
--- a/go/interpreter/lfvm/memory.go
+++ b/go/interpreter/lfvm/memory.go
@@ -180,22 +180,12 @@ func (m *Memory) SetWithCapacityAndGasCheck(offset, size uint64, value []byte, c
 	if err != nil {
 		return err
 	}
-	err = m.Set(offset, size, value)
-	if err != nil {
-		return err
-	}
-	return nil
+	return m.Set(offset, size, value)
 }
 
 func (m *Memory) SetWithCapacityCheck(offset, size uint64, value []byte) error {
 	m.EnsureCapacityWithoutGas(size)
-	if size > 0 {
-		if offset+size < offset {
-			return errGasUintOverflow
-		}
-		copy(m.store[offset:offset+size], value)
-	}
-	return nil
+	return m.Set(offset, size, value)
 }
 
 func (m *Memory) CopyWord(offset uint64, trg *uint256.Int, c *context) error {
--- a/go/interpreter/lfvm/memory.go
+++ b/go/interpreter/lfvm/memory.go
@@ -69,43 +69,52 @@ func (m *Memory) expansionCosts(size uint64) tosca.Gas {
 	return fee
 }
 
-// ensureCapacity tries to expand memory to the given size.
+// expandMemoryAndCharge tries to expand memory to the given size.
 // If the memory is already large enough or size is 0, it does nothing.
 // if not enough gas in the context, it returns an error.
-func (m *Memory) ensureCapacity(offset, size uint64, c *context) error {
-	if size <= 0 {
-		return nil
-	}
+func (m *Memory) expandMemoryAndCharge(offset, size uint64, c *context) error {
+
 	needed := offset + size
 	// check overflow
 	if needed < offset {
 		c.signalError()
 		return errGasUintOverflow
 	}
-	if m.len() < needed {
-		needed = toValidMemorySize(needed)
-		fee := m.expansionCosts(needed)
+	err := m.expandMemoryWithGasFun(needed, func(fee tosca.Gas) error {
 		if !c.useGas(fee) {
 			c.status = statusOutOfGas
 			return errOutOfGas
 		}
-		m.total_memory_cost += fee
-		m.store = append(m.store, make([]byte, needed-m.len())...)
+		return nil
+	})
+	if err != nil {
+		return err
 	}
+
 	return nil
 }
 
-// ensureCapacityWithoutGas expands the memory to the given size without checking gas.
-func (m *Memory) ensureCapacityWithoutGas(size uint64) {
+// expandMemory expands the memory to the given size without checking gas.
+func (m *Memory) expandMemory(size uint64) {
+	m.expandMemoryWithGasFun(size, nil)
+}
+
+func (m *Memory) expandMemoryWithGasFun(size uint64, gasFun func(tosca.Gas) error) error {
 	if size <= 0 {
-		return
+		return nil
 	}
 	if m.len() < size {
 		size = toValidMemorySize(size)
 		fee := m.expansionCosts(size)
+		if gasFun != nil {
+			if err := gasFun(fee); err != nil {
+				return err
+			}
+		}
 		m.total_memory_cost += fee
 		m.store = append(m.store, make([]byte, size-m.len())...)
 	}
+	return nil
 }
 
 func (m *Memory) len() uint64 {
@@ -114,7 +123,7 @@ func (m *Memory) len() uint64 {
 
 // setByte sets a byte at the given offset, expanding memory as needed and charging for it.
 func (m *Memory) setByte(offset uint64, value byte, c *context) error {
-	err := m.ensureCapacity(offset, 1, c)
+	err := m.expandMemoryAndCharge(offset, 1, c)
 	if err != nil {
 		return err
 	}
@@ -124,7 +133,7 @@ func (m *Memory) setByte(offset uint64, value byte, c *context) error {
 
 // GetByte returns a byte at the given offset, expanding memory as needed and charging for it.
 func (m *Memory) setWord(offset uint64, value *uint256.Int, c *context) error {
-	err := m.ensureCapacity(offset, 32, c)
+	err := m.expandMemoryAndCharge(offset, 32, c)
 	if err != nil {
 		return err
 	}
@@ -187,7 +196,7 @@ func makeInsufficientMemoryError(memSize, size, offset uint64) error {
 }
 
 func (m *Memory) setWithCapacityAndGasCheck(offset, size uint64, value []byte, c *context) error {
-	err := m.ensureCapacity(offset, size, c)
+	err := m.expandMemoryAndCharge(offset, size, c)
 	if err != nil {
 		return err
 	}
@@ -195,7 +204,7 @@ func (m *Memory) setWithCapacityAndGasCheck(offset, size uint64, value []byte, c
 }
 
 func (m *Memory) getWord(offset uint64, trg *uint256.Int, c *context) error {
-	err := m.ensureCapacity(offset, 32, c)
+	err := m.expandMemoryAndCharge(offset, 32, c)
 	if err != nil {
 		return err
 	}
@@ -232,28 +241,9 @@ func (m *Memory) getSlice(offset, size uint64) []byte {
 }
 
 func (m *Memory) getSliceWithCapacityAndGas(offset, size uint64, c *context) ([]byte, error) {
-	err := m.ensureCapacity(offset, size, c)
+	err := m.expandMemoryAndCharge(offset, size, c)
 	if err != nil {
 		return nil, err
 	}
 	return m.getSlice(offset, size), nil
 }
-
-// func (m *Memory) Print() string {
-// 	returnString := strings.Builder{}
-// 	returnString.WriteString(fmt.Sprintf("### mem %d bytes ###\n", len(m.store)))
-// 	if len(m.store) > 0 {
-// 		addr := 0
-// 		for i := 0; i+32 <= len(m.store); i += 32 {
-// 			returnString.WriteString(fmt.Sprintf("%03d: % x\n", addr, m.store[i:i+32]))
-// 			addr++
-// 		}
-// 		if len(m.store)%32 != 0 {
-// 			returnString.WriteString(fmt.Sprintf("%03d: % x\n", addr, m.store[len(m.store)/32*32:]))
-// 		}
-// 	} else {
-// 		returnString.WriteString("-- empty --\n")
-// 	}
-// 	returnString.WriteString("####################\n")
-// 	return returnString.String()
-// }
--- a/go/interpreter/lfvm/memory.go
+++ b/go/interpreter/lfvm/memory.go
@@ -13,7 +13,6 @@ package lfvm
 import (
 	"fmt"
 	"math"
-	"strings"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	"github.com/holiman/uint256"
@@ -44,7 +43,8 @@ const (
 	maxMemoryExpansionSize = 0x1FFFFFFFE0
 )
 
-func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
+// expansionCosts returns the gas cost of expanding the memory to the given size.
+func (m *Memory) expansionCosts(size uint64) tosca.Gas {
 
 	// static assert
 	const (
@@ -54,7 +54,7 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
 		_                 = int64(maxInWords*maxInWords/512 + 3*maxInWords)
 	)
 
-	if m.Len() >= size {
+	if m.len() >= size {
 		return 0
 	}
 	size = toValidMemorySize(size)
@@ -69,7 +69,10 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
 	return fee
 }
 
-func (m *Memory) EnsureCapacity(offset, size uint64, c *context) error {
+// ensureCapacity tries to expand memory to the given size.
+// If the memory is already large enough or size is 0, it does nothing.
+// if not enough gas in the context, it returns an error.
+func (m *Memory) ensureCapacity(offset, size uint64, c *context) error {
 	if size <= 0 {
 		return nil
 	}
@@ -79,37 +82,39 @@ func (m *Memory) EnsureCapacity(offset, size uint64, c *context) error {
 		c.signalError()
 		return errGasUintOverflow
 	}
-	if m.Len() < needed {
+	if m.len() < needed {
 		needed = toValidMemorySize(needed)
-		fee := m.ExpansionCosts(needed)
+		fee := m.expansionCosts(needed)
 		if !c.useGas(fee) {
 			c.status = statusOutOfGas
 			return errOutOfGas
 		}
 		m.total_memory_cost += fee
-		m.store = append(m.store, make([]byte, needed-m.Len())...)
+		m.store = append(m.store, make([]byte, needed-m.len())...)
 	}
 	return nil
 }
 
-func (m *Memory) EnsureCapacityWithoutGas(size uint64) {
+// ensureCapacityWithoutGas expands the memory to the given size without checking gas.
+func (m *Memory) ensureCapacityWithoutGas(size uint64) {
 	if size <= 0 {
 		return
 	}
-	if m.Len() < size {
+	if m.len() < size {
 		size = toValidMemorySize(size)
-		fee := m.ExpansionCosts(size)
+		fee := m.expansionCosts(size)
 		m.total_memory_cost += fee
-		m.store = append(m.store, make([]byte, size-m.Len())...)
+		m.store = append(m.store, make([]byte, size-m.len())...)
 	}
 }
 
-func (m *Memory) Len() uint64 {
+func (m *Memory) len() uint64 {
 	return uint64(len(m.store))
 }
 
-func (m *Memory) SetByte(offset uint64, value byte, c *context) error {
-	err := m.EnsureCapacity(offset, 1, c)
+// setByte sets a byte at the given offset, expanding memory as needed and charging for it.
+func (m *Memory) setByte(offset uint64, value byte, c *context) error {
+	err := m.ensureCapacity(offset, 1, c)
 	if err != nil {
 		return err
 	}
@@ -117,8 +122,9 @@ func (m *Memory) SetByte(offset uint64, value byte, c *context) error {
 	return nil
 }
 
-func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
-	err := m.EnsureCapacity(offset, 32, c)
+// GetByte returns a byte at the given offset, expanding memory as needed and charging for it.
+func (m *Memory) setWord(offset uint64, value *uint256.Int, c *context) error {
+	err := m.ensureCapacity(offset, 32, c)
 	if err != nil {
 		return err
 	}
@@ -163,13 +169,13 @@ func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
 	return nil
 }
 
-func (m *Memory) Set(offset, size uint64, value []byte) error {
+func (m *Memory) set(offset, size uint64, value []byte) error {
 	if size > 0 {
 		if offset+size < offset {
 			return errGasUintOverflow
 		}
-		if offset+size > m.Len() {
-			return makeInsufficientMemoryError(m.Len(), size, offset)
+		if offset+size > m.len() {
+			return makeInsufficientMemoryError(m.len(), size, offset)
 		}
 		copy(m.store[offset:offset+size], value)
 	}
@@ -180,21 +186,16 @@ func makeInsufficientMemoryError(memSize, size, offset uint64) error {
 	return tosca.ConstError(fmt.Sprintf("memory too small, size %d, attempted to write %d bytes at %d", memSize, size, offset))
 }
 
-func (m *Memory) SetWithCapacityAndGasCheck(offset, size uint64, value []byte, c *context) error {
-	err := m.EnsureCapacity(offset, size, c)
+func (m *Memory) setWithCapacityAndGasCheck(offset, size uint64, value []byte, c *context) error {
+	err := m.ensureCapacity(offset, size, c)
 	if err != nil {
 		return err
 	}
-	return m.Set(offset, size, value)
+	return m.set(offset, size, value)
 }
 
-func (m *Memory) SetWithCapacityCheck(offset, size uint64, value []byte) error {
-	m.EnsureCapacityWithoutGas(size)
-	return m.Set(offset, size, value)
-}
-
-func (m *Memory) CopyWord(offset uint64, trg *uint256.Int, c *context) error {
-	err := m.EnsureCapacity(offset, 32, c)
+func (m *Memory) getWord(offset uint64, trg *uint256.Int, c *context) error {
+	err := m.ensureCapacity(offset, 32, c)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +53,230 @@ func TestExpansionCosts(t *testing.T) {
...
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			m := NewMemory()
+			m.store = test.store
+			target := make([]byte, 8)
+
+			m.CopyData(test.offset, target)
+			if !slices.Equal(target, test.expected) {
```
LuisPH3, line 204: nit: Recently I learned `bytes.Equal`.

LuisPH3 (approved):

go/interpreter/lfvm/memory.go
```diff
@@ -188,35 +180,19 @@ func (m *Memory) SetWithCapacityAndGasCheck(offset, size uint64, value []byte, c
...
-	err = m.Set(offset, size, value)
-	if err != nil {
-		return err
-	}
-	return nil
+	return m.Set(offset, size, value)
 }
 
 func (m *Memory) SetWithCapacityCheck(offset, size uint64, value []byte) error {
 	m.EnsureCapacityWithoutGas(size)
```
LuisPH3, line 187: This may be a little out of scope, but I really dont understand the name of this function:
```
func (m *Memory) EnsureCapacityWithoutGas(size uint64) {
	if size <= 0 {
		return
	}
	if m.Len() < size {
		size = toValidMemorySize(size)
		fee := m.ExpansionCosts(size)
		m.total_memory_cost += fee
		m.store = append(m.store, make([]byte, size-m.Len())...)
	}
}
```
Because the changes in this PR rely on this function, could be rename/document it so that I can see if we are doing here the right thing?

LuisPH3 (approved):

LuisPH3 (dismissed):

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +54,230 @@ func TestExpansionCosts(t *testing.T) {
...
+			m.store = make([]byte, 8)
+			ctxt.gas = 100
+
+			err := m.Set(test.offset, test.size, []byte{0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef})
+
+			if err == nil {
+				if test.expected != nil {
+					t.Errorf("expected error %v, got nil", test.expected)
+				}
+			} else if !strings.Contains(err.Error(), test.expected.Error()) {
```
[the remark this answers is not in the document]
facuMH: currently, `Set` returns an error that is instantiated as needed, so `errors.Is` does not work, and I need to compare string. Also not sure I could make the error into a `const` (this would enable using `errors.is` ), because the error takes a couple of parameters.

go/interpreter/lfvm/memory.go
```diff
@@ -188,35 +180,19 @@ func (m *Memory) SetWithCapacityAndGasCheck(offset, size uint64, value []byte, c
...
-		if offset+size < offset {
-			return errGasUintOverflow
-		}
-		if offset+size > m.Len() {
-			return fmt.Errorf("memory too small, size %d, attempted to write %d bytes at %d", m.Len(), size, offset)
-		}
-		copy(m.store[offset:offset+size], value)
-	}
-	return nil
+	return m.Set(offset, size, value)
```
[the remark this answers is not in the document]
facuMH: it could be the caller's responsibility to call it with `offset+size`, but I don't like that approach. this does seem weird, specially because whenever `EnsureCapacityWithoutGas` is called, the memory operation that comes after always have offset and size. 
That said, this change should probably be in a different PR with a little more analysis.

go/interpreter/lfvm/memory_test.go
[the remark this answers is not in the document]
facuMH: I've added an error generating function, that allows the use of `errors.Is`, we should make the usage of such a functions a project wide practice, since we don't need to compare strings anymore.

go/interpreter/lfvm/memory.go
facuMH: this was clarified offline

go/interpreter/lfvm/memory.go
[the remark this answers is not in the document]
HerbertJordan: It looks like there is only a single call of this function which indeed uses the offset `0`. It should be checked whether this function can be eliminated in a follow-up PR.

go/interpreter/lfvm/memory_test.go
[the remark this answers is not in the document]
HerbertJordan: I agree

go/interpreter/lfvm/memory.go
```diff
@@ -176,47 +168,35 @@ func (m *Memory) Set(offset, size uint64, value []byte) error {
...
 		if offset+size > m.Len() {
-			return fmt.Errorf("memory too small, size %d, attempted to write %d bytes at %d", m.Len(), size, offset)
+			return errSetMemTooSmall(m.Len(), size, offset)
 		}
 		copy(m.store[offset:offset+size], value)
 	}
 	return nil
 }
 
+func errSetMemTooSmall(memSize, size, offset uint64) error {
```
LuisPH3, line 178: I think that this an elegant solution.

go/interpreter/lfvm/memory_test.go
```diff
@@ -11,13 +11,18 @@
...
 	"math"
+	"os"
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/holiman/uint256"
 )
 
-func TestExpansionCosts(t *testing.T) {
+func TestMemory_ExpansionCosts(t *testing.T) {
```
LuisPH3, line 25:
```suggestion
func TestMemory_ExpansionCosts_ComputesCorrectCosts(t *testing.T) {
```

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+			expectedTarget: *uint256.NewInt(0),
+			expectedError:  errOutOfGas},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			ctxt := getEmptyContext()
+			m := NewMemory()
+			target := uint256.NewInt(0)
+			m.store = make([]byte, 32)
```
LuisPH3, line 83: Initialize data to other than zero: By having zero values, you cannot differentiate zero initialization from faulty assignment.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+	tests := map[string]struct {
+		gas            tosca.Gas
+		offset         uint64
+		expectedTarget uint256.Int
+		expectedError  error
+	}{
+		"regular": {
+			gas:            100,
+			offset:         0,
+			expectedTarget: *uint256.NewInt(0).SetBytes8(testValue),
```
LuisPH3, line 69: expectedData

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+			target := uint256.NewInt(0)
+			m.store = make([]byte, 32)
+			copy(m.store[24:], testValue)
+			ctxt.gas = test.gas
+
+			err := m.CopyWord(test.offset, target, &ctxt)
+			if !errors.Is(err, test.expectedError) {
+				t.Errorf("unexpected error, want: %v, got: %v", test.expectedError, err)
+			}
+			if target.Cmp(&test.expectedTarget) != 0 {
```
LuisPH3, line 91: If the previous check was an error, we do not check the data read.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+			gas:            100,
+			offset:         0,
+			expectedTarget: *uint256.NewInt(0).SetBytes8(testValue),
+			expectedError:  nil},
+		"not enough gas": {
+			gas:            6,
+			offset:         64,
+			expectedTarget: *uint256.NewInt(0),
+			expectedError:  errOutOfGas},
+	}
```
LuisPH3, line 76: Add test where offset < mem size and offeset == mem size.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
 		}
 	}
 }
+
+func TestMemory_CopyWord(t *testing.T) {
```
LuisPH3, line 56:
```suggestion
func TestMemory_CopyWord_CopiesDataOrRetursError(t *testing.T) {
```

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			ctxt := getEmptyContext()
+			m := NewMemory()
+			ctxt.gas = test.gas
+
+			err := m.SetByte(test.offset, test.value, &ctxt)
+			if !errors.Is(err, test.expected) {
+				t.Errorf("unexpected error, want: %v, got: %v", test.expected, err)
+			}
+			if err == nil && m.store[test.offset] != test.value {
```
LuisPH3, line 120: nit: do not mix concepts in conditions.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+
+	tests := map[string]struct {
+		value    byte
+		offset   uint64
+		gas      tosca.Gas
+		expected error
+	}{
+		"regular":        {value: 0x12, offset: 0, gas: 100, expected: nil},
+		"not enough gas": {value: 0x12, offset: 64, gas: 0, expected: errOutOfGas},
+	}
```
LuisPH3, line 108: Add scenarios where memory has already data, and scenarios where it does not. 
aim for the bounduaries

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+				t.Errorf("unexpected error, want: %v, got: %v", test.expected, err)
+			}
+			if err == nil && m.store[test.offset] != test.value {
+				t.Errorf("unexpected value, want: %v, got: %v", test.value, m.store[test.offset])
+			}
+		})
+	}
+}
+
+func TestSetWordProperlyReportsNotEnoughGas(t *testing.T) {
```
LuisPH3, line 127:
```suggestion
func TestMemory_SetWord_ReportsNotEnoughGas(t *testing.T) {
```

LuisPH3: This test may need to add a table like in `CopyWord`
Test in-out bounds, test that value is correct or an error.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+		"padding needed": {store: baseStore,
+			offset: 4, target: make([]byte, 8),
+			expected: []byte{0x90, 0xab, 0xcd, 0xef, 0, 0, 0, 0}},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			m := NewMemory()
+			m.store = test.store
+			target := make([]byte, 8)
```
LuisPH3, line 202: nit: by not initializing with zeros, you can check that the memory is actually copied and not just expanded and copy suffers from some bonduary error.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,232 @@ func TestExpansionCosts(t *testing.T) {
...
+		store    []byte
+		offset   uint64
+		size     uint64
+		expected []byte
+	}{
+		"regular":   {store: baseStore, offset: 0, size: 8, expected: baseStore},
+		"size zero": {store: baseStore, offset: 0, size: 0, expected: nil},
+		"size+offset bigger than memory": {store: baseStore, offset: 4, size: 8,
+			expected: nil},
+	}
```
LuisPH3, line 227: This test needs to add scenarios for expansions. 
with/without prefilled memory, offset in/out boundaries, stress boundaries.

LuisPH3 (changes requested):

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,304 @@ func TestExpansionCosts(t *testing.T) {
...
+	}{
+		"regular empty memory": {value: 0x12, offset: 0, gas: 100, expected: nil},
+		"not enough gas":       {value: 0x12, offset: 64, gas: 0, expected: errOutOfGas},
+		"offset overflow": {value: 0x12, offset: math.MaxUint64, gas: 100,
+			expected: errGasUintOverflow},
+		"regular with memory": {memory: []byte{0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef},
+			value: 0x12, offset: 0, gas: 100, expected: nil},
+		"regular with memory and offset": {memory: []byte{0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef},
+			value: 0x12, offset: 4, gas: 100, expected: nil},
+	}
```
LuisPH3, line 129: Still missing memory off-bounds access (with expansion) 
multiline would help review, a lot.

LuisPH3 (changes requested):
The unit tests are missing cases:
For each unit test, that could have memory expansion.

- Test empty
- test case without expansion
- test case in expansion bonduary
- test case out of bounds

And then make the product with gas/not enough gas/gas overflow for each of them. The list before may not even be complete, please have a look at it and add all the relevant failure scenarios.

go/interpreter/lfvm/memory.go
```diff
@@ -176,48 +168,37 @@ func (m *Memory) Set(offset, size uint64, value []byte) error {
...
+	return ConstError(fmt.Sprintf("memory too small, size %d, attempted to write %d bytes at %d", memSize, size, offset))
```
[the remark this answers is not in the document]
facuMH: using `ConstError` it can be compared using `errors.Is` but using `fmt.Errorf` the comparison fails.

go/interpreter/lfvm/memory.go
```diff
@@ -176,48 +168,37 @@ func (m *Memory) Set(offset, size uint64, value []byte) error {
...
 	err := m.EnsureCapacity(offset, 32, c)
 	if err != nil {
 		return err
 	}
-	if m.Len() < offset+32 {
-		return fmt.Errorf("memory too small, size %d, attempted to read 32 byte at position %d", m.Len(), offset)
-	}
-	trg.SetBytes32(m.store[offset : offset+32])
+	toWrite := m.store[offset : offset+32]
+	trg.SetBytes32(toWrite) //m.store[offset : offset+32])
```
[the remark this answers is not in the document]
facuMH: sorry, debug code that got away. I'll remove it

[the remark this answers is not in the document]
facuMH: sorry, debug code that got away. I'll remove it

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,304 @@ func TestExpansionCosts(t *testing.T) {
...
+		"offset fits in memory": {
+			gas:           100,
+			offset:        24,
+			expectedData:  *uint256.NewInt(1).SetBytes32(testValue2),
+			expectedError: nil},
+		"offset same as memory size": {
+			gas:           100,
+			offset:        8,
+			expectedData:  *uint256.NewInt(1).SetBytes32(testValue3),
+			expectedError: nil},
```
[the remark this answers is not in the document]
facuMH: I renamed the test cases and expected data so that now it is easier to read (and makes sense)

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+	baseData := []byte{0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef}
+
+	tests := map[string]struct {
+		memory []byte
+		offset uint64
+	}{
+		"regular empty memory": {},
+		"regular with memory": {
+			memory: baseData,
+			offset: 0},
```
LuisPH3, line 154: Egyptian brackets? 
I know that you do not like multiline, but it really helps with the review. 
```go
"regular with memory": {
			memory: baseData,
			offset: 0,
},
``` 
Trailing comas help to keep sane structure as well.

LuisPH3 (approved):
Test scenarios have improved a lot.

LuisPH3 (changes requested):
please remove WIP commit.

LuisPH3 (dismissed):

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+			target := baseTargetValue
+			m.store = make([]byte, memorySize)
+			copy(m.store[24:], valueSmall.Bytes())
+			ctxt.gas = test.gas
+
+			err := m.CopyWord(test.offset, target, &ctxt)
+			if !errors.Is(err, test.expectedError) {
+				t.Fatalf("unexpected error, want: %v, got: %v", test.expectedError, err)
+			}
+			if target.Cmp(baseTargetValue) != 0 {
```
HerbertJordan, line 136: By default, if there is an error, there should not be any guarantees on the target value unless explicitly documented. So this check may be unnecessary.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+			expectedData: testValue},
+		"not enough gas": {
+			gas:           1,
+			offset:        32,
+			expectedError: errOutOfGas},
+		"offset fits in memory": {
+			gas:          6,
+			offset:       24,
+			expectedData: testValue},
+		"offset same as memory size": {
```
HerbertJordan, line 243: this test case actually has an offset of 0

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			ctxt := getEmptyContext()
+			m := NewMemory()
+			ctxt.gas = test.gas
+			err := m.SetWord(test.offset, uint256.NewInt(1).SetBytes(test.expectedData), &ctxt)
```
HerbertJordan, line 256: You keep initializing an `uint256.Int` with a value of 0 or 1 for no reason. A cleaner way of handing a situation where you need an instance but the value is irrelevant could look like this:
```suggestion
			err := m.SetWord(test.offset, new(uint256.Int).SetBytes(test.expectedData), &ctxt)
```

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+			if !errors.Is(err, test.expectedError) {
+				t.Errorf("unexpected error, want: %v, got: %v", test.expectedError, err)
+			}
+			if err == nil {
+				if !bytes.Equal(m.store[test.offset:test.offset+32], test.expectedData) {
```
HerbertJordan, line 261: It would be good to check the full memory content to make sure nothing else got modified.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+	}
+}
+
+func TestMemory_Set_SuccessfulCases(t *testing.T) {
+
+	tests := map[string]struct {
+		size   uint64
+		offset uint64
+	}{
+		"regular": {size: 8, offset: 0},
```
HerbertJordan, line 276: A table driven test for a single test? [Cargo cult](https://en.wikipedia.org/wiki/Cargo_cult_programming).

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+			store:    baseStore,
+			offset:   4,
+			expected: []byte{0x90, 0xab, 0xcd, 0xef, 0, 0, 0}},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			m := NewMemory()
+			m.store = test.store
+			target := test.target
```
HerbertJordan, line 374: Since the `test.target` are reusing the same `baseTarget` slice, the effects of one test may influence other tests.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+			target:   baseTarget,
+			store:    baseStore,
+			offset:   4,
+			expected: []byte{0x90, 0xab, 0xcd, 0xef, 0, 0, 0}},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			m := NewMemory()
+			m.store = test.store
```
HerbertJordan, line 373: Since `test.store` is sharing the same `baseStore` slice, tests may influence each other.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +52,429 @@ func TestExpansionCosts(t *testing.T) {
...
+			old := os.Stdout
+			r, w, _ := os.Pipe()
+			os.Stdout = w
+
+			// Run testing code
+			m.Print()
+			// read the output
+			w.Close()
+			out, _ := io.ReadAll(r)
+			os.Stdout = old
```
HerbertJordan, line 471: Please update the "Print()" function to return a string instead of writing stuff to `Stdout`.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory_test.go
facuMH: I did the setup for multiple cases and then realized one was enough, should have removed it.

go/interpreter/lfvm/memory.go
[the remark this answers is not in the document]
facuMH: renamed to `errInsufficientMemory`

go/interpreter/lfvm/memory_test.go
facuMH: since the function `CopyData` reads from `memory` and writes into `target`, there is no modification to `m.store`.

go/interpreter/lfvm/memory_test.go
simonlechner: Why was this conversation marked as resolved but not taken care of or answered?

facuMH: I was sure I had done it. will fix.

go/interpreter/lfvm/memory.go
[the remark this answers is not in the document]
HerbertJordan: A problem of the function name is that it does not start with a verb. A name like this would be associated to a constant.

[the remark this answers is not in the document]
facuMH: mmmm I see. Well my idea for this is to be a sort of error generator, that can be used to compare with `errors.Is`, or `errors.As`, to be a sort of "group" of errors, and this can not be made a constant. Perhaps it could be `makeInsufficientMemoryErr` instead. wdyt ?

go/interpreter/lfvm/memory.go
```diff
@@ -176,47 +169,35 @@ func (m *Memory) Set(offset, size uint64, value []byte) error {
...
 		if offset+size > m.Len() {
-			return fmt.Errorf("memory too small, size %d, attempted to write %d bytes at %d", m.Len(), size, offset)
+			return errInsufficientMemory(m.Len(), size, offset)
 		}
 		copy(m.store[offset:offset+size], value)
 	}
 	return nil
 }
 
+func errInsufficientMemory(memSize, size, offset uint64) error {
```
LuisPH3, line 179: The naming for this function was commented before. 
I really like the technique, it yields clean code. 
To follow the function naming style, we can rename it to `makeInsuficientMemoryError`  
wdyt?.

LuisPH3 (approved):

LuisPH3 (approved):

go/interpreter/lfvm/memory.go
```diff
@@ -68,68 +69,75 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
 	return fee
 }
 
-func (m *Memory) EnsureCapacity(offset, size uint64, c *context) error {
-	if size <= 0 {
-		return nil
-	}
+// expandMemoryAndCharge tries to expand memory to the given size.
+// If the memory is already large enough or size is 0, it does nothing.
```
HerbertJordan, line 73: This seems to be not true any more. If `size` is `0` it will still grow the memory now if the `offset` is > `m.len()`.

Todo:
- check whether this property is important
- if so, add unit tests for this in `memory_test.go` and re-add support
- if not, remove this documentation

go/interpreter/lfvm/memory.go
```diff
@@ -68,68 +69,75 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
...
+// if not enough gas in the context, it returns an error.
+func (m *Memory) expandMemoryAndCharge(offset, size uint64, c *context) error {
```
HerbertJordan, line 75: Instead of using closures and callbacks this could be simpler implemented by one function calling the other:
```
func (m *Memory) expandMemoryAndCharge(offset, size uint64, c *context) error {
	if size == 0 {
		return nil
	}

	needed := offset + size
	// check overflow
	if needed < offset {
		c.signalError()
		return errGasUintOverflow
	}

	fee := m.expansionCosts(needed)
	if !c.useGas(fee) {
		c.status = statusOutOfGas
		return errOutOfGas
	}
	m.total_memory_cost += fee

	m.expandMemory(needed)
	return nil
}

// expandMemory expands the memory to the given size without checking gas.
func (m *Memory) expandMemory(needed uint64) {
	needed = toValidMemorySize(needed)
	size := m.len()
	if size < needed {
		m.store = append(m.store, make([]byte, needed-size)...)
	}
}
```

go/interpreter/lfvm/memory.go
```diff
@@ -68,68 +69,75 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
...
+		}
 		m.total_memory_cost += fee
-		m.store = append(m.store, make([]byte, size-m.Len())...)
+		m.store = append(m.store, make([]byte, size-m.len())...)
 	}
+	return nil
 }
 
-func (m *Memory) Len() uint64 {
+func (m *Memory) len() uint64 {
```
HerbertJordan, line 120: nit: I would favor the name `length` or `size` just to avoid abbreviation.

go/interpreter/lfvm/memory.go
```diff
@@ -68,68 +69,75 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
...
-	if m.Len() < offset+1 {
-		return fmt.Errorf("memory too small, size %d, attempted to write at position %d", m.Len(), offset)
-	}
 	m.store[offset] = value
 	return nil
 }
 
-func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
-	err := m.EnsureCapacity(offset, 32, c)
+// GetByte returns a byte at the given offset, expanding memory as needed and charging for it.
```
HerbertJordan, line 134: `GetByte`?

go/interpreter/lfvm/memory.go
```diff
@@ -170,60 +178,43 @@ func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
 	return nil
 }
 
-func (m *Memory) Set(offset, size uint64, value []byte) error {
+func (m *Memory) set(offset, size uint64, value []byte) error {
```
HerbertJordan, line 181: Please document how this function is supposed to deal with capacity issues.

go/interpreter/lfvm/memory.go
```diff
@@ -170,60 +178,43 @@ func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
...
+	if err != nil {
+		return err
 	}
-	return nil
+	return m.set(offset, size, value)
 }
 
-func (m *Memory) CopyWord(offset uint64, trg *uint256.Int, c *context) error {
-	err := m.EnsureCapacity(offset, 32, c)
+func (m *Memory) getWord(offset uint64, trg *uint256.Int, c *context) error {
```
HerbertJordan, line 206: Should this return a `uint256.Int` instead (a value, not a pointer) since it is a `get`? Otherwise maybe call it `readWord`?

go/interpreter/lfvm/memory.go
```diff
@@ -170,60 +178,43 @@ func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
...
-		return fmt.Errorf("memory too small, size %d, attempted to read 32 byte at position %d", m.Len(), offset)
-	}
 	trg.SetBytes32(m.store[offset : offset+32])
 	return nil
 }
 
 // Copies data from the memory to the given slice.
-func (m *Memory) CopyData(offset uint64, trg []byte) {
-	if m.Len() < offset {
+func (m *Memory) getData(offset uint64, trg []byte) {
```
HerbertJordan, line 216: also here, maybe `readData`?

HerbertJordan, line 216: Please specify what should happen with out-of-bound data.

go/interpreter/lfvm/memory.go
```diff
@@ -237,48 +228,22 @@ func (m *Memory) CopyData(offset uint64, trg []byte) {
 	}
 }
 
-func (m *Memory) GetSlice(offset, size uint64) []byte {
+func (m *Memory) getSlice(offset, size uint64) []byte {
```
HerbertJordan, line 231: Please document what should happen with out-of-bound data requests and add tests for that of there are none yet.

go/interpreter/lfvm/memory.go
```diff
@@ -43,7 +43,8 @@ const (
 	maxMemoryExpansionSize = 0x1FFFFFFFE0
 )
 
-func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
+// expansionCosts returns the gas cost of expanding the memory to the given size.
+func (m *Memory) expansionCosts(size uint64) tosca.Gas {
```
HerbertJordan, line 47: This function name should have a verb. Something like `getExpansionCosts`.

go/interpreter/lfvm/memory.go
```diff
@@ -68,68 +69,75 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
...
+	if err != nil {
+		return err
 	}
+
 	return nil
 }
 
-func (m *Memory) EnsureCapacityWithoutGas(size uint64) {
+// expandMemory expands the memory to the given size without checking gas.
+func (m *Memory) expandMemory(size uint64) {
```
HerbertJordan, line 98: There was a reason why the old functions have been called `EnsureCapacity` and `EnsureCapacityWithoutGas`. The idea was that the default should be the safe one, the one doing charging gas and expanding memory as required, while any user would have to explicitly write "WithoutGas" to mark that this operation is side-stepping the gas checks.

With the new naming schema, the critical one is named `expandMemory`, which looks correct in any use case unless the reader recalls that there is also a `expandMemoryAndCharge` that is most likely the one you should use in practically every use case that is not test infrastructure.

For those reasons, I think it should be explicitly warned that `expandMemory` is probably not what you want. Maybe by calling it `expandMemoryUnsafe`.

go/interpreter/lfvm/memory_test.go
HerbertJordan, line 1: There seem to be no tests for the `expandMemory` functions. Please tests for those.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory.go
facuMH: mmmm yes, I see your point. What about renaming old `EnsureCapacity` to `expandMemory` and `EnsureCapacityWithoutGas` to `expandMemoryWithoutCharging` ? I rather avoid the `WithoutGas` part, since it needs to add the fee into `m.total_memory_cost`

go/interpreter/lfvm/memory.go
facuMH: this was changed as suggested, so the comment is still valid

go/interpreter/lfvm/memory_test.go
```diff
@@ -41,9 +44,475 @@ func TestExpansionCosts(t *testing.T) {
...
+			err := m.readWord(test.offset, target, &ctxt)
+			if err != nil {
+				t.Fatalf("unexpected error, want: %v, got: %v", nil, err)
+			}
+			if target.Cmp(test.expectedData) != 0 {
+				t.Errorf("unexpected target value, want: %x, got: %x", *test.expectedData, *target)
+			}
+		})
+	}
+}
```
HerbertJordan, line 101: I have been reading this test so many times in the last week ... an every time I get a full mental overload. It takes me forever to get what it tries to do and to I never get full confidence that it does what I think it is doing. One of the problems is the naming of stuff, another that one needs to connect multiple different definitions from within this single function to get what is tested. Yet another, that the test covers so many aspects that are just implicitly taken for granted (e.g. byte-granularity addressing, big endian encoding of integer values in the memory, ...). To cover all of this, this test should be named `TestMemory_GetWord_RetrievesDataAtGivenByteOffsetInProperNumericalInterpretationAndAddsZeroPadding`.

The word-addressing and zero padding could for instance be tested like this:
```
func TestMemory_GetWord_ReadsDataWithByteAddressingAndZeroPadding(t *testing.T) {
	pattern := make([]byte, 32)
	for i := 0; i < 32; i++ {
		pattern[i] = byte(i + 1)
	}

	for i := 0; i <= 32; i++ {
		ctxt := getEmptyContext()
		ctxt.gas = 100

		m := NewMemory()
		m.expandMemory(32)
		copy(m.store, pattern)

		var target uint256.Int
		err := m.getWord(uint64(i), &target, &ctxt)
		if err != nil {
			t.Fatalf("unexpected error, want: %v, got: %v", nil, err)
		}

		// Check that the memory pattern is shifted by i bytes to the left.
		want := [32]byte{}
		copy(want[:], pattern[i:])
		if got := target.Bytes32(); want != got {
			t.Errorf("unexpected value, want: %x, got: %x", want, got)
		}
	}
}
```
This test covers the same property as the test you provided, covers it for all offsets between 0-32, does not require anyone to think about the byte- and word-order `uint256` is using and the effect the left-shift may have on it, and is significantly shorter.

I think you need to focus more on the quality of the tests and their comprehensibility.

HerbertJordan (changes requested):

facuMH:
I will split PR into smaller PRs, changing one function per PR, and implementing the corresponding tests.

PR #667, opened 2024-08-23
LFVM: vm and opcodes tests

facuMH:
This PR adds:
- tests for basic cases of `VM.Run` 
- tests for `HasArguments` function and `opcodes.Strings`

### DIFF
--- a/go/interpreter/lfvm/lfvm_test.go
+++ b/go/interpreter/lfvm/lfvm_test.go
@@ -1,0 +1,92 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package lfvm
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+)
+
+func TestVm_Run(t *testing.T) {
+
+	tests := map[string]struct {
+		code                  []byte
+		codeHash              tosca.Hash
+		revision              tosca.Revision
+		withSuperInstructions bool
+		withCodeCache         bool
+		expectedResult        tosca.Result
+		expectedError         error
+	}{
+		"empty code": {
+			revision: tosca.R13_Cancun,
+			expectedResult: tosca.Result{Success: true, GasLeft: 1000000,
+				GasRefund: 0, Output: []byte{}},
+		},
+		"invalid code": {
+			code:     []byte{0x0C},
+			revision: tosca.R13_Cancun,
+			expectedResult: tosca.Result{Success: false, GasLeft: 0,
+				GasRefund: 0, Output: []byte{}},
+		},
+		"newer unsupported revision": {
+			revision: newestSupportedRevision + 1,
+			expectedResult: tosca.Result{Success: false, GasLeft: 0,
+				GasRefund: 0, Output: []byte{}},
+			expectedError: &tosca.ErrUnsupportedRevision{Revision: newestSupportedRevision + 1},
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			params := tosca.Parameters{
+				Gas:      1000000,
+				Code:     test.code,
+				CodeHash: &test.codeHash,
+				BlockParameters: tosca.BlockParameters{
+					Revision: test.revision,
+				},
+			}
+
+			vm := &VM{
+				with_super_instructions: test.withSuperInstructions,
+				no_code_cache:           !test.withCodeCache,
+			}
+
+			result, err := vm.Run(params)
+			if err != test.expectedError && strings.Compare(err.Error(), test.expectedError.Error()) != 0 {
+				t.Fatalf("unexpected error: want %v but got %v", test.expectedError, err)
+			}
+			if result.Success != test.expectedResult.Success {
+				t.Errorf("unexpected result, want %v but got %v",
+					test.expectedResult.Success, result.Success)
+			}
+			if result.GasLeft != test.expectedResult.GasLeft {
+				t.Errorf("unexpected GasLeft, want %v but got %v",
+					test.expectedResult.GasLeft, result.GasLeft)
+			}
+			if result.GasRefund != test.expectedResult.GasRefund {
+				t.Errorf("unexpected GasRefund, want %v but got %v",
+					test.expectedResult.GasRefund, result.GasRefund)
+			}
+			if !bytes.Equal(result.Output, test.expectedResult.Output) {
+				t.Errorf("unexpected Output, want %v but got %v",
+					test.expectedResult.Output, result.Output)
+			}
+
+		})
+	}
+
+}
--- a/go/interpreter/lfvm/lfvm_test.go
+++ b/go/interpreter/lfvm/lfvm_test.go
@@ -66,7 +66,7 @@ func TestVm_Run(t *testing.T) {
 				ConversionConfig: ConversionConfig{
 					WithSuperInstructions: false,
 				},
-				NoShaCache: true,
+				WithShaCache: true,
 			})
 			if err != nil {
 				t.Fatalf("failed to create vm: %v", err)
--- a/go/interpreter/lfvm/lfvm_test.go
+++ b/go/interpreter/lfvm/lfvm_test.go
@@ -59,9 +59,14 @@ func TestVm_Run(t *testing.T) {
 				},
 			}
 
-			vm := &VM{
-				with_super_instructions: test.withSuperInstructions,
-				no_code_cache:           !test.withCodeCache,
+			vm, err := NewVm(Config{
+				ConversionConfig: ConversionConfig{
+					WithSuperInstructions: test.withSuperInstructions,
+				},
+				NoShaCache: !test.withCodeCache,
+			})
+			if err != nil {
+				t.Fatalf("failed to create vm: %v", err)
 			}
 
 			result, err := vm.Run(params)
--- a/go/interpreter/lfvm/lfvm_test.go
+++ b/go/interpreter/lfvm/lfvm_test.go
@@ -16,6 +16,7 @@ import (
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
 )
 
 func TestVm_Run(t *testing.T) {
@@ -34,8 +35,8 @@ func TestVm_Run(t *testing.T) {
 				GasRefund: 0,
 				Output:    []byte{}},
 		},
-		"invalid code": {
-			code:     []byte{0x0C},
+		"invalid opcode": {
+			code:     []byte{byte(vm.INVALID)},
 			revision: tosca.R13_Cancun,
 			expectedResult: tosca.Result{
 				Success:   false,
--- a/go/interpreter/lfvm/lfvm_test.go
+++ b/go/interpreter/lfvm/lfvm_test.go
@@ -41,9 +41,7 @@ func TestVm_Run(t *testing.T) {
 				GasRefund: 0, Output: []byte{}},
 		},
 		"newer unsupported revision": {
-			revision: newestSupportedRevision + 1,
-			expectedResult: tosca.Result{Success: false, GasLeft: 0,
-				GasRefund: 0, Output: []byte{}},
+			revision:      newestSupportedRevision + 1,
 			expectedError: &tosca.ErrUnsupportedRevision{Revision: newestSupportedRevision + 1},
 		},
 	}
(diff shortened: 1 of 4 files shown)


go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"slices"
+	"testing"
+)
+
+func TestOpcode_String(t *testing.T) {
+	tests := []struct {
+		name   string
+		opcode OpCode
+	}{
```
LuisPH3, line 22: Because every print returns a new string, I am afraid we should test all of them.
(the problem being here that someone misspelled an opcode)
Can we do this with reflection?

go/interpreter/lfvm/lfvm_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+			vm := &VM{
+				with_super_instructions: test.withSuperInstructions,
+				no_code_cache:           !test.withCodeCache,
+			}
+
+			result, err := vm.Run(params)
+			if err != test.expectedError && strings.Compare(err.Error(), test.expectedError.Error()) != 0 {
+				t.Fatalf("unexpected error: want %v but got %v", test.expectedError, err)
+			}
```
LuisPH3, line 71: If error was expected, we should skip the rest of the test. 
Code changes that return a different payload in an error scenario should not trigger test changes.

LuisPH3 (approved):

LuisPH3 (changes requested):
Please have a look at the comments and we can discuss them.

LuisPH3 (approved):

LuisPH3 (dismissed):

go/interpreter/lfvm/lfvm_test.go
```diff
@@ -0,0 +1,93 @@
...
+				},
+			}
+
+			vm := &VM{
+				with_super_instructions: test.withSuperInstructions,
+				no_code_cache:           !test.withCodeCache,
+			}
+
+			result, err := vm.Run(params)
+			if err != test.expectedError && strings.Compare(err.Error(), test.expectedError.Error()) != 0 {
```
simonlechner, line 67:
```suggestion
			if err != test.expectedError && errors.Is(err, test.expectedError) {
```

go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"slices"
+	"testing"
+)
+
+func TestOpcode_String(t *testing.T) {
+	for i := 0; i < 256; i++ {
+		op := OpCode(i)
+		want, found := op_to_string[op]
+		got := op.String()
```
simonlechner, line 23: The `String` function uses the `op_to_string` function internally. This test is testing whether the same function returns the same result when called with the same argument.

simonlechner (changes requested):
I think the opcode string test should be redesigned.

simonlechner (dismissed):
Thank you for the changes.

go/interpreter/lfvm/lfvm_test.go
```diff
@@ -0,0 +1,97 @@
...
+		codeHash              tosca.Hash
+		revision              tosca.Revision
+		withSuperInstructions bool
+		withCodeCache         bool
+		expectedResult        tosca.Result
+		expectedError         *tosca.ErrUnsupportedRevision
+	}{
+		"empty code": {
+			revision: tosca.R13_Cancun,
+			expectedResult: tosca.Result{Success: true, GasLeft: 1000000,
```
LuisPH3, line 34: nit: multiline, easier to review

go/interpreter/lfvm/lfvm_test.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+
+			vm := &VM{
+				with_super_instructions: test.withSuperInstructions,
+				no_code_cache:           !test.withCodeCache,
+			}
+
+			result, err := vm.Run(params)
+			if err != nil || test.expectedError != nil {
+				if !errors.As(err, &test.expectedError) {
```
LuisPH3, line 69: Double pointer?  `expectedError` is already a pointer

go/interpreter/lfvm/lfvm_test.go
```diff
@@ -0,0 +1,97 @@
...
+				},
+			}
+
+			vm := &VM{
+				with_super_instructions: test.withSuperInstructions,
+				no_code_cache:           !test.withCodeCache,
+			}
+
+			result, err := vm.Run(params)
+			if err != nil || test.expectedError != nil {
```
LuisPH3, line 68: This code is cryptic, I think the new pattern with the error constructor function is better than the error type + `errors.As`
The new method allows us to use `errors.Is without checking for nil before in the both branches: yields straight forward code:

```go
   expecedErr := unsportedRevision(x) // or nil
   // ....
   if !errors.Is(err, expectedErr) { 
        // this branch satisfies: 
        //  - err is nil && expectedErr is not
        //  - err is not nil && err is != expectedErr
   }
   // jump is not taken if both are nil !
```

go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+		}
+		got := op.String()
+		if got != want {
+			t.Errorf("expected %s, got %s", want, got)
+		}
+	}
+}
+
+func findIndex(slice []OpCode, element OpCode) int {
```
LuisPH3, line 77: [slices.Index](https://pkg.go.dev/slices#Index) Or
[slices.IndexFunc](https://pkg.go.dev/slices#IndexFunc)

LuisPH3 (approved):

HerbertJordan (commented):
While it is a step toward increase code coverage, I think we should first finish up the refactoring of the operations and the interpreter and then see what of this code is still left that needs to be tested.

go/interpreter/lfvm/lfvm_test.go
facuMH: I thought so as well, but without the that ref I get the following error `second argument to errors.As must be a non-nil pointer to either a type that implements error, or to any interface type`.

go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,99 @@
...
+		if v == element {
+			return i
+		}
+	}
+	return 256
+}
+
+func TestOpcode_HasArgument(t *testing.T) {
+
+	haveArgument := []OpCode{DATA, JUMP_TO, PUSH2_JUMP, PUSH2_JUMPI, PUSH1_PUSH4_DUP3}
```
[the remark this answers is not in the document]
facuMH: should this have been detected before ? it is not in `HasArguments` func either.

LuisPH3 (approved):

LuisPH3 (dismissed):

go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"regexp"
+	"slices"
+	"testing"
+)
+
+func TestOpcode_String(t *testing.T) {
+	validName := regexp.MustCompile(`^0x00[0-9A-Fa-f]{2}$`)
+	for i := 0; i < 256; i++ {
+		op := OpCode(i)
+		if !validName.MatchString(op.String()) && op > NUM_OPCODES {
```
HerbertJordan, line 23: This test only fails if the tested operation is bigger than `NUM_OPCODES` ?

go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"regexp"
+	"slices"
+	"testing"
+)
+
+func TestOpcode_String(t *testing.T) {
+	validName := regexp.MustCompile(`^0x00[0-9A-Fa-f]{2}$`)
+	for i := 0; i < 256; i++ {
```
HerbertJordan, line 21: The lfvm OpCodes not necessarily end at 256.

HerbertJordan (changes requested):

go/interpreter/lfvm/opcode_test.go
facuMH: no, if the teste op is bigger than `NUM_OPCODES` and it is not printed in the fashion of `0x00XX`, which should be the case for all opcodes bigger than `NUM_OPCODES`

HerbertJordan: Got it now - this test is checking that for everything that is not a defined instruction the hex value is printed. This is not what would be expected for the only test that checks the string.

Please fix the name of the test accordingly to state what this test is covering. Also, in this case, why `i` not starting with `NUM_OPCODES`?

HerbertJordan (changes requested):

go/interpreter/lfvm/lfvm_test.go
```diff
@@ -0,0 +1,106 @@
...
+		"empty code": {
+			revision: tosca.R13_Cancun,
+			expectedResult: tosca.Result{
+				Success:   true,
+				GasLeft:   1000000,
+				GasRefund: 0,
+				Output:    []byte{}},
+		},
+		"invalid code": {
+			code:     []byte{0x0C},
```
HerbertJordan, line 38: Technically this code is not invalid. It can be executed. It is just the case that the first instruction is an invalid instruction.

HerbertJordan, line 38: To make the intention clearer, please use the explicitly defined `INVALID` op-code
```suggestion
			code:     []byte{vm.INVALID},
```

HerbertJordan (changes requested):

go/interpreter/lfvm/opcode_test.go
facuMH: `in this case, why i not starting with NUM_OPCODES?` because I also intended to test that for all the "known" opcodes, a string not resembling `0x00XX` is provided.

LuisPH3 (dismissed):

go/interpreter/lfvm/opcode.go
```diff
@@ -444,6 +444,9 @@ func (o OpCode) HasArgument() bool {
 		return true
 	case PUSH1_PUSH4_DUP3:
 		return true
+	case PC:
+		return true
```
LuisPH3, line 448: Can you explain this to me? 
Why wasn't it needed before?

go/interpreter/lfvm/opcode_test.go
LuisPH3: To avoid this kind of logic errors, I would try to make more clear what range is being tested:
```
if op is named { // where named is the defined opcodes only (including INVALID) 
    t.Skip()
}
// do the test for invalid opcodes. 
```

It happens to be that now all lfvm opcodes are < NUM_OPCODES, but this may change.

go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+}
+
+func TestOpcode_NamedOpcodes(t *testing.T) {
+	validName := regexp.MustCompile(`^0x00[0-9A-Fa-f]{2}$`)
+	for i := 0; i < math.MaxUint16; i++ {
+		op := OpCode(i)
+		// NUM_EXECUTABLE_OPCODES is a special case, it does not have a string name
+		if validName.MatchString(op.String()) && op < NUM_OPCODES && op != NUM_EXECUTABLE_OPCODES {
```
LuisPH3, line 35: As mentioned before, the contiguous property of opcodes may change.
Not sure if this is easy to model somehow else at this point.

go/interpreter/lfvm/opcode_test.go
```diff
@@ -0,0 +1,54 @@
...
+			t.Errorf("Invalid print for op %v (%d)", op, i)
+		}
+	}
+}
+
+func TestOpcode_HasArgument(t *testing.T) {
+
+	haveArgument := []OpCode{PC, DATA, JUMP_TO, PUSH2_JUMP, PUSH2_JUMPI, PUSH1_PUSH4_DUP3}
```
LuisPH3, line 43: You miss here PC. which you added in the definition file. Please check if it is right and if it is, add it to the test.

LuisPH3 (changes requested):

go/interpreter/lfvm/opcode.go
facuMH: it's not that it was not needed, perhaps it was simply not noticed that it was missing. 
This should be here since when the code is converted the position where the instruction is called, is saved as argument. Hence when printing the converted code (sole use of `HasArgument`) we want to also show the value of this position.

go/interpreter/lfvm/opcode_test.go
facuMH: PC is first instruction tested.

go/interpreter/lfvm/opcode_test.go
facuMH: as an answer to both your comments, I cannot make a test for a property that might change in the future. I can only make a test for existing code. Currently the constant `NUM_OPCODES` and `NUM_EXECUTABLE_OPCODES` are the way to distinguish whether an opcodes is valid or executable or neither, and this dictates what style of printing it should use. Given that there is also no bidirectional mapping to `tosca.OpCodes`. I'm not sure this can be tested otherwise

facuMH:
This PR has been rendered obsolete by newer refactors.

PR #668, opened 2024-08-26
Fix Differential fuzzing error found in nightly builds

LuisPH3:
Memory release was releasing a nil reference in scenarios that should be skipped.


facuMH (approved):
nice catch

HerbertJordan (approved):

simonlechner (approved):

PR #678, opened 2024-08-26
LFVM: remove unused shadow db

facuMH:
This PR removes `shadowStateDB` which is a testing implementation of `StateDB` interface that was never used.


HerbertJordan (approved):

LuisPH3 (approved):
Off with the head.

simonlechner (approved):

PR #696, opened 2024-08-26
Add ExecutionTxContext and RunResult

LorenzSchueler:
- add `ExecutionTxContext` (like `evmc::ExecutionTxContext` but with pointer and length fields converted to slices)
- add `RunResult`

### DIFF
--- a/rust/src/interpreter/stack.rs
+++ b/rust/src/interpreter/stack.rs
@@ -6,11 +6,7 @@ use crate::types::u256;
 pub struct Stack(Vec<u256>);
 
 impl Stack {
-    pub fn new() -> Self {
-        Self(Vec::new())
-    }
-
-    pub fn from_vec(inner: Vec<u256>) -> Self {
+    pub fn new(inner: Vec<u256>) -> Self {
         Self(inner)
     }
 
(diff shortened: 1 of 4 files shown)


rust/src/interpreter/stack.rs
```diff
@@ -6,11 +6,7 @@ use crate::types::u256;
 pub struct Stack(Vec<u256>);
 
 impl Stack {
-    pub fn new() -> Self {
-        Self(Vec::new())
-    }
-
-    pub fn from_vec(inner: Vec<u256>) -> Self {
+    pub fn new(inner: Vec<u256>) -> Self {
```
LuisPH3, line 9: I liked the empty constructor :(

I think the reason why you removed it is that you got tired of repeating code. This has to do with ergonomics. 
IMO, new (for empty) and new_with(slice)  would be less code (no vector). Yes, they hide a vector copy, data is so small that it could be optimized away, and overhead in tests will be negligible, I would prefer ergonomics to premature optimization in this case.

LuisPH3 (dismissed):

rust/src/interpreter/stack.rs
LorenzSchueler: I removed it because after some refactoring of the next PRs the empty constructor was no longer needed. And in the tests I think it is cleaner to provide an empty vector to the (few) cases where an empty Stack is needed than to use a different constructor.

rust/src/interpreter/tx_context.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// The array of transaction initcodes (TXCREATE).
+    pub initcodes: &'a [evmc_tx_initcode],
+}
+
+impl<'a> From<&'a evmc_vm::ExecutionTxContext> for ExecutionTxContext<'a> {
+    fn from(context: &'a evmc_tx_context) -> Self {
+        let blob_hashes = if context.blob_hashes.is_null() {
+            &[]
+        } else {
+            assert_ne!(context.blob_hashes_count, 0);
```
HerbertJordan, line 47: We try to avoid assertions in the code for reasons we can discuss offline one day. In this case and the one below, I assume those could be easily avoided by just handling the empty case and the null case the same, right?

HerbertJordan (dismissed):

rust/src/interpreter/tx_context.rs
LorenzSchueler: Yes, that is true. However the expectation is that those slices are either empty (pointer is null and length is 0) or non empty (non zero pointer and length > 0). Having pointer = null and length > 0 violates this assumption and (at least in my mind) is a bug on the Go side. But I guess you prefer continuing with an empty slice over crashing. I'm fine with that too.

LuisPH3: I agree with @LorenzSchueler, receiving a null ptr from the FFI is unexpected and can be though of as a bug. Nevertheless, there is a critical responsibility on not panicking. I see two ways out:
- return an error right away. 
- continue in safety and let someone else figure out that there was an error.

LorenzSchueler: But if you use TryFrom what do you do in the calling rust code with this error? There is now proper way to continue but also no way to signal to Go that the expectations of the evmc interface are violated. In the ffi code which I adapted from the evmc rust bindings they just abort if something like this happens. Even on the go side there is (as far as I know) now way to deal with this because it is a bug in the go code itself.

LorenzSchueler: In my mind the only viable options are either to panic/ abort or to simply ignore this error completely and continue with an empty slice. The former would at least make sure that transactions are not processed incorrectly.

LorenzSchueler: Should I merge this as is or return an empty slice?

LuisPH3: Let's return the empty slice for now, just to prevent the inherent panic.

LuisPH3 (approved):

PR #699, opened 2024-08-27
LFVM: remove geth dependencies

### LINKED ISSUE #410: lfvm: remove remaining go-ethereum dependencies from the LFVM code
Right now, the LFVM interpreter still contains a few dependencies to go-ethereum packages. In particular, for its definition of the OpCodes and some other constants, dependencies are still present. Those should be removed.

facuMH:
This PR  removes remaining geth dependencies in lfvm by defining our own constants.
This PR fixes #410

### DIFF
--- a/go/interpreter/lfvm/gas.go
+++ b/go/interpreter/lfvm/gas.go
@@ -15,11 +15,41 @@ import (
 	"fmt"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
-	"github.com/ethereum/go-ethereum/params" // < TODO: remove
 	"github.com/holiman/uint256"
 )
 
-const UNKNOWN_GAS_PRICE = 999999
+const (
+	CallNewAccountGas    = 25000 // Paid for CALL when the destination address didn't exist prior.
+	CallValueTransferGas = 9000  // Paid for CALL when the value transfer is non-zero.
+	CallStipend          = 2300  // Free gas given at beginning of call.
+
+	ColdSloadCostEIP2929         = 2100 // Cost of cold SLOAD after EIP 2929
+	ColdAccountAccessCostEIP2929 = 2600 // Cost of cold account access after EIP 2929
+
+	// CreateBySelfdestructGas is used when the refunded account is one that does
+	// not exist. This logic is similar to call.
+	// Introduced in Tangerine Whistle (Eip 150)
+	CreateBySelfdestructGas = 25000
+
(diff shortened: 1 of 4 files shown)


go/interpreter/lfvm/gas.go
```diff
@@ -297,7 +327,7 @@ func gasSStore(c *context) (tosca.Gas, error) {
 //     2.2.2.2. Otherwise, add SSTORE_RESET_GAS - SLOAD_GAS gas to refund counter.
 func gasSStoreEIP2200(c *context) (tosca.Gas, error) {
 	// If we fail the minimum gas availability invariant, fail (0)
-	if c.gas <= tosca.Gas(params.SstoreSentryGasEIP2200) {
+	if c.gas <= tosca.Gas(SstoreSentryGasEIP2200) {
```
HerbertJordan, line 330: The `tosca.Gas(...)` cast would not really be needed here since constants are of the exotic "untyped int" type.

However, we could enforce that all those constants are Gas values (and thereby avoid accidental miss-use) by defining them all as gas values like this:
```
SstoreResetGasEIP2200      tosca.Gas = 5000
```
or
```
SstoreResetGasEIP2200      = tosca.Gas(5000)
```
Wdyt?

go/interpreter/lfvm/stack.go
```diff
@@ -15,12 +15,13 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/ethereum/go-ethereum/params"
 	"github.com/holiman/uint256"
 )
 
 var staticStackBoundry = [NUM_OPCODES]InstructionStack{}
 
+const StackLimit uint64 = 1024 // Maximum size of VM stack allowed.
```
HerbertJordan, line 23: Should be private until needed to be public.

HerbertJordan, line 23: If you omit the `uint64` type here you do not need the `int` cast below.

HerbertJordan (changes requested):

go/interpreter/lfvm/stack.go
facuMH: I was actually thinking of making it a tosca constant that can be used by lfvm, ct and wherever else needed. But maybe that should a separate PR. For now I will just make it private

simonlechner (approved):

LuisPH3 (approved):
I have concerns about the maintainability of this, we would have to skim through the code now to find where do we need to change the parameter.

PR #700, opened 2024-08-28
LFVM: gas tests 

facuMH:
This PR adds tests to verify that a few special cases of gas cost calculations are correctly handled.

### DIFF
--- a/go/interpreter/lfvm/gas.go
+++ b/go/interpreter/lfvm/gas.go
@@ -99,9 +99,13 @@ func getStaticGasPriceInternal(op OpCode) tosca.Gas {
 	if SWAP1 <= op && op <= SWAP16 {
 		return 3
 	}
+	// this range covers: LT, GT, SLT, SGT, EQ, ISZERO, AND, OR,
+	//					  XOR, NOT, BYTE, SHL, SHR, SAR
 	if LT <= op && op <= SAR {
 		return 3
 	}
+	// this range covers: COINBASE, TIMESTAMP, NUMBER, DIFFICULTY/PREVRANDO,
+	// 					GAS, GASLIMIT, CHAINID
 	if COINBASE <= op && op <= CHAINID {
 		return 2
 	}
--- a/go/interpreter/lfvm/gas.go
+++ b/go/interpreter/lfvm/gas.go
@@ -11,9 +11,6 @@
 package lfvm
 
 import (
-	"errors"
-	"fmt"
-
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	"github.com/holiman/uint256"
 )
@@ -51,6 +48,12 @@ const (
 	UNKNOWN_GAS_PRICE = 999999
 )
 
+// Gas errors definitions
+const (
+	errNotEnoughGasReentrancy  = ConstError("not enough gas for reentrancy sentry")
+	errAddressNotFoundInSstore = ConstError("address was not present in access list during sstore op")
+)
+
 var static_gas_prices = [NUM_OPCODES]tosca.Gas{}
 var static_gas_prices_berlin = [NUM_OPCODES]tosca.Gas{}
 
@@ -62,15 +65,6 @@ func init() {
 		static_gas_prices_berlin[i] = gp
 	}
 	initBerlinGasPrice()
-
-	for i := 0; i < int(NUM_EXECUTABLE_OPCODES); i++ {
-		if static_gas_prices[i] == UNKNOWN_GAS_PRICE {
-			panic(fmt.Sprintf("Gas price for %v is unknown", OpCode(i)))
-		}
-		if static_gas_prices_berlin[i] == UNKNOWN_GAS_PRICE {
-			panic(fmt.Sprintf("Berlin gas price for %v is unknown", OpCode(i)))
-		}
-	}
 }
 
 func initBerlinGasPrice() {
@@ -210,8 +204,6 @@ func getStaticGasPriceInternal(op OpCode) tosca.Gas {
 		return 2
 	case MCOPY:
 		return 3
-	case GAS:
-		return 2
 	case LOG0:
 		return 375
 	case LOG1:
@@ -237,8 +229,6 @@ func getStaticGasPriceInternal(op OpCode) tosca.Gas {
 	case STOP:
 		return 0
 	case REVERT:
-		return 0
-	case INVALID:
 		return 0
 	case DELEGATECALL:
 		return 700 // Should be 100 according to evm.code
@@ -329,7 +319,7 @@ func gasSStoreEIP2200(c *context) (tosca.Gas, error) {
 	// If we fail the minimum gas availability invariant, fail (0)
 	if c.gas <= SstoreSentryGasEIP2200 {
 		c.status = statusOutOfGas
-		return 0, errors.New("not enough gas for reentrancy sentry")
+		return 0, errNotEnoughGasReentrancy
 	}
 	// Gas sentry honoured, do the actual gas calculation based on the stored value
 	var (
@@ -380,7 +370,7 @@ func gasSStoreEIP2929(c *context) (tosca.Gas, error) {
 	// If we fail the minimum gas availability invariant, fail (0)
 	if c.gas <= SstoreSentryGasEIP2200 {
 		c.status = statusOutOfGas
-		return 0, errors.New("not enough gas for reentrancy sentry")
+		return 0, errNotEnoughGasReentrancy
 	}
 	// Gas sentry honoured, do the actual gas calculation based on the stored value
 	var (
@@ -395,7 +385,7 @@ func gasSStoreEIP2929(c *context) (tosca.Gas, error) {
 	if addrPresent, slotPresent := c.context.IsSlotInAccessList(c.params.Recipient, slot); !slotPresent {
 		if !addrPresent {
 			c.status = statusError
-			return 0, errors.New("address was not present in access list during sstore op")
+			return 0, errAddressNotFoundInSstore
 		}
 		cost = ColdSloadCostEIP2929
 		// If the caller cannot afford the cost, this change will be rolled back
--- a/go/interpreter/lfvm/gas_test.go
+++ b/go/interpreter/lfvm/gas_test.go
@@ -59,3 +59,17 @@ func TestGas_CallGasCalculation(t *testing.T) {
 		})
 	}
 }
+
+func TestGas_CheckGasCostsForAllExecutableOpcodes(t *testing.T) {
+	for i := 0; i < int(NUM_OPCODES); i++ {
+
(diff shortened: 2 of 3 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -17,6 +17,7 @@ import (
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/ethereum/go-ethereum/params"
```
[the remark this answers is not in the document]
facuMH: indeed. Maybe this PR should have been a follow up to that one.

go/interpreter/lfvm/gas_test.go
```diff
@@ -59,3 +59,17 @@ func TestGas_CallGasCalculation(t *testing.T) {
...
 }
+
+func TestGas_CheckGasCostsForAllExecutableOpcodes(t *testing.T) {
+	for i := 0; i < int(NUM_OPCODES); i++ {
+
+		if static_gas_prices[i] == UNKNOWN_GAS_PRICE &&
+			op_2_op[i] != INVALID {
+			t.Errorf("gas price for %v is unknown", OpCode(i))
+		}
+		if static_gas_prices_berlin[i] == UNKNOWN_GAS_PRICE &&
```
simonlechner, line 70: Since the test looks so similar, I am wondering why it is required to have a separate list of gas prices for the Berlin revision?

go/interpreter/lfvm/instructions_test.go
```diff
@@ -870,3 +873,189 @@ func TestExpansionCostOverflow(t *testing.T) {
...
+
+	tests := map[string]struct {
+		runtimeSetup   func(*tosca.MockRunContext)
+		revision       tosca.Revision
+		gas            tosca.Gas
+		expectedStatus Status
+	}{
+		"berlin-enough-gas": {
+			runtimeSetup: func(runContext *tosca.MockRunContext) {
+				runContext.EXPECT().IsSlotInAccessList(gomock.Any(), gomock.Any()).Return(true, true).AnyTimes()
```
simonlechner, line 887: In all these testcases the `AnyTimes()` can be removed.
The `minimum-gas-invariant` case does not need to prime the mock as it does not call `IsSlotInAccessList`.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -870,3 +873,189 @@ func TestExpansionCostOverflow(t *testing.T) {
...
+			if ctxt.status != test.expectedStatus {
+				t.Errorf("unexpected status after call, wanted %v, got %v", test.expectedStatus, ctxt.status)
+			}
+		})
+	}
+}
+
+func TestSelfDestruct(t *testing.T) {
+
+	tests := map[string]struct {
```
simonlechner, line 969: Similar to the other test, the `AnyTimes()` can be removed for all cases by removing the following lines:
`berlin-regular`: `GetBalance`
`london-address-not-in-list`: `HasSelfDestructed`
`london-create-new-account`: `HasSelfDestructed`

simonlechner (commented):
I added some questions as well as some suggestions regarding mocking.
I think the `AnyTimes()` function should not be used in specific cases as it takes away information which functions are really called.

go/interpreter/lfvm/gas.go
```diff
@@ -210,8 +204,6 @@ func getStaticGasPriceInternal(op OpCode) tosca.Gas {
 		return 2
 	case MCOPY:
 		return 3
-	case GAS:
-		return 2
```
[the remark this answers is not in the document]
facuMH: in `lfvm/opcodes.go#186`, `GAS` is defined, between `COINBASE` and `CHAINID`. There is an if in `gas.go#105` that already covers this range of opcodes.

go/interpreter/lfvm/gas_test.go
facuMH: this is only testing that all `INVALID` opcodes have `UNKNOWN_GAS_PRICE`. The two maps are needed because there 10 instructions with a different gas cost.

go/interpreter/lfvm/gas.go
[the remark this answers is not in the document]
simonlechner: I saw that line but I didn't know that `Gas` was part of that range, maybe add a comment with the opcodes included in that range. Also for the range from `LT` to `SAR`.

go/interpreter/lfvm/gas.go
```diff
@@ -329,7 +319,7 @@ func gasSStoreEIP2200(c *context) (tosca.Gas, error) {
 	// If we fail the minimum gas availability invariant, fail (0)
 	if c.gas <= SstoreSentryGasEIP2200 {
```
LuisPH3, line 320: This check is done again in line 381. I don't see at this moment where is the gas modified in between. 
- What does the coverage say about these two ifs (320 and 381)?
- Would it be equivalent to have only one of the two?

go/interpreter/lfvm/gas_test.go
```diff
@@ -59,3 +59,17 @@ func TestGas_CallGasCalculation(t *testing.T) {
 		})
 	}
 }
+
+func TestGas_CheckGasCostsForAllExecutableOpcodes(t *testing.T) {
+	for i := 0; i < int(NUM_OPCODES); i++ {
+
+		if static_gas_prices[i] == UNKNOWN_GAS_PRICE &&
+			op_2_op[i] != INVALID {
```
LuisPH3, line 67: nit: early bail out, `if op_2_op[i] == invalid => continue`

LuisPH3: nit: op_2_op is cryptic and not a good name

go/interpreter/lfvm/instructions_test.go
```diff
@@ -870,3 +873,184 @@ func TestExpansionCostOverflow(t *testing.T) {
...
+
+func TestSStoreGasCost(t *testing.T) {
+
+	tests := map[string]struct {
+		runtimeSetup   func(*tosca.MockRunContext)
+		revision       tosca.Revision
+		gas            tosca.Gas
+		expectedStatus Status
+	}{
+		"berlin-enough-gas": {
```
LuisPH3, line 885: nit: use spaces or underscores

(go test names are going to replace spaces with underscores, and using `-` will end up with names like 
`test/bla_bla-bla`)

go/interpreter/lfvm/instructions_test.go
```diff
@@ -870,3 +873,184 @@ func TestExpansionCostOverflow(t *testing.T) {
...
+				context: runContext,
+				gas:     test.gas,
+			}
+			ctxt.revision = test.revision
+
+			// Prepare stack arguments.
+			ctxt.stack.push(uint256.NewInt(1))
+			ctxt.stack.push(uint256.NewInt(1))
+
+			runContext.EXPECT().SetStorage(gomock.Any(), gomock.Any(), gomock.Any()).Return(tosca.StorageAssigned).AnyTimes()
```
LuisPH3, line 952: I guess that the number of invocations is a property of the scenario.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -870,3 +873,184 @@ func TestExpansionCostOverflow(t *testing.T) {
...
+			runContext.EXPECT().GetStorage(gomock.Any(), gomock.Any()).Return(tosca.Word{}).AnyTimes()
```
LuisPH3, line 953: idem: I guess that the number of invocations is a property of the scenario.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -870,3 +873,184 @@ func TestExpansionCostOverflow(t *testing.T) {
...
+			runContext.EXPECT().GetCommittedStorage(gomock.Any(), gomock.Any()).Return(tosca.Word{}).AnyTimes()
```
LuisPH3, line 954: idem: I guess that the number of invocations is a property of the scenario.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -870,3 +873,184 @@ func TestExpansionCostOverflow(t *testing.T) {
...
+				memory:  NewMemory(),
+				context: runContext,
+			}
+			ctxt.revision = test.revision
+
+			// Prepare stack arguments.
+			ctxt.stack.push(uint256.NewInt(1))
+			ctxt.gas = test.gas
+
+			runContext.EXPECT().SelfDestruct(gomock.Any(), gomock.Any()).Return(true).AnyTimes()
```
LuisPH3, line 1044: idem: I guess that the number of invocations is a property of the scenario.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -218,10 +218,11 @@ func TestCallChecksBalances(t *testing.T) {
...
-		context: runContext,
-		stack:   NewStack(),
-		memory:  NewMemory(),
-		gas:     1 << 20,
+		context:  runContext,
+		stack:    NewStack(),
+		memory:   NewMemory(),
+		gas:      1 << 20,
+		revision: tosca.R09_Berlin,
 	}
```
LuisPH3, line 226: - add tests where account does not exist.
- add test where address is not in access list.

go/interpreter/lfvm/gas.go
facuMH: they are in 2 different functions, `gasSStoreEIP2200(c *context)` ends in line 371, `gasSStoreEIP2929(c *context)` starts in line 373

go/interpreter/lfvm/gas_test.go
facuMH: `vmOp_to_lfvmOp` seems like a better, natural name, but a little more visually noisy, but I can change it to that if you approve.

LuisPH3: I think it is still not very good. It out of the scope of this PR. we can think about it later on.

HerbertJordan, line 67: As mentioned in other PRs, `op_2_op` is an internal utility of the converter and not supposed to be used as an universal mapping between EVM and LFVM op codes.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -208,43 +208,85 @@ func TestPush4(t *testing.T) {
 }
 
 func TestCallChecksBalances(t *testing.T) {
```
HerbertJordan, line 210: I am having a hard time figuring out what this test does. Please improve the name of the test, also for other cases.

LuisPH3 (dismissed):

LuisPH3 (dismissed):

facuMH:
This PR was splitted into multipler smaller ones

Issue #686, opened 2024-08-26
Remove the hash cache's dependency to the LFVM context


HerbertJordan:
Right now, the hash cache requires the LFVM context to be passed whenever a hash is to be computed ([here](https://github.com/Fantom-foundation/Tosca/blob/ff084e3912f85e8f2b732d1d61664b3869b80eb6/go/interpreter/lfvm/hash_cache.go#L106)). This is unnecessary, once #685 is added and should be removed.

PR #701, opened 2024-08-28
LFVM: remove context from hash cache

facuMH:
This PR removes the context from the `hash_cache.hash` calls, and replaces the old `getHash` for corresponding calls to the new `keccak256` hash function. 
This PR contributes to #686


HerbertJordan (approved):
I am surprised (maybe not) that there are no unit tests that need to be changed :)

simonlechner (approved):

LuisPH3 (approved):

PR #706, opened 2024-08-28
Cleanup of LFVM's code converter

HerbertJordan:
This PR refactors the implementation of LFVM's code conversion infrastructure by:
- introducing an explicit `Converter` type handling the conversions and the caching of results. This
   - eliminates the global code cache
   - fixes the problem that codes in the cache could be a mixture of codes with and without super instructions
- cleanup of the configuration flags of the converter; previously, a range of boolean flags have been passed; now a config-struct is used
- made the cache capacity configurable
- enforces an upper boundary for codes being included in the cache


go/interpreter/lfvm/converter.go
```diff
@@ -21,41 +20,78 @@ import (
...
 	return res
 }
 
+// maxCachedCodeLength is the maximum length of a code in bytes that are
+// retained in the cache. To avoid excessive memory usage, longer codes are not
+// cached. The defined limit is the current limit for codes stored on the chain.
+// Only initialization codes can be longer. Since the Shanghai hard fork, the
+// maximum size of initialization codes is 2 * 24_576 = 49_152 bytes (see
+// https://eips.ethereum.org/EIPS/eip-3860). Such init codes are deliberately
+// not cached to to the expected limited re-use and the missing code hash.
```
facuMH, line 90:
```suggestion
// not cached to the expected limited re-use and the missing code hash.
```

go/interpreter/lfvm/converter.go
```diff
@@ -21,41 +20,78 @@ import (
...
+}
 
-func init() {
-	res, err := lru.New[tosca.Hash, Code](codeCacheCapacity)
-	if err != nil {
-		panic(fmt.Errorf("failed to create cache: %v", err))
+// NewConverter creates a new code converter with the provided configuration.
+func NewConverter(config ConversionConfig) (*Converter, error) {
+	if config.CacheSize == 0 {
+		config.CacheSize = (1 << 30) // = 1GiB
```
facuMH, line 44: nit: if in the future for whatever reason the default size changes, we will have to look for all occurrences of `1<<30`, if it is a constant, it's a simpler change.

facuMH (approved):
nice clean-up !! just a couple of questions/suggestions

go/interpreter/lfvm/converter.go
```diff
@@ -21,41 +20,78 @@ import (
...
+		cache:  cache,
+	}, nil
 }
 
-func Convert(code []byte, withSuperInstructions bool, isInitCode bool, noCodeCache bool, codeHash tosca.Hash) Code {
-	// Do not cache use-once code in create calls.
-	// In those cases the codeHash is also invalid.
-	if isInitCode || noCodeCache {
-		return convert(code, withSuperInstructions)
+// Convert converts EVM code to LFVM code. If the provided code hash is not nil,
```
LuisPH3, line 62: newline in docstring.

go/interpreter/lfvm/lfvm.go
```diff
@@ -10,60 +10,93 @@
...
+					vm, err := NewVm(Config{
+						ConversionConfig: ConversionConfig{
+							WithSuperInstructions: si == "-si",
+						},
+						WithStatistics: stats == "-stats",
+						NoShaCache:     shaCache == "-no-sha-cache",
+						Logging:        logging == "-logging",
+					})
+					if err != nil {
+						panic(err)
```
LuisPH3, line 36:
```suggestion
						panic(fmt.Errorf("faild to initialize cache %v: %v,"lfvm" + si + stats + shaCache + logging, err ))
```

go/interpreter/lfvm/lfvm.go
```diff
@@ -10,60 +10,93 @@
...
+			}
+		}
+	}
+	vm, err := NewVm(Config{
+		ConversionConfig: ConversionConfig{
+			CacheSize: -1,
+		},
+	})
+	if err != nil {
+		panic(err)
```
LuisPH3, line 51:
```suggestion
		panic(fmt.Errorf("faild to initialize cache lfvm-no-code-cache: %v, err ))
```

go/interpreter/lfvm/converter_test.go
```diff
@@ -88,28 +159,45 @@ func TestConverterCheckInvalidCode(t *testing.T) {
...
+	}
+}
+
 func BenchmarkConvertLongExampleCode(b *testing.B) {
+	converter, err := NewConverter(ConversionConfig{})
+	if err != nil {
+		b.Fatalf("failed to create converter: %v", err)
+	}
 	for i := 0; i < b.N; i++ {
-		clearConversionCache()
```
LuisPH3, line 93: Why no clear between runs? 
I'm not sure what are we benchmarking here. I guess both make sense

LuisPH3 (changes requested):
There is a little problem with one of the docstrings, other than that it looks good.

go/interpreter/lfvm/converter.go
```diff
@@ -21,41 +20,78 @@ import (
...
-	cache = res
-}
 
-func clearConversionCache() {
-	cache.Purge()
+	var cache *lru.Cache[tosca.Hash, Code]
+	if config.CacheSize > 0 {
+		var err error
+		capacity := config.CacheSize / maxCachedCodeLength
+		cache, err = lru.New[tosca.Hash, Code](capacity)
```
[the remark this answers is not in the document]
HerbertJordan: yes

go/interpreter/lfvm/converter.go
HerbertJordan: This is only present exactly here and it is the one place where it is needed and need to be changed. I deliberately did not expose a constant to avoid having code depending in whatever way on it.

go/interpreter/lfvm/converter.go
LuisPH3: nevermind, my fault. line wrap

LuisPH3 (dismissed):

go/interpreter/lfvm/converter_test.go
HerbertJordan: Since I pass a `nil` value as the hash for the conversion call, the result is anyway not cached.

LuisPH3 (approved):

PR #709, opened 2024-08-29
Simplify PC map in CT adapter

HerbertJordan:
This PR simplifies the implementation of PcMap's required by the CT adapter to translate between EVM and LFVM code positions. In particuluar:
- it eliminates the duplication of code conversion code in the `convert` function and the `genPcMap` function
- it eliminates a global variable retaining a cache for PcMap instances
- it replaces the maps used by the PcMap with slices
- it moves the PcMap code from the converter file to the CT file, since it is logically a part of the CT adapter
- it removes `error` return values in function signatures where only `nil` was returned

### DIFF
--- a/go/interpreter/lfvm/converter_test.go
+++ b/go/interpreter/lfvm/converter_test.go
@@ -11,9 +11,11 @@
 package lfvm
 
 import (
+	"math/rand"
 	"slices"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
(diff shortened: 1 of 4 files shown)


go/interpreter/lfvm/converter_test.go
```diff
@@ -159,10 +161,84 @@ func TestConvert_AllValidOperationsAreCoveredByConversionTable(t *testing.T) {
...
+		}
+		var pairs []pair
+		res := convertWithObserver(code, ConversionConfig{}, func(evm, lfvm int) {
+			pairs = append(pairs, pair{evm, lfvm})
+		})
+
+		// Check that all operations are mapped to matching operations.
+		for _, p := range pairs {
+			in := vm.OpCode(code[p.evm])
+			want := op_2_op[in]
```
LuisPH3, line 224: I mentioned this in @facuMH PR, op_2_op is not a good name: it has both a cryptic name and signature. 
A name which explains in which direction and between what domains are we translating would be a nice addition.

LuisPH3 (approved):
I really like the member cache instead of global.

go/interpreter/lfvm/converter_test.go
HerbertJordan: Out of scope for this PR, but noted as an element for a future round of improvements.

LuisPH3 (approved):

PR #710, opened 2024-08-29
CT: run uses random seed by default

### LINKED ISSUE #707: Randomize CT seed by default and on Jenkins
In order to increase the range of test cases we are testing on Jenkins, and to avoid re-running the absolute same tests every night, we should consider randomizing the seed of the CT driver at every run -- unless explicitly provided through a command line flag.

facuMH:
Since CT runs nightly or every other night, we do not want to test for exactly the same set of states each run. To fix this issue, this PR changes the seed flag `Fetch` method, so that when `--seed` or `-s` are not set, it uses time as a seed. 
This PR fixes #707

### DIFF
--- a/go/ct/driver/cli/flags.go
+++ b/go/ct/driver/cli/flags.go
@@ -16,6 +16,7 @@ import (
 	"regexp"
 	"runtime"
 	"runtime/pprof"
+	"time"
 
 	"github.com/urfave/cli/v2"
 )
@@ -62,12 +63,16 @@ var SeedFlag = &seedFlagType{
 	cli.Uint64Flag{
 		Name:    "seed",
 		Aliases: []string{"s"},
-		Usage:   "seed for the random number generator",
+		Usage:   "seed for the random number generator. If not provided, the seed is the current time stamp",
 	},
 }
 
 func (f *seedFlagType) Fetch(context *cli.Context) uint64 {
-	return context.Uint64(f.Name)
+	if context.IsSet(f.Name) {
+		return context.Uint64(f.Name)
+	} else {
+		return uint64(time.Now().UnixNano())
+	}
 }
 
 type cpuProfileType struct {
--- a/go/ct/driver/cli/flags.go
+++ b/go/ct/driver/cli/flags.go
@@ -63,16 +63,13 @@ var SeedFlag = &seedFlagType{
 	cli.Uint64Flag{
 		Name:    "seed",
 		Aliases: []string{"s"},
+		Value:   uint64(time.Now().UnixNano()),
 		Usage:   "seed for the random number generator. If not provided, the seed is the current time stamp",
 	},
 }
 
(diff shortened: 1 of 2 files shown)


simonlechner (changes requested):
To ensure that found errors are reproducible, the seed should be printed somewhere.

LuisPH3 (approved):

simonlechner (dismissed):

go/ct/driver/cli/flags.go
```diff
@@ -62,12 +64,16 @@ var SeedFlag = &seedFlagType{
...
 		Name:    "seed",
 		Aliases: []string{"s"},
-		Usage:   "seed for the random number generator",
+		Usage:   "seed for the random number generator. If not provided, the seed is randomly generated",
 	},
 }
 
 func (f *seedFlagType) Fetch(context *cli.Context) uint64 {
-	return context.Uint64(f.Name)
+	if slices.Contains(context.FlagNames(), "seed") {
```
HerbertJordan, line 72: there is a shortcut for this:
```suggestion
	if context.IsSet(f.Name) {
```

go/ct/driver/cli/flags.go
```diff
@@ -62,12 +64,16 @@ var SeedFlag = &seedFlagType{
 	cli.Uint64Flag{
 		Name:    "seed",
 		Aliases: []string{"s"},
-		Usage:   "seed for the random number generator",
+		Usage:   "seed for the random number generator. If not provided, the seed is randomly generated",
```
HerbertJordan, line 67: nit: technically it uses the current time stamp ...

HerbertJordan (dismissed):

HerbertJordan (approved):

simonlechner (commented):
While re-reviewing the code I noticed that if you run the CT in your terminal and it fails after hours, you will most likely not be able to scroll back all the way to the seed. Should the seed also be printed at the end to ensure it is retrievable for everyone?

go/ct/driver/cli/flags.go
```diff
@@ -62,12 +63,16 @@ var SeedFlag = &seedFlagType{
 	cli.Uint64Flag{
 		Name:    "seed",
 		Aliases: []string{"s"},
-		Usage:   "seed for the random number generator",
+		Usage:   "seed for the random number generator. If not provided, the seed is the current time stamp",
```
LuisPH3, line 66: now that I think about it, the same could have been done with a default value:https://cli.urfave.org/v2/examples/flags/

```go
cli.Uint64Flag{
		Name:    "seed",
		Aliases: []string{"s"},
		Value:    uint64(time.Now().UnixNano()),
		Usage:   "seed for the random number generator. If not provided, the seed is the current time stamp",
```

LuisPH3 (approved):

LuisPH3 (approved):

Issue #685, opened 2024-08-26
Introduce free function to compute keccak256 hash


HerbertJordan:
In the LFVM implementation, [hasher](https://github.com/Fantom-foundation/Tosca/blob/8755348be23e93be6b02e87fe5a823d8a8baa865/go/interpreter/lfvm/interpreter.go#L67) instances are retained as part of the execution context to provide a thread-private instance whenever the hashing of some input data is required. This adds complexity to the code and introduces overhead for creating and discarding those instances.

In order to simplify the architecture, a free function capable of computing keccak256 hashes similar to the one offered by the [Carmen](https://github.com/Fantom-foundation/Carmen/blob/main/go/common/keccak.go) project is to be implemented.

PR #712, opened 2024-08-29
Remove hash buffer in LFVM context

HerbertJordan:
This PR removes the hash instance retained in an execution context for hashing input data. After #685 has been added, this is no longer needed.


LuisPH3 (approved):
This is a good cleanup, keccakState was some kind of odd glue.

facuMH (approved):

PR #715, opened 2024-08-30
Fix code style issues in LFVM interpreter

HerbertJordan:
This PR fixes numerous code-style issues in the interpreter file without changing any functionality. The main elements are:
- naming for state constants and variables did not follow our general coding style guidelines
- types and constants had public visibility where it was not needed
- types have been found in identifier names, documentation, and comments
- flag names should be defined in the positive form (e.g. "noShaCache" => "withShaCache")

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -214,7 +214,7 @@ func TestCallChecksBalances(t *testing.T) {
 	source := tosca.Address{1}
 	target := tosca.Address{2}
 	ctxt := context{
-		status: RUNNING,
+		status: statusRunning,
 		params: tosca.Parameters{
 			Recipient: source,
 		},
@@ -235,7 +235,7 @@ func TestCallChecksBalances(t *testing.T) {
 
 	opCall(&ctxt)
 
-	if want, got := RUNNING, ctxt.status; want != got {
+	if want, got := statusRunning, ctxt.status; want != got {
 		t.Errorf("unexpected status after call, wanted %v, got %v", want, got)
 	}
 
@@ -254,7 +254,7 @@ func TestCreateChecksBalance(t *testing.T) {
 
 	source := tosca.Address{1}
 	ctxt := context{
-		status: RUNNING,
+		status: statusRunning,
 		params: tosca.Parameters{
 			Recipient: source,
 		},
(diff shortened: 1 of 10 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -733,7 +733,7 @@ func TestTransientStorageOperations(t *testing.T) {
 		setup    func(*tosca.MockRunContext)
 		stackPtr int
 		revision tosca.Revision
-		status   Status
+		status   status
```
facuMH, line 736: nit: field name same as type name is something I think should be avoided, but I guess in tests it is not such a terrible thing

go/interpreter/lfvm/interpreter.go
```diff
@@ -20,18 +20,18 @@ import (
 	"github.com/holiman/uint256"
 )
 
-type Status byte
+type status byte
```
facuMH, line 23: since field status is really good fit for all it's uses, perhaps this type could be named `statusType` or `statusT` ? wdyt ?

I am aware this is totally legal in go, but I think having variables/fields named the same as their types, is just propense to confusion.

go/interpreter/lfvm/interpreter.go
```diff
@@ -143,26 +145,26 @@ func generateResult(ctxt *context) (tosca.Result, error) {
...
 		}, nil
-	case REVERTED:
+	case statusReverted:
 		return tosca.Result{
 			Success: false,
 			Output:  res,
 			GasLeft: ctxt.gas,
 		}, nil
-	case INVALID_INSTRUCTION, OUT_OF_GAS, MAX_INIT_CODE_SIZE_EXCEEDED, ERROR:
+	case statusInvalidInstruction, statusOutOfGas, statusMaximumInitCodeSizeExceeded, statusError: // < TODO: if all these are handled the same, no need to have anything but statusError
```
facuMH, line 167: considering that throughout the interpreter these errors are treated differently and they are only considered equal in this point. Maybe if there was an if before the switch converting all theses status into `statusError` this could make sense.

facuMH (approved):
thanks for the changes !! it is a broad task, but simple. had to be done. 
just a couple of comment.s

go/interpreter/lfvm/interpreter.go
HerbertJordan: I would argue that it is not good style (and we have not done this before) to encode in an identifier the nature of the concept that is defined. For instance, we do not go for the convention that all interfaces should start with an `I` or that constants have a `k` prefix. I think not having to say that some identifier is a type by not adding a `T` or `Type` into its name falls in the same category.

go/interpreter/lfvm/interpreter.go
HerbertJordan: As I understand, they are not really "treated" differently, they are just "produced" differently. But this code is probably the only consumer, and it does not differentiate between those. So there is no point to have them differentiated.

We need to check whether there are places where those are actually used.

PR #716, opened 2024-08-30
LFVM: stats mode panics with non-terminal code

### LINKED ISSUE #650: LFVM with stats panics with incorrect code
Interpreters should stop if the code is wrong and does not end with a STOP, RETURN or REVERT. LFVM with stats enabled panics due to an out of bounds exception.

facuMH:
during integration testing, it was found that "lfvm-stat" panic when feed incorrect code (aka code that did not end in status `STOPPED`, `REVERTED`, `RETURNED`.

This case is made into a test to verify all variants handle this case properly without producing a panic, and fixed in `lfvm-stat`

This Fixes #650

### DIFF
--- a/go/integration_test/interpreter/integration_test.go
+++ b/go/integration_test/interpreter/integration_test.go
@@ -1110,27 +1110,23 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
 	return mockStateDB
 }
 
-func TestEVMCanSuccessfullyProcessIncorrectCode(t *testing.T) {
+func TestEVM_CanSuccessfullyProcessPcBiggerThanCodeLength(t *testing.T) {
 	for _, variant := range Variants {
-		for _, revision := range revisions {
-			t.Run(fmt.Sprintf("%s/%s", variant, revision), func(t *testing.T) {
-				ctrl := gomock.NewController(t)
-				mockStateDB := newMockStateDBForIntegrationTests(ctrl)
-				// make inner call and return 32byte value with 0 offset from memory
-				codeReturn := []byte{
-					byte(vm.PUSH1), byte(32),
-				}
-				setDefaultCallStateDBMock(mockStateDB, tosca.Address{}, codeReturn)
-				evm := GetCleanEVM(revision, variant, mockStateDB)
+		revision := Istanbul
+		t.Run(variant, func(t *testing.T) {
+			// all implementations should be able to handle PC that goes beyond the code length.
+			// this is reproducible by not calling STOP, RETURN, REVERT or SELFDESTRUCT at the end of the code.
+			code := []byte{byte(vm.PUSH1), byte(32)}
+			evm := GetCleanEVM(revision, variant, nil)
 
-				// Run an interpreter
-				result, err := evm.Run(codeReturn, []byte{})
+			// Run an interpreter
+			result, err := evm.Run(code, []byte{})
 
-				// Check the result.
-				if err != nil || !result.Success {
-					t.Errorf("execution should fail, error is: %v, result %v", err, result)
-				}
-			})
-		}
+			// Check the result.
+			if err != nil || !result.Success {
+				t.Errorf("execution should not fail and err should be nil, error is: %v, success %v", err, result.Success)
+			}
+		})
+
 	}
 }
--- a/go/integration_test/interpreter/integration_test.go
+++ b/go/integration_test/interpreter/integration_test.go
@@ -1109,3 +1109,28 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
 
 	return mockStateDB
 }
+
+func TestEVMCanSuccessfullyProcessIncorrectCode(t *testing.T) {
+	for _, variant := range Variants {
+		for _, revision := range revisions {
+			t.Run(fmt.Sprintf("%s/%s", variant, revision), func(t *testing.T) {
+				ctrl := gomock.NewController(t)
+				mockStateDB := newMockStateDBForIntegrationTests(ctrl)
+				// make inner call and return 32byte value with 0 offset from memory
+				codeReturn := []byte{
+					byte(vm.PUSH1), byte(32),
+				}
+				setDefaultCallStateDBMock(mockStateDB, tosca.Address{}, codeReturn)
+				evm := GetCleanEVM(revision, variant, mockStateDB)
+
+				// Run an interpreter
+				result, err := evm.Run(codeReturn, []byte{})
+
+				// Check the result.
+				if err != nil || !result.Success {
+					t.Errorf("execution should fail, error is: %v, result %v", err, result)
+				}
+			})
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


go/integration_test/interpreter/integration_test.go
```diff
@@ -1109,3 +1109,28 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
...
+func TestEVMCanSuccessfullyProcessIncorrectCode(t *testing.T) {
+	for _, variant := range Variants {
+		for _, revision := range revisions {
+			t.Run(fmt.Sprintf("%s/%s", variant, revision), func(t *testing.T) {
+				ctrl := gomock.NewController(t)
+				mockStateDB := newMockStateDBForIntegrationTests(ctrl)
+				// make inner call and return 32byte value with 0 offset from memory
+				codeReturn := []byte{
+					byte(vm.PUSH1), byte(32),
+				}
```
HerbertJordan, line 1122: This code fragment is completely out of context ... the documentation seems to not match, and I have no idea why this code might be "Incorrect" as stated in the test case name. What problem is this test supposed to trigger?

HerbertJordan (changes requested):
Please improve the test case defined in this PR. It is very cryptic what this test is supposed to do.

go/integration_test/interpreter/integration_test.go
facuMH: I'll correct the comments.

go/integration_test/interpreter/integration_test.go
```diff
@@ -1109,3 +1109,29 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
 
 	return mockStateDB
 }
+
+func TestEVMCanSuccessfullyProcessIncorrectCode(t *testing.T) {
+	for _, variant := range Variants {
+		for _, revision := range revisions {
```
simonlechner, line 1115: Is it really required to run this for all revisions, the stats handling is revision independent.

go/integration_test/interpreter/integration_test.go
```diff
@@ -1109,3 +1109,29 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
...
+			t.Run(fmt.Sprintf("%s/%s", variant, revision), func(t *testing.T) {
+				ctrl := gomock.NewController(t)
+				mockStateDB := newMockStateDBForIntegrationTests(ctrl)
+				// code that does not end with a call to STOP, RETURN, REVERT, or SELFDESTRUCT
+				// is considered incorrect code. All implementations should handle this case gracefully
+				codeReturn := []byte{
+					byte(vm.PUSH1), byte(32),
+				}
```
HerbertJordan, line 1123: This is not incorrect code. It is perfectly valid.

The problem this test is covering is when the program counter of an execution is moving beyond the end of the code. This is allowed, and valid code can do that. Just because the statistics mode of the LFVM failed to handle this properly does not make code invalid.

Please update the comments in the code and the test name accordingly.

go/integration_test/interpreter/integration_test.go
HerbertJordan: This test should not target specifically the bug in the `stats`. It checks a general requirement on all implementations to be able to handle situations in which the PC moves beyond the end of the code.

However, this is a property of the interpreter that is likely revision independent. So the revision loop might indeed be dropped.

HerbertJordan (changes requested):

go/integration_test/interpreter/integration_test.go
```diff
@@ -1109,3 +1109,29 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
...
+			}
+			setDefaultCallStateDBMock(mockStateDB, tosca.Address{}, codeReturn)
+			evm := GetCleanEVM(revision, variant, mockStateDB)
+
+			// Run an interpreter
+			result, err := evm.Run(codeReturn, []byte{})
+
+			// Check the result.
+			if err != nil || !result.Success {
+				t.Errorf("execution should fail, error is: %v, result %v", err, result)
```
simonlechner, line 1132: This error message is going to be very confusing when the tests fails. Although the execution should fail, the error should still be nil. Maybe specify that the execution was not successful and not print the whole result.

simonlechner (dismissed):

go/integration_test/interpreter/integration_test.go
```diff
@@ -1109,3 +1109,29 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
...
+
+func TestEVM_CanSuccessfullyProcessPcBiggerThanCodeLength(t *testing.T) {
+	for _, variant := range Variants {
+		revision := Istanbul
+		t.Run(variant, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			mockStateDB := newMockStateDBForIntegrationTests(ctrl)
+			// all implementations should be able to handle PC that goes beyond the code length.
+			// this is reproducible by not calling STOP, RETURN, REVERT or SELFDESTRUCT at the end of the code.
+			codeReturn := []byte{
```
HerbertJordan, line 1121: The name `codeReturn` is likely still from the code section this code was originally copied from. Please rename to something appropirate.

go/integration_test/interpreter/integration_test.go
```diff
@@ -1109,3 +1109,29 @@ func newMockStateDBForIntegrationTests(ctrl *gomock.Controller) *MockStateDB {
...
+			setDefaultCallStateDBMock(mockStateDB, tosca.Address{}, codeReturn)
+			evm := GetCleanEVM(revision, variant, mockStateDB)
+
+			// Run an interpreter
+			result, err := evm.Run(codeReturn, []byte{})
+
+			// Check the result.
+			if err != nil || !result.Success {
+				t.Errorf("execution should not fail and err should be nil, error is: %v, success %v", err, result.Success)
+			}
```
HerbertJordan, line 1133: This test contains code that is not relevant for the test and that can be removed without changing anything:
```suggestion
			// all implementations should be able to handle PC that goes beyond the code length.
			// this is reproducible by not calling STOP, RETURN, REVERT or SELFDESTRUCT at the end of the code.
			codeReturn := []byte{
				byte(vm.PUSH1), byte(32),
			}
			evm := GetCleanEVM(revision, variant, nil)

			// Run an interpreter
			result, err := evm.Run(codeReturn, []byte{})

			// Check the result.
			if err != nil || !result.Success {
				t.Errorf("execution should fail, error is: %v, result %v", err, result)
			}
```

Please always strive for minimal tests.

HerbertJordan (changes requested):

LuisPH3 (commented):
What is the status of this pr? I guess it is influenced by https://github.com/Fantom-foundation/Tosca/pull/719?

facuMH:
> What is the status of this pr? I guess it is influenced by #719?

they are independent changes, since this one is small enough, it can be done before/after without big impacts

HerbertJordan (dismissed):

simonlechner (approved):

LuisPH3 (approved):

LuisPH3:
> > What is the status of this pr? I guess it is influenced by #719?
> 
> they are independent changes, since this one is small enough, it can be done before/after without big impacts

not so indenpendent after all.

LuisPH3 (approved):

PR #717, opened 2024-08-30
Add Fuzzing function for lfvm opCodes convert function

LuisPH3:
Since the convert function accepts an unconstrained byte buffer coming from outside sources, the convert function is a good candidate for fuzzing, and increase resilience of the algorithm to invalid input.

### DIFF
--- a/go/interpreter/lfvm/converter_fuzz_test.go
+++ b/go/interpreter/lfvm/converter_fuzz_test.go
@@ -1,0 +1,79 @@
+package lfvm
+
+import (
+	"testing"
+
+	"github.com/Fantom-foundation/Tosca/go/ct/gen"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
+
+	"pgregory.net/rand"
+)
+
+func FuzzLfvmConverter(f *testing.F) {
+
+	// Add empty code
+	f.Add([]byte{})
+
+	// Use CT code generator to generate one contract starting with each
+	// opcode
+	rnd := rand.New(1) // deterministic to preserve initial corpus coherence
+	generator := gen.NewCodeGenerator()
+	empty := generator.Clone()
+	for i := 0; i <= 0xFF; i++ {
+		op := vm.OpCode(i)
+		if !vm.IsValid(op) {
+			continue
+		}
+		generator.Restore(empty)
+		generator.SetOperation(0, op)
+		code, err := generator.Generate(gen.Assignment{}, rnd)
+		if err != nil {
+			f.Errorf("Error generating code for opCode %v", op)
+		}
+		f.Add(code.Copy())
+	}
+
+	f.Fuzz(func(t *testing.T, code []byte) {
+
+		// EIP-170 stablish maximum code size
+		// (see https://eips.ethereum.org/EIPS/eip-170)
+		maxCodeSize := 24_576
+		if len(code) > maxCodeSize {
+			t.Skip()
+		}
+
+		type pair struct {
+			originalPos, lfvmPos int
+		}
+		var pairs []pair
+		res := convertWithObserver(code, ConversionConfig{}, func(evm, lfvm int) {
+			pairs = append(pairs, pair{evm, lfvm})
+		})
+
+		// Check that all operations are mapped to matching operations.
+		for _, p := range pairs {
+			var want OpCode
+			if vm.OpCode(code[p.originalPos]) >= vm.PUSH1 && vm.OpCode(code[p.originalPos]) <= vm.PUSH32 {
+				want = PUSH1 + OpCode(res[p.lfvmPos].opcode-PUSH1)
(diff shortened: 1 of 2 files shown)


go/interpreter/lfvm/converter_fuzz_test.go
```diff
@@ -0,0 +1,416 @@
...
+			continue
+		}
+		generator.Restore(empty)
+		generator.SetOperation(0, op)
+		code, err := generator.Generate(gen.Assignment{}, rnd)
+		if err != nil {
+			f.Errorf("Error generating code for opCode %v", op)
+		}
+		f.Add(code.Copy())
+	}
```
HerbertJordan, line 44: It might be way easier (and with less dependencies) to just produce random byte slices and set the first byte to the desired op-code. Wdyt?

go/interpreter/lfvm/converter_fuzz_test.go
```diff
@@ -0,0 +1,416 @@
...
+
+	f.Fuzz(func(t *testing.T, toscaCode []byte) {
+
+		// EIP-170 stablish maximum code size
+		// (see https://eips.ethereum.org/EIPS/eip-170)
+		maxCodeSize := 24_576
```
HerbertJordan, line 50: This is actually not the upper boundary. Initiation codes can be twice as large since Shanghai; before that, init codes have been unrestricted. And we need to converting init codes as well.

HerbertJordan (changes requested):

go/interpreter/lfvm/converter_fuzz_test.go
LuisPH3: This is something I wanted to discuss. Yes, fuzzing runs barely need any initial corpus, the empty slice had a very fast start.
The fuzzing executes millions of inputs per second, I begin to doubt of the utility of feeding any state for this particular test.

go/interpreter/lfvm/converter_fuzz_test.go
LuisPH3: thanks.

go/interpreter/lfvm/converter_fuzz_test.go
HerbertJordan: Then drop it.

go/interpreter/lfvm/converter_fuzz_test.go
```diff
@@ -0,0 +1,398 @@
...
+
+	f.Fuzz(func(t *testing.T, toscaCode []byte) {
+
+		// EIP-170 stablish maximum code size to 24KB
+		// (see https://eips.ethereum.org/EIPS/eip-170)
+		// EIP-3860 stablish maximum init code size to 49_152 bytes
+		// https://eips.ethereum.org/EIPS/eip-3860
+		// Both codes need to be converted, therefore the largest
+		// is used
+		maxCodeSize := 49_152
```
HerbertJordan, line 32: nit: just to be precise ...
```suggestion
		// EIP-170 stablish maximum code size to 0x6000 bytes, ~22 KB
		// (see https://eips.ethereum.org/EIPS/eip-170)
		// EIP-3860 stablish maximum init code size to 49_152 bytes
		// (see https://eips.ethereum.org/EIPS/eip-3860)
		// Before EIP-3860, any size was allowed, but in the Fantom
		// network anything larger than 2^16 was observed, which is
		// also the limit for the conversion code (due to a 16-bit PC
		// counter representation in the PC instruction). Thus, we test
		// the code up to this level.
		maxCodeSize := 1<<16 - 1
```

HerbertJordan (approved):

go/interpreter/lfvm/converter_fuzz_test.go
```diff
@@ -0,0 +1,401 @@
...
+		if !vm.IsValid(op) {
+			continue
+		}
+		if toscaToLfvmOpcode(op) == INVALID {
+			t.Errorf("Opcode %v is not mapped to LFVM", op)
+		}
+	}
+}
+
+func toscaToLfvmOpcode(op vm.OpCode) OpCode {
```
facuMH, line 89: maybe this function and its mirror should be in a separate file and tested. possibly another PR

facuMH (approved):

go/interpreter/lfvm/converter_fuzz_test.go
HerbertJordan: Let's wait with this until we have settled the question on whether and how we are cleaning up the op code definition.

HerbertJordan (approved):

simonlechner (approved):
lgtm

go/interpreter/lfvm/converter_fuzz_test.go
LuisPH3: I make a note on that: #723

PR #720, opened 2024-08-30
Clean up LFVM stack implementation

HerbertJordan:
This PR performs a clean-up pass over LFVM's stack implementation by
- fixing visibility
- aligning the naming of member functions (`Back` => `peekN`, `Data` => `get`)
- adding documentation
- adding unit test coverage

Note to reviewer: this PR does not cover the stack boundary infrastructure that happens to be located in the `stack.go` file but should be moved since it is much closer related to instructions than to the `stack`. This will be covered in a follow-up.

### DIFF
--- a/go/interpreter/lfvm/stack.go
+++ b/go/interpreter/lfvm/stack.go
@@ -59,7 +59,7 @@ func (s *stack) pushUndefined() *uint256.Int {
 
 // pop removes the top element from the stack and returns a pointer to it. The
 // obtained pointer is only valid until the next push operation. The pointer
-// can be used to obtain the pushed element without the need to copy it.
+// can be used to obtain the popped element without the need to copy it.
 func (s *stack) pop() *uint256.Int {
 	s.stackPointer--
 	return &s.data[s.stackPointer]
@@ -84,15 +84,15 @@ func (s *stack) len() int {
 }
 
 // swap exchanges the top element with the n-th element from the top. The top
-// element is at index 1. Thus, swap(1) is a no-op.
+// element is at index 0. Thus, swap(0) is a no-op.
 func (s *stack) swap(n int) {
-	s.data[s.len()-n], s.data[s.len()-1] = s.data[s.len()-1], s.data[s.len()-n]
+	s.data[s.len()-n-1], s.data[s.len()-1] = s.data[s.len()-1], s.data[s.len()-n-1]
 }
 
 // dup duplicates the n-th element from the top and pushes it to the top of the
-// stack. The top element is at index 1. Thus, dup(1) duplicates the top element.
+// stack. The top element is at index 0. Thus, dup(0) duplicates the top element.
 func (s *stack) dup(n int) {
-	s.data[s.stackPointer] = s.data[s.stackPointer-n]
+	s.data[s.stackPointer] = s.data[s.stackPointer-n-1]
 	s.stackPointer++
 }
 
--- a/go/interpreter/lfvm/stack.go
+++ b/go/interpreter/lfvm/stack.go
@@ -11,85 +11,113 @@
 package lfvm
 
 import (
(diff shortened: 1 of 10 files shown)


go/interpreter/lfvm/stack.go
```diff
@@ -11,85 +11,113 @@
...
+// Example usage:
+//
+//	s := NewStack()
+//	defer ReturnStack(s)
+//	<use the stack in your local scope>
+//
+// The stack is not thread-safe. NewStack() and ReturnStack() are thread-safe.
+type stack struct {
+	data         [maxStackSize]uint256.Int
+	stackPointer int
```
facuMH, line 43: since you are renaming, do you think `stackHeight` or `stackSize` could be better names for this field ?

go/interpreter/lfvm/stack.go
```diff
@@ -11,85 +11,113 @@
...
+	s.stackPointer++
+	return &s.data[s.stackPointer-1]
 }
 
-func (s *Stack) push(d *uint256.Int) {
-	s.data[s.stack_ptr] = *d
-	s.stack_ptr++
+// pop removes the top element from the stack and returns a pointer to it. The
+// obtained pointer is only valid until the next push operation. The pointer
+// can be used to obtain the pushed element without the need to copy it.
```
facuMH, line 62:
```suggestion
// can be used to obtain the popped element without the need to copy it.
```

go/interpreter/lfvm/stack.go
```diff
@@ -11,85 +11,113 @@
...
-	return s.stack_ptr
+// len returns the number of elements on the stack.
+func (s *stack) len() int {
+	return s.stackPointer
 }
 
-func (s *Stack) swap(n int) {
+// swap exchanges the top element with the n-th element from the top. The top
+// element is at index 1. Thus, swap(1) is a no-op.
+func (s *stack) swap(n int) {
```
facuMH, line 88: would it not be more intuitive that `swap(0)` is the no-op ? since the documentation says "top element" in `peekN` it says   "The top element is at index 0 Thus, peekN(0) is equivalent to peek().", maybe we should try to have some consistency between methods.

go/interpreter/lfvm/stack.go
```diff
@@ -98,26 +126,40 @@ func (s *Stack) String() string {
...
 	},
 }
 
-func NewStack() *Stack {
-	return stackPool.Get().(*Stack)
+// NewStack returns a new stack instance from the a reuse pool. Heavy stack
+// users should use this function to prevent memory reallocation overhead.
+// This function is thread-safe.
+func NewStack() *stack {
+	return stackPool.Get().(*stack)
```
facuMH, line 137: is the number stacks in the pool a configurable value ? if not, should it be ? This is probably for a follow-up PR

facuMH (commented):
overall looks good, but I have a couple of questions/suggestions before approving

go/interpreter/lfvm/stack.go
HerbertJordan: No, the way the [Pool](https://pkg.go.dev/sync#Pool) works is that every time a request is received, it checks whether there is an element. If there is none, a new one is created and handed out. If one is returned, it is added to the pool. There is no way to set a limit on the pool size. However, the implementation can automatically discard elements if there are too many.

go/interpreter/lfvm/stack.go
HerbertJordan: I think the concept of a `stackPointer` is a well established idea and the name is matching the role it plays here.

go/interpreter/lfvm/stack.go
HerbertJordan: Done. I also updated `dup` to start at index 0 to have all functions aligned.

PR #721, opened 2024-09-02
Integration tests stressing existens of accounts

simonlechner:
This PR adds integration tests stressing two properties of the processor:

1. Accounts can only be called if they exist
2. Accounts can only be created if they do not exist

### DIFF
--- a/go/integration_test/processor/calls_test.go
+++ b/go/integration_test/processor/calls_test.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"math/big"
 	"slices"
+	"strings"
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
@@ -390,6 +391,67 @@ func TestProcessor_RecursiveCallsAfterAStaticCallAreStatic(t *testing.T) {
 				}
 				if !slices.Equal(result.Output, bytes.Repeat([]byte{0}, 32)) {
 					t.Errorf("%s did not enforce static mode and returned %v", callName, result.Output)
+				}
+			})
+		}
+	}
+}
+
+func TestProcessor_CallingNonExistentAccountIsHandledCorrectly(t *testing.T) {
+	tests := map[string]struct {
+		account            Account
+		callValue          tosca.Value
+		accountExistsAfter bool
+	}{
+		"empty": {
+			account:            Account{},
+			callValue:          tosca.Value{},
+			accountExistsAfter: false,
+		},
+		"callWithValue": {
+			account:            Account{},
+			callValue:          tosca.NewValue(42),
+			accountExistsAfter: true,
+		},
+		"existingAccount": {
+			account:            Account{Nonce: 1},
+			callValue:          tosca.Value{},
+			accountExistsAfter: true,
+		},
+	}
+	for processorName, processor := range getProcessors() {
+		// TODO: implement in Floria
+		if strings.Contains(processorName, "floria") {
+			continue
+		}
+		for testName, test := range tests {
+			t.Run(processorName+"/"+testName, func(t *testing.T) {
+				sender := tosca.Address{1}
+				receiver := tosca.Address{2}
+
+				state := WorldState{
+					sender:   Account{Balance: tosca.NewValue(100)},
+					receiver: test.account,
+				}
+				transaction := tosca.Transaction{
+					Sender:    sender,
+					Recipient: &receiver,
+					GasLimit:  sufficientGas,
+					Value:     test.callValue,
+				}
+
+				transactionContext := newScenarioContext(state)
+
+				// Run the processor
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil {
+					t.Errorf("execution failed with error %v", err)
+				}
+				if !result.Success {
+					t.Errorf("execution was not successful")
+				}
+				if test.accountExistsAfter != transactionContext.AccountExists(receiver) {
+					t.Errorf("account has not been created")
 				}
 			})
 		}
--- a/go/integration_test/processor/create_test.go
+++ b/go/integration_test/processor/create_test.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"math/big"
 	"slices"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


go/integration_test/processor/calls_test.go
```diff
@@ -396,6 +397,66 @@ func TestProcessor_RecursiveCallsAfterAStaticCallAreStatic(t *testing.T) {
...
+
+				// Run the processor
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil {
+					t.Errorf("execution failed with error %v", err)
+				}
+				if !result.Success {
+					t.Errorf("execution was not successful")
+				}
+				if test.accountExistsAfter && !transactionContext.AccountExists(receiver) {
```
facuMH, line 452: else case should also test that the account that should not have been created also does not exists ?

go/integration_test/processor/calls_test.go
```diff
@@ -396,6 +397,66 @@ func TestProcessor_RecursiveCallsAfterAStaticCallAreStatic(t *testing.T) {
 	}
 }
 
+func TestProcessor_CallingNonExistentAccountFails(t *testing.T) {
```
facuMH, line 400: this is probably not the best name since the test is actually testing that if the account does not exists, it actually gets created.

go/integration_test/processor/calls_test.go
```diff
@@ -396,6 +397,66 @@ func TestProcessor_RecursiveCallsAfterAStaticCallAreStatic(t *testing.T) {
...
+		}
+		for testName, test := range tests {
+			t.Run(processorName+"/"+testName, func(t *testing.T) {
+				sender := tosca.Address{1}
+				receiver := tosca.Address{2}
+
+				state := WorldState{
+					sender:   Account{Balance: tosca.NewValue(100)},
+					receiver: test.account,
+				}
```
facuMH, line 434: for some reason this was confusing to read. I originally thought `sender` and `reciever` were fields that were getting assigned the other values. 
```suggestion
				state := WorldState{}
				state[sender] = Account{Balance: tosca.NewValue(100)}
				state[receiver] = test.account
```
maybe this is a more clear way of showing what the test is doing.

go/integration_test/processor/create_test.go
```diff
@@ -451,6 +452,87 @@ func TestProcessor_CorrectAddressIsCreated(t *testing.T) {
...
+
+		// Different to ethereum, on Sonic accounts can only have set storage if their nonce != 0
+		"withStorage": {
+			account: Account{Nonce: 1, Storage: storageWithEntry},
+			success: false,
+		},
+	}
+
+	for processorName, processor := range getProcessors() {
+		if strings.Contains(processorName, "floria") {
```
facuMH, line 490: why is floria not being tested ? maybe add a comment with a  `TODO`

go/integration_test/processor/create_test.go
```diff
@@ -451,6 +452,87 @@ func TestProcessor_CorrectAddressIsCreated(t *testing.T) {
...
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil {
+					t.Errorf("execution failed with error %v", err)
+				}
+				if result.Success != test.success {
+					t.Errorf("execution success was %v, expected %v", result.Success, test.success)
+				}
+				if result.Success && !slices.Equal(result.Output, append(bytes.Repeat([]byte{0}, 31), checkValue)) {
+					t.Errorf("creation returned successful but with wrong output %v", result.Output)
+				}
```
facuMH, line 530:
```suggestion
				if result.Success != test.success {
					t.Errorf("execution success was %v, expected %v", result.Success, test.success)
				} else  if  !slices.Equal(result.Output, append(bytes.Repeat([]byte{0}, 31), checkValue)) {
					t.Errorf("creation returned successful but with wrong output %v", result.Output)
				}
```
wdyt ?

facuMH (commented):
looks good, just a couple of suggestions to improve clarity

go/integration_test/processor/calls_test.go
HerbertJordan: A stricter version of this would be
```suggestion
				if test.accountExistsAfter != transactionContext.AccountExists(receiver) {
```

go/integration_test/processor/calls_test.go
HerbertJordan: I agree, the name is confusing since calls to empty accounts should actually always succeed.

If I recall correctly, for such calls the empty code is assumed, which is equivalent to a code that just has a `STOP` instruction. Stops do not lead to reverts, thus the calls should succeed.

go/integration_test/processor/create_test.go
```diff
@@ -451,6 +452,87 @@ func TestProcessor_CorrectAddressIsCreated(t *testing.T) {
...
+			})
```
HerbertJordan, line 531: Unless this is already tested somewhere else, the following post-condition checks should be added:
- in case the account was created, the new account has a nonce of 1, no balance, empty storage, and the given code
- in case the account was not created, it should be checked that the resulting world state is the same as the initial state (except for the nonce of the account paying for the creation being increased by 1)

HerbertJordan (changes requested):

go/integration_test/processor/calls_test.go
simonlechner: I don't think this is readable, since it just repeats the name of the constructed object. I think it is very common in golang to directly fill maps with values.

facuMH (approved):

go/integration_test/processor/calls_test.go
```diff
@@ -396,6 +397,67 @@ func TestProcessor_RecursiveCallsAfterAStaticCallAreStatic(t *testing.T) {
...
+				if err != nil {
+					t.Errorf("execution failed with error %v", err)
+				}
+				if !result.Success {
+					t.Errorf("execution was not successful")
+				}
+				if test.accountExistsAfter != transactionContext.AccountExists(receiver) {
+					t.Errorf("account has not been created")
+				}
+			})
```
HerbertJordan, line 456: Should the sender and receiver account balances after the transaction be tested as well or is this beyond the scope of this test?

HerbertJordan (approved):

PR #722, opened 2024-09-02
lfvm: introduce CT sub package

facuMH:
CT adapter is not an essential part of the interpreter implementation. Because of this, we move it to its own folder and package.
To enable this separation a few members of `lfvm` had to be made public. The amount of public conversions was kept to a minimum. IF this makes the package looks inconsistent, we can discuss how to best approach it.
Members made public:
- `convertWithObserver` method from `converter.go`
- `newestSupportedRevision` constant
- `status` enum (type and values)

Introduced methods to prevent making public types or members public:
- `NewInstruction`, `IsIndexOp` and `GetArgOf`, to create, consult `index[i] == op` and getter, correspondingly 
- `NewContext`, a series of getters, `IsRunning` and `SignalOutofGas`.

Stack:
- `ReturnStack` is modified to take a `Context` as parameter and return its stack to the pool, this way the context's stack pointer does not need to be exposed.
- `stackLimit` constant is made public 
- `len`, `push` and `pushEmpty` are also made public

### DIFF
--- a/go/interpreter/lfvm/ct/ct.go
+++ b/go/interpreter/lfvm/ct/ct.go
@@ -8,7 +8,7 @@
 // On the date above, in accordance with the Business Source License, use of
 // this software will be governed by the GNU Lesser General Public License v3.
 
-package lfvm
+package ct
 
 import (
 	"fmt"
@@ -17,12 +17,17 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/ct/common"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
 	"github.com/Fantom-foundation/Tosca/go/ct/utils"
+	"github.com/Fantom-foundation/Tosca/go/interpreter/lfvm"
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/holiman/uint256"
 )
 
 func NewConformanceTestingTarget() ct.Evm {
-	converter, err := NewConverter(ConversionConfig{
+	// can only produce an error if ConversionConfig.CacheSize is set to
+	// less than maxCachedCodeLength = 24_576 bytes, but default is 1GiB,
+	// so this can't fail.
+	converter, err := lfvm.NewConverter(lfvm.ConversionConfig{
 		WithSuperInstructions: false,
 	})
 	if err != nil {
@@ -36,13 +41,13 @@ func NewConformanceTestingTarget() ct.Evm {
 }
 
 type ctAdapter struct {
-	converter  *Converter
+	converter  *lfvm.Converter
 	pcMapCache *lru.Cache[[32]byte, *pcMap]
 }
 
 func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 	params := utils.ToVmParameters(state)
-	if params.Revision > newestSupportedRevision {
+	if params.Revision > lfvm.NewestSupportedRevision {
 		return state, &tosca.ErrUnsupportedRevision{Revision: params.Revision}
 	}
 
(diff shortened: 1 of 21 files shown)


go/interpreter/lfvm/ct/ct.go
```diff
@@ -64,47 +69,48 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
...
-		context:    params.Context,
-		gas:        params.Gas,
-		refund:     tosca.Gas(state.GasRefund),
-		stack:      convertCtStackToLfvmStack(state.Stack),
-		memory:     memory,
-		status:     statusRunning,
-		code:       converted,
-		returnData: state.LastCallReturnData.ToBytes(),
-	}
+	var ctxt = lfvm.NewContext(
```
LuisPH3, line 72: IMO this is not better, 
We had named fields initialization and the new code has way too many arguments for a function call, including 3 new undocumented arguments at the end that could be defaulted or explained. 
If the goal is to prevent leaking private types, I would reconsider first if making the context public is that bad. 
If it cannot happen. Then I would remove any defaulted value from this signature like the three obscure last arguments, status, last call return data, etc... 

If this function is really unavoidable and still has more than 6 arguments, I would consider adding a single argument being a struct, so that we can have the named arguments back.

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -117,7 +117,7 @@ type example struct {
 	function uint32 // The identifier of the function in the contract to be called
 }
 
-const MAX_STACK_SIZE int = 1024
+const MAX_STACK_SIZE int = 1024 // TODO: use StackLimit
```
LuisPH3, line 120: stack limit is now public, did you miss it, or can't it be used yet?

go/interpreter/lfvm/ct/ct.go
```diff
@@ -17,12 +17,17 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/ct/common"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
 	"github.com/Fantom-foundation/Tosca/go/ct/utils"
+	"github.com/Fantom-foundation/Tosca/go/interpreter/lfvm"
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/holiman/uint256"
 )
```
LuisPH3, line 24: Just looking at the import list of this module, it makes me think that this code relates more to CT than to lfvm. I think it may be better there.

LuisPH3 (commented):
This change removes dependency of lfvm to ct:

- It seems to make a lot of things public that seem implementation, convertWithObserver was a surprise, but maybe unavoidable
- New imports in existing ct code have alias that rename the module. It looks to me like a code smell. 
- NewContext function has way too many arguments to be a clean public interface.

go/interpreter/lfvm/interpreter_test.go
facuMH: for some reason I could not use this constant, this is why I made `StackLimit` public.

go/interpreter/lfvm/ct/ct.go
facuMH: actually, if you look into the other folder of `go/interpreters/` this is the correct place for the `ct` folder, it is perhaps not the right place for the lfvm implementaiton. But the idea of the `interpreter` folder was to remove the `ct` adaptors from the ct implementation packages. Perhaps `lfvm` package should be moved to another folder. @HerbertJordan

go/interpreter/lfvm/interpreter_test.go
LuisPH3: I don't understand your reply.
- does it mean that it does not need to be public anymore? 
- it can be used, I tested it. Even better, we can remove MAX_STACK_SIZE and use StackLimit

facuMH: I tried using this constant from ct package, but compiler complained it did not exist. I would prefer to have just one constant rather than 2.

facuMH:
To move CT adapter to another package would need too many lfvm internals to be made public. Because of this, we are not pushing forward with this idea. at least for now.

PR #726, opened 2024-09-04
Floria: Implement account exists check

### LINKED ISSUE #713: Floria: add account exits check

simonlechner:
This PR adds the checks whether an account exists or not to Floria.
By adding these checks in between precompiled contract handling and the interpreter call, several checks can be removed compared to the reference implementation.
Fixes #713.

### DIFF
--- a/go/processor/floria/run_context.go
+++ b/go/processor/floria/run_context.go
@@ -20,6 +20,8 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
+var emptyCodeHash = tosca.Hash(crypto.Keccak256(nil))
+
 type runContext struct {
 	tosca.TransactionContext
 	interpreter           tosca.Interpreter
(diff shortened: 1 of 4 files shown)


go/processor/floria/run_context.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
+var emptyCodeHash = tosca.Hash(crypto.Keccak256(nil))
```
facuMH, line 23: global variable to be used just once ?

HerbertJordan (approved):

go/processor/floria/run_context.go
simonlechner: Yes, because I don't want to recompute the hash every time the function is called.

facuMH (approved):

LuisPH3 (approved):

Issue #714, opened 2024-08-30
Floria: support new revisions


simonlechner:
Multiple EIPs impact the processor since Istanbul, these changes have to be added to Tosca's processor implementation.

Most notably:
- [x] Access list
- [x] Reject code starting with 0xEF
- [x] Limit code size

PR #731, opened 2024-09-05
Floria: Add access list

simonlechner:
This PR adds the access list introduced in the Berlin revision to Floria.
The integration test scenario had to been changed to also deal with keys. The current implementation requires an iteration through all (worst case) entries which is not performant but should not be a problem for the scenario, which is only used in tests.  
Contributes to #714.

### DIFF
--- a/go/integration_test/processor/scenario.go
+++ b/go/integration_test/processor/scenario.go
@@ -115,7 +115,7 @@ type scenarioContext struct {
 	current    WorldState
 	logs       []tosca.Log
 	undo       []func()
-	accessList map[tosca.Address]tosca.AccessStatus
+	accessList []tosca.AccessTuple
 }
 
 func NewScenarioContext() *scenarioContext {
@@ -223,16 +223,29 @@ func (c *scenarioContext) SetTransientStorage(tosca.Address, tosca.Key, tosca.Wo
 }
 
 func (c *scenarioContext) AccessAccount(address tosca.Address) tosca.AccessStatus {
-	_, ok := c.accessList[address]
-	if !ok {
-		return tosca.WarmAccess
+	for _, tuple := range c.accessList {
+		if tuple.Address == address {
+			return tosca.WarmAccess
+		}
 	}
-	c.accessList[address] = true
+	c.accessList = append(c.accessList, tosca.AccessTuple{Address: address})
 	return tosca.ColdAccess
 }
 
-func (c *scenarioContext) AccessStorage(tosca.Address, tosca.Key) tosca.AccessStatus {
-	panic("implement me")
+func (c *scenarioContext) AccessStorage(addr tosca.Address, key tosca.Key) tosca.AccessStatus {
+	for _, tuple := range c.accessList {
+		if tuple.Address == addr {
+			for _, k := range tuple.Keys {
+				if k == key {
+					return tosca.WarmAccess
+				}
+			}
+			tuple.Keys = append(tuple.Keys, key)
+			return tosca.ColdAccess
+		}
+	}
+	c.accessList = append(c.accessList, tosca.AccessTuple{Address: addr, Keys: []tosca.Key{key}})
+	return tosca.ColdAccess
 }
 
 func (c *scenarioContext) EmitLog(log tosca.Log) {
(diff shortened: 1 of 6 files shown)


facuMH (approved):
looks good. I got a little intrigued about the difference in performance between `map` search and `slice` search (iteration). if you could provide a little info on this, that would be great.

go/integration_test/processor/scenario.go
```diff
@@ -223,16 +223,29 @@ func (c *scenarioContext) SetTransientStorage(tosca.Address, tosca.Key, tosca.Wo
 }
 
 func (c *scenarioContext) AccessAccount(address tosca.Address) tosca.AccessStatus {
-	_, ok := c.accessList[address]
-	if !ok {
-		return tosca.WarmAccess
+	for _, tuple := range c.accessList {
```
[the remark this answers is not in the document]
simonlechner: I don't think a benchmark will be necessary. Here is a quick reminder on search complexities of maps and slices: https://stackoverflow.com/a/18863041

LuisPH3 (approved):

facuMH (dismissed):

go/processor/floria/precompiled.go
```diff
@@ -36,6 +36,12 @@ func handlePrecompiledContract(revision tosca.Revision, input tosca.Data, addres
 }
 
 func getPrecompiledContract(address tosca.Address, revision tosca.Revision) (geth.PrecompiledContract, bool) {
+	precompiles := getPrecompiledContracts(address, revision)
+	contract, ok := precompiles[common.Address(address)]
+	return contract, ok
+}
+
+func getPrecompiledContracts(address tosca.Address, revision tosca.Revision) map[common.Address]geth.PrecompiledContract {
```
HerbertJordan, line 44: This function does not need the `address` parameter.

go/processor/floria/processor.go
```diff
@@ -108,6 +112,27 @@ func (p *processor) Run(
 	}, nil
 }
 
+func setUpAccessList(transaction tosca.Transaction, context tosca.TransactionContext, revision tosca.Revision) {
+	if transaction.AccessList == nil {
+		return
+	}
+
+	if transaction.Recipient != nil {
+		context.AccessStorage(*transaction.Recipient, tosca.Key{})
```
HerbertJordan, line 121: Should this use [`AccessAccount`](https://github.com/Fantom-foundation/Tosca/blob/cd54781393a81426f6174a67d8a6b08b95677a9d/go/tosca/interpreter.go#L95) instead?

go/processor/floria/processor.go
```diff
@@ -108,6 +112,27 @@ func (p *processor) Run(
...
+	}
+
+	precompiles := getPrecompiledAddresses(revision)
+	for _, address := range precompiles {
+		context.AccessStorage(address, tosca.Key{})
```
HerbertJordan, line 126: Also here, use `AccessAccount`?

go/processor/floria/processor_test.go
```diff
@@ -367,3 +367,28 @@ func TestProcessor_CallParameters(t *testing.T) {
...
+
+	context.EXPECT().AccessStorage(*transaction.Recipient, tosca.Key{})
+	for _, contract := range getPrecompiledAddresses(tosca.R09_Berlin) {
+		context.EXPECT().AccessStorage(contract, tosca.Key{})
+	}
+	context.EXPECT().AccessStorage(tosca.Address{2}, tosca.Key{1})
+	context.EXPECT().AccessStorage(tosca.Address{2}, tosca.Key{2})
+
+	setUpAccessList(transaction, context, tosca.R09_Berlin)
+	ctrl.Finish()
```
HerbertJordan, line 393: This should not be needed. If you initialize the controller with `t` it will add automatically an end-of-test callback to finalize the controller.

HerbertJordan (changes requested):
Looks good, some minor requests for changes.

HerbertJordan (approved):

LuisPH3 (approved):

facuMH (approved):

PR #734, opened 2024-09-05
 Refactor vm.OpCode IsValid 

LuisPH3:
Replace string comparing function with something faster.
Unit test for the function already existed. 

image
Note: benchmark has been refactored, values here are still valid but factor 256 in each case.

### DIFF
--- a/go/tosca/vm/opcodes_test.go
+++ b/go/tosca/vm/opcodes_test.go
@@ -93,8 +93,6 @@ func TestOpCode_NumberOfOpCodes(t *testing.T) {
 
 func BenchmarkOpcodesIsValid(b *testing.B) {
 	for i := 0; i < b.N; i++ {
-		for j := 0; j < 256; j++ {
-			IsValid(OpCode(j))
-		}
+		IsValid(OpCode(i % 256))
 	}
 }
--- a/go/tosca/vm/opcodes_test.go
+++ b/go/tosca/vm/opcodes_test.go
@@ -90,3 +90,11 @@ func TestOpCode_NumberOfOpCodes(t *testing.T) {
 	}
 
 }
+
+func BenchmarkOpcodesIsValid(b *testing.B) {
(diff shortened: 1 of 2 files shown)


simonlechner (dismissed):
Nice change!

facuMH (approved):
it might not be the most elegant solution, but it is a nice that can be expanded in other implementations, thanks !

facuMH (dismissed):

go/tosca/vm/opcodes_test.go
```diff
@@ -90,3 +90,11 @@ func TestOpCode_NumberOfOpCodes(t *testing.T) {
 	}
 
 }
+
+func BenchmarkOpcodesIsValid(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		for j := 0; j < 256; j++ {
+			IsValid(OpCode(j))
+		}
+	}
```
HerbertJordan, line 99: nit: This actually benchmarks the time to make 256 IsValid requests, since what is reported is the time of each loop iteration. This could be misleading since it is off by 2 orders of magnitude.

I would suggest to do something like this:
```suggestion
	for i := 0; i < b.N; i++ {
		IsValid(OpCode(i%256))
	}
```

HerbertJordan (approved):

go/tosca/vm/opcodes_test.go
LuisPH3: Because the function is so fast, I wanted to increase the workload for a compound measurement. I am not very familiar with the benchmark tooling: I can see that count just repeats the same benchmark, and I am worried it will get very fluctuating measurements. 
On the other hand, measuring the time it takes to convert every op-code seems like the only fair comparison algorithm to me, since it could be implemented using a switch, a map or other structure with non O(1) access cost.

facuMH (dismissed):

HerbertJordan (approved):

facuMH (approved):

PR #735, opened 2024-09-05
Floria: Invalid code starting with 0xEF is rejected

simonlechner:
Starting with the London revision, it is no longer possible to create accounts with a code starting with 0xEF. 
This simple check is added to Floria and checked with an integration test.

### DIFF
--- a/go/integration_test/processor/create_test.go
+++ b/go/integration_test/processor/create_test.go
@@ -549,6 +549,97 @@ func TestProcessor_CreateExistingAccountFails(t *testing.T) {
 	}
 }
 
+func TestProcessor_CodeStartingWith0xEFCanNotBeCreated(t *testing.T) {
+	tests := map[string]struct {
+		revision         tosca.Revision
+		firstInstruction byte
+		success          bool
+	}{
+		"preLondon0xEF": {
+			revision:         tosca.R09_Berlin,
+			firstInstruction: byte(0xEF),
+			success:          true,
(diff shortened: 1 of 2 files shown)


go/integration_test/processor/create_test.go
```diff
@@ -549,6 +549,97 @@ func TestProcessor_CreateExistingAccountFails(t *testing.T) {
...
+	}
+
+	for processorName, processor := range getProcessors() {
+		for testName, test := range tests {
+			t.Run(processorName+"/"+testName, func(t *testing.T) {
+				sender := tosca.Address{1}
+				addressToBeCreated := tosca.Address(crypto.CreateAddress(common.Address(sender), 0))
+
+				initCode := []byte{
+					byte(vm.PUSH1), test.firstInstruction,
```
LuisPH3, line 597: Second instruction?

LuisPH3 (changes requested):
I think I don't get it. The input code changes the second instruction.

facuMH (approved):

go/integration_test/processor/create_test.go
simonlechner: The code that is used to create a new account is the output of the execution of the init code. The init code in the test only returns the `firstInstruction`.

LuisPH3 (approved):

HerbertJordan (approved):

PR #740, opened 2024-09-06
Cleanup of LFVM interpreter loop

HerbertJordan:
This PR eliminates redundant code from the LFVM interpreter loop. This cuts 4-11% of the execution time for dispatcher-intensive benchmarks like Fib.

```
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/integration_test/interpreter
                  │ baseline.txt │             modified.txt             │
                  │    sec/op    │    sec/op     vs base                │
Fib/10/lfvm-12       179.4µ ± 5%   160.7µ ±  5%  -10.45% (p=0.002 n=10)
Fib/10/lfvm-si-12    127.5µ ± 7%   116.0µ ± 10%   -9.04% (p=0.004 n=10)
geomean              151.3µ        136.5µ         -9.75%
```

```
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/integration_test/interpreter
cpu: Intel(R) Xeon(R) CPU @ 2.80GHz
                  │ baseline.txt │            modified.txt             │
                  │    sec/op    │   sec/op     vs base                │
Fib/10/lfvm-16       118.8µ ± 0%   114.0µ ± 2%   -4.08% (p=0.000 n=10)
Fib/10/lfvm-si-16    92.74µ ± 1%   81.93µ ± 0%  -11.66% (p=0.000 n=10)
geomean              105.0µ        96.63µ        -7.94%
```

Pending:
- [x] run CT check

### DIFF
--- a/go/interpreter/lfvm/interpreter.go
+++ b/go/interpreter/lfvm/interpreter.go
@@ -243,30 +243,12 @@ func steps(c *context, oneStepOnly bool) {
 
 		op := c.code[c.pc].opcode
 
-		// JUMP_TO is an LFVM specific operation that has no gas costs nor stack usage.
-		if c.code[c.pc].opcode == JUMP_TO {
-			c.pc = int32(c.code[c.pc].arg)
-			op = c.code[c.pc].opcode
-		}
-
-		// Catch invalid op-codes here, to avoid the need to check them at other places multiple times.
-		if op >= NUM_EXECUTABLE_OPCODES {
-			c.signalError()
-			return
-		}
-
-		// Need to check Call stack boundary before using static gas
-		// TODO: check whether this can be removed
-		if op == CALL && !satisfiesStackRequirements(c, op) {
-			return
-		}
-
 		// If the interpreter is operating in readonly mode, make sure no
 		// state-modifying operation is performed. The 3rd stack item
 		// for a call operation is the value. Transferring value from one
 		// account to the others means the state is modified and should also
 		// return with an error.
-		if c.params.Static && (isWriteInstruction(op) || (op == CALL && c.stack.peekN(2).Sign() != 0)) {
+		if c.params.Static && (isWriteInstruction(op) || (op == CALL && c.stack.len() > 3 && c.stack.peekN(2).Sign() != 0)) {
 			c.signalError()
 			return
 		}
--- a/go/interpreter/lfvm/interpreter_test.go
+++ b/go/interpreter/lfvm/interpreter_test.go
@@ -652,16 +652,12 @@ func TestStepsProperlyHandlesJUMP_TO(t *testing.T) {
 
 func TestStepsDetectsNonExecutableCode(t *testing.T) {
 	// Create execution context.
-	instructions := []struct {
-		instruction []Instruction
-		status      status
-	}{
-		{[]Instruction{{NUM_EXECUTABLE_OPCODES - 1, 0x0101}, {DATA, 0x0001}, {STOP, 0}}, statusStopped},
-		{[]Instruction{{NUM_EXECUTABLE_OPCODES, 0}}, statusError},
-		{[]Instruction{{NUM_EXECUTABLE_OPCODES + 1, 0}}, statusError},
+	opCodes := []OpCode{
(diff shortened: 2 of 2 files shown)


go/interpreter/lfvm/interpreter.go
```diff
@@ -243,30 +243,12 @@ func steps(c *context, oneStepOnly bool) {
 
 		op := c.code[c.pc].opcode
 
-		// JUMP_TO is an LFVM specific operation that has no gas costs nor stack usage.
-		if c.code[c.pc].opcode == JUMP_TO {
```
LuisPH3, line 247: This check comes as a surprise to me, but I guess it makes sense, it would tax every other instruction and behaviour is trivial.

go/interpreter/lfvm/interpreter.go
```diff
@@ -243,30 +243,12 @@ func steps(c *context, oneStepOnly bool) {
...
-			c.pc = int32(c.code[c.pc].arg)
-			op = c.code[c.pc].opcode
-		}
-
-		// Catch invalid op-codes here, to avoid the need to check them at other places multiple times.
-		if op >= NUM_EXECUTABLE_OPCODES {
-			c.signalError()
```
LuisPH3, line 254: this is something we could improve next:



The new invalid instruction (that replaces the deleted code), does something slightly different now, this has been noticed and the test has been rewritten. 
```
default:
			c.status = statusInvalidInstruction
			return
		}
```


There is a different style to report errors, signalError | c.status assign.
- I am not a big fan of signalError, as it seems to do something else but an assignment, maybe setError would be better.
- on the other hand, we may not need this fine granularity of errors, as mentioned here: https://github.com/Fantom-foundation/Tosca/blob/c9b04766a92b31bcab146f1b0ba64e459086384e/go/interpreter/lfvm/interpreter.go#L167


It is true that some other testing tools may be expecting some of these codes, but the invalid instruction is not one of them because it was triggering the deleted if evey time, therefore producing generic error status.

facuMH (approved):

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -652,16 +652,12 @@ func TestStepsProperlyHandlesJUMP_TO(t *testing.T) {
...
-	instructions := []struct {
-		instruction []Instruction
-		status      status
-	}{
-		{[]Instruction{{NUM_EXECUTABLE_OPCODES - 1, 0x0101}, {DATA, 0x0001}, {STOP, 0}}, statusStopped},
-		{[]Instruction{{NUM_EXECUTABLE_OPCODES, 0}}, statusError},
-		{[]Instruction{{NUM_EXECUTABLE_OPCODES + 1, 0}}, statusError},
+	opCodes := []OpCode{
+		NUM_EXECUTABLE_OPCODES,
+		NUM_EXECUTABLE_OPCODES + 1,
```
LuisPH3, line 657: I think it is important to have data input that does not trigger the property.

By not testing an executable opcode, we cannot test whenever the function is just returning always true. 
nit: I would add INVALID, NOP, and DATA because they are located inside of the "range"

LuisPH3 (changes requested):
It looks pretty good, nevertheless I am not satisfied with the test.

go/interpreter/lfvm/interpreter_test.go
HerbertJordan: You are right, there should always be tests for the positive and negative cases. And this should be improved here. However, I think all the unit tests in this file are a mess and they should get a complete do-over -- and I plan to do so before the release. Thus, I do not really care about these changes now.

LuisPH3 (approved):

go/interpreter/lfvm/interpreter.go
HerbertJordan: The function `signalError()` should probably be renamed to `abort()`. About the different error cases -- I think the fine granularity is nowhere utilized and we can generalize wherever we need to.

LuisPH3: #744

PR #741, opened 2024-09-06
Add super-instruction decomposition support

HerbertJordan:
This PR adds two function to identify super instructions and and to decompose them into their basic instructions.


facuMH (approved):
looks good.

PR #743, opened 2024-09-06
Refactor allowed operations in static context check. 

LuisPH3:
Some Instructions are not allowed in static contexts, such checks can be done per opcode basis instead of in the main dispatching function. 

Checks have been moved into each op execution routine, Call check (regarding zero value) was already in place. 

- [x] CT

image

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -244,6 +244,13 @@ func opMsize(c *context) {
 }
 
 func opSstore(c *context) {
+
+	// SStore is a write instruction, it shall not be executed in static mode.
+	if c.params.Static {
+		c.signalError()
+		return
+	}
+
 	gasfunc := gasSStore
 	if c.isAtLeast(tosca.R09_Berlin) {
 		gasfunc = gasSStoreEIP2929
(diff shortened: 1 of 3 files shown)


go/interpreter/lfvm/interpreter.go
```diff
@@ -659,18 +649,3 @@ func _precomputeStackLimits() [256]stackLimits {
 	}
 	return res
 }
-
-func isWriteInstruction(opCode OpCode) bool {
```
facuMH, line 663: is there already a test that checks that these opcodes produce `ERROR` status ?

go/interpreter/lfvm/instructions.go
HerbertJordan, line 1: Suspiciously the CALL operation is missing ...

HerbertJordan (changes requested):

go/interpreter/lfvm/interpreter.go
LuisPH3: yap: 
https://github.com/Fantom-foundation/Tosca/blob/adfa2779d6536498ed286d3f25c0f87724e52a43/go/interpreter/lfvm/interpreter_test.go#L680

simonlechner (approved):

HerbertJordan (approved):
Can you add benchmark results for the impact of this change to the PR description? A table showing the before/after speed of Fib10 is sufficient. Having this info as part of the PR rounds it off.

facuMH (approved):

Issue #616, opened 2024-07-16
Upgrade deprecated function calls to use replacements.


LuisPH3:
[Staticcheck](https://staticcheck.dev/docs/) reports use of deprecated functions.

install `go install honnef.co/go/tools/cmd/staticcheck@latest`
call `staticcheck ./go/...`

```
go/ct/utils/adapter_test.go:166:26: ctxt.GetCommittedStorage is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/ct/utils/adapter_test.go:179:18: ctxt.GetCommittedStorage is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/ct/utils/adapter_test.go:188:15: ctxt.IsSlotInAccessList is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/ct/utils/adapter_test.go:200:15: ctxt.IsSlotInAccessList is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/ct/utils/adapter_test.go:314:19: ctxt.IsAddressInAccessList is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/ct/utils/adapter_test.go:324:18: ctxt.IsAddressInAccessList is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/interpreter/geth/geth.go:303:21: s.context.GetCommittedStorage is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/interpreter/geth/geth.go:332:9: s.context.HasSelfDestructed is deprecated: should not be needed  (SA1019)
go/interpreter/geth/geth.go:364:9: s.context.IsAddressInAccessList is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/interpreter/geth/geth.go:368:9: s.context.IsSlotInAccessList is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/interpreter/lfvm/gas.go:315:14: c.context.GetCommittedStorage is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/interpreter/lfvm/gas.go:363:33: c.context.IsSlotInAccessList is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
go/interpreter/lfvm/gas.go:377:14: c.context.GetCommittedStorage is deprecated: should not be needed when using result of SetStorage(..)  (SA1019)
[...]

PR #745, opened 2024-09-06
Cleanup of LFVM's SLOAD/SSTORE operations

HerbertJordan:
This PR applies some clean-up steps to the implementation of the SLOAD and SSTORE operations:
- replaces usage of deprecated functions with the updates alternatives
- inlines the gas and refund cost computation
- adds exhaustive test coverage for gas cost and refund computation using a specification derived from [evm.codes](https://evm.codes)


LuisPH3:
relates to #616

go/interpreter/lfvm/gas.go
```diff
@@ -316,131 +310,71 @@ func callGas(availableGas, base tosca.Gas, callCost *uint256.Int) tosca.Gas {
...
-//     2.2.1.2. If new value is 0 (also means that current value is not 0), add SSTORE_CLEARS_SCHEDULE gas to refund counter.
-//     2.2.2. If original value equals new value (this storage slot is reset):
-//     2.2.2.1. If original value is 0, add SSTORE_SET_GAS - SLOAD_GAS to refund counter.
-//     2.2.2.2. Otherwise, add SSTORE_RESET_GAS - SLOAD_GAS gas to refund counter.
-func gasSStoreEIP2200(c *context) (tosca.Gas, error) {
-	// If we fail the minimum gas availability invariant, fail (0)
-	if c.gas <= SstoreSentryGasEIP2200 {
-		c.status = statusOutOfGas
-		return 0, errNotEnoughGasReentrancy
+	cost := tosca.Gas(800)
```
facuMH, line 319: this might be a personal opinion but having `cost` be assigned named variables with significant values (such as `SstoreSetGasEIP2200` or whatever the case would be), seems a lot more significant than just a constant number without any background or reason as for why is that number chosen and not another one. I know this numbers have been calculated with evm codes or something, but having them in a set of predefined constants was very helpful to understand the different cases when reading the code.

go/interpreter/lfvm/gas_test.go
```diff
@@ -59,3 +59,171 @@ func TestGas_CallGasCalculation(t *testing.T) {
...
+		tosca.R07_Istanbul: makeSpec(20000, 5000, 800),
+		// source: https://www.evm.codes/?fork=berlin
+		tosca.R09_Berlin: makeSpec(20000, 2900, 100),
+	}
+
+	// All other revisions are based on Berlin.
+	specs[tosca.R10_London] = specs[tosca.R09_Berlin]
+	specs[tosca.R11_Paris] = specs[tosca.R10_London]
+	specs[tosca.R12_Shanghai] = specs[tosca.R11_Paris]
+	specs[tosca.R13_Cancun] = specs[tosca.R12_Shanghai]
```
facuMH, line 105: if the comment says `all other revisions are base don berlin`  why not just do this ? 
```suggestion
	specs[tosca.R10_London] = specs[tosca.R09_Berlin]
	specs[tosca.R11_Paris] = specs[tosca.R09_Berlin]
	specs[tosca.R12_Shanghai] = specs[tosca.R09_Berlin]
	specs[tosca.R13_Cancun] = specs[tosca.R09_Berlin]
```

go/interpreter/lfvm/gas_test.go
```diff
@@ -59,3 +59,171 @@ func TestGas_CallGasCalculation(t *testing.T) {
...
+		tosca.R07_Istanbul: makeSpec(15000, 19200, 4200),
+		// source: https://www.evm.codes/?fork=berlin
+		tosca.R09_Berlin: makeSpec(15000, 20000-100, 5000-2100-100),
+		// source: https://www.evm.codes/?fork=london
+		tosca.R10_London: makeSpec(4800, 20000-100, 5000-2100-100),
+	}
+	// All other revisions are based on London.
+	specs[tosca.R11_Paris] = specs[tosca.R10_London]
+	specs[tosca.R12_Shanghai] = specs[tosca.R11_Paris]
+	specs[tosca.R13_Cancun] = specs[tosca.R12_Shanghai]
```
facuMH, line 181:
```suggestion
	specs[tosca.R11_Paris] = specs[tosca.R10_London]
	specs[tosca.R12_Shanghai] = specs[tosca.R10_London]
	specs[tosca.R13_Cancun] = specs[tosca.R10_London]
```

facuMH (changes requested):
I do not thing using literal numbers instead of named constants help readability at all. That aside the refactoring looks good.

go/interpreter/lfvm/gas.go
LuisPH3, line 319: this was quite confusing and it took me some time to realize that the switch in line 324 is not exhaustive. 
Nit: I would move this declaration and the if isberlinOrNewer if into a swtich default clause. to avoid mutation of state (cost varaible),  and having a more declarative function flow.

go/interpreter/lfvm/instructions.go
```diff
@@ -244,45 +244,48 @@ func opMsize(c *context) {
...
 	}
 
 	var key = tosca.Key(c.stack.pop().Bytes32())
 	var value = tosca.Word(c.stack.pop().Bytes32())
 
-	// Perform storage update
-	c.context.SetStorage(c.params.Recipient, key, value)
+	cost := tosca.Gas(0)
+	if c.isAtLeast(tosca.R09_Berlin) {
+		if c.context.AccessStorage(c.params.Recipient, key) == tosca.ColdAccess {
```
LuisPH3, line 258: nit: and + newline removes one level of nesting

go/interpreter/lfvm/instructions.go
```diff
@@ -244,45 +244,48 @@ func opMsize(c *context) {
...
-				return
-			}
-		} else {
-			if !c.useGas(WarmStorageReadCostEIP2929) {
-				return
-			}
+		// charge costs for warm/cold slot access
+		costs := WarmStorageReadCostEIP2929
+		if c.context.AccessStorage(addr, slot) == tosca.ColdAccess {
+			costs = ColdSloadCostEIP2929
```
LuisPH3, line 282: Line 259 writes 2100, here constant. please unify.

go/tosca/revisions.go
```diff
@@ -36,6 +36,17 @@ func (r Revision) String() string {
 	}
 }
 
+func GetAllKnownRevisions() []Revision {
```
LuisPH3, line 39: nit: would a loop based on an upper limit constant be more future-proof?

LuisPH3 (dismissed):
No showstopper. Just some nits.

go/interpreter/lfvm/gas.go
HerbertJordan: Yes, good idea, applied to both functions.

go/interpreter/lfvm/gas_test.go
HerbertJordan: I prefer to update the comment to state that they inherit the specification from their predecessor -- which is more accurate.

go/tosca/revisions.go
HerbertJordan: Done

go/interpreter/lfvm/gas.go
HerbertJordan: I think there is not a lot of value of defining constants that are only to be used at one place. Also, naming is hard and if not done systematically and correctly, adds confusion. As it is now, a function like `getRefundForSstore` maps to each potential use case a constant (numeric) value representing its cost, this way documenting the "meaning" of each of those constants.

For instance, if you delete an added memory location in revision London or later, you get a refund of `-4800`. You can see this by simply looking it up in the function, that is (by now) practically a table. Introducing extra constant names like `gasRefundForDeletedAndReAddedStoragePostLondon` would not help readability. Also, if it ever changes in a revision down the road, the `PostLondon` becomes a lie.

Referencing `EIPs` in constants is adding additional confusion as it is very unclear which EIP was active in past or will be active in future Revisions.

facuMH (dismissed):

facuMH (approved):

PR #746, opened 2024-09-09
Relax comparison of program counters beyond end of code

HerbertJordan:
This PR relaxes the CT state equality condition by treating any program counter beyond the end of the input code as equivalent.

Since any code at a PC beyond the end of the code is to be considered a STOP instruction (OpCode 0x00), it does not matter which actual location is read. Implementations like the LFVM, using their internal code positions, would be unnecessarily complicated by demanding that the PC is identical in such cases.

This issue was recently detected by a full-mode run of the CT.


go/ct/st/state.go
```diff
@@ -256,9 +256,16 @@ func (s *State) Eq(other *State) bool {
 			s.ReturnData == other.ReturnData
 	}
 
+	// All PC beyond the end of the code are equal, as all PCs in this range
+	// point to the STOP instruction.
+	equivalentPc := s.Pc == other.Pc
```
LuisPH3, line 261: nit: because type is bool, a boolean expression as initializer would read better than control flow.

LuisPH3 (approved):

facuMH (dismissed):

simonlechner (dismissed):

go/ct/st/state.go
HerbertJordan: Done

LuisPH3 (approved):

facuMH (approved):

PR #753, opened 2024-09-11
LFVM: interpreter integration test.

LuisPH3:
This PR adds an end-to-end test for the interpreter RUN function, name of file and test TBD.

A test case defines a code and context to check properties for each `OpCode`.

 For the sake of simplicity, they do not test specific behaviour of each opcode,
 but instead, they evaluate correctness of the following parameters:
   - Stack size before and after execution, where specific values are not
     checked (e.g. binary operators stack size is 2 before and 1 after
     execution, specific values vary with each case)
   - PC position after execution
   - Status after execution (stopped, returned, reverted, error...)
   - Gas cost (both static and dynamic, since it's computation reasonability
     is shared between the interpreter and opcodes implementation)
   - Gas refund (if any)
   - Memory TBD
   - call data & return data TBD

### DIFF
--- a/go/interpreter/lfvm/interpreter_intergration_test.go
+++ b/go/interpreter/lfvm/interpreter_intergration_test.go
@@ -43,6 +43,13 @@ type testDeclaration struct {
 	returnData []byte
 }
 
+func (td testDeclaration) name() string {
+	if len(td.nameOverride) == 0 {
+		return td.code[0].opcode.String()
+	}
+	return td.nameOverride
+}
+
 // generateSingleTestCase generates a test case for each opcode.
 // The instances in the slice test functionality provided by the interpreter.
 // For the sake of simplicity, they do not test specific behavior of each opcode,
@@ -332,7 +339,6 @@ func generateTestCases() []testDeclaration {
 	tests = append(tests, []testDeclaration{
 
 		// BALANCE
-
 		{
 			code:  instructions{instruction(BALANCE)},
 			stack: stackSize(1, 1),
@@ -375,7 +381,6 @@ func generateTestCases() []testDeclaration {
 		},
 
 		// EXTCODECOPY
-
 		{
 			code:               instructions{instruction(EXTCODECOPY)},
 			stack:              stackSize(4, 0),
@@ -573,6 +578,7 @@ func generateTestCases() []testDeclaration {
 	retSize := toWord(6)
 
 	tests = append(tests, []testDeclaration{
+		// CALL
 		{
 			code: instructions{instruction(CALL)},
 			stack: stackWithValues([]tosca.Word{
@@ -589,14 +595,13 @@ func generateTestCases() []testDeclaration {
 				mock.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{}, nil)
 			},
 			gas: cost(
-				700 + // static
+				700 + // static cost
 					3 + // mem-expansion
 					9000 - 2300 + // value cost minus call-stipend
 					200 + // value-transfer
 					2300, // call-stipend
 			),
 		},
-
 		{
 			nameOverride: "CALL static context",
 			code:         instructions{instruction(CALL)},
@@ -612,7 +617,6 @@ func generateTestCases() []testDeclaration {
 			gas:                cost(700),
 			staticCtx:          true,
 		},
-
 		{
 			nameOverride: "CALL zero value",
 			code:         instructions{instruction(CALL)},
@@ -628,14 +632,13 @@ func generateTestCases() []testDeclaration {
 				mock.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{}, nil)
 			},
 			gas: cost(
-				700 + // static
+				700 + // static cost
 					3 + // mem-expansion
 					0 - 2300 + // value cost minus call-stipend
 					200 + // value-transfer
 					2300, // call-stipend
 			),
 		},
-
 		{
 			nameOverride: "CALL zero gas",
 			code:         instructions{instruction(CALL)},
@@ -653,13 +656,12 @@ func generateTestCases() []testDeclaration {
 				mock.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{}, nil)
 			},
 			gas: cost(
-				700 + // static
+				700 + // static cost
 					3 + // mem-expansion
 					9000 - 2300 + // value cost minus call-stipend
 					2300, // call-stipend
 			),
 		},
-
 		{
 			nameOverride: "CALL no balance",
 			code:         instructions{instruction(CALL)},
@@ -676,12 +678,11 @@ func generateTestCases() []testDeclaration {
 				mock.EXPECT().GetBalance(gomock.Any()).Return(tosca.Value{}) // not enough balance to transfer value
 			},
 			gas: cost(
-				700 + // static
+				700 + // static cost
 					3 + // mem-expansion
 					9000 - 2300, // value cost minus call-stipend
 			),
 		},
-
 		{
 			nameOverride: "CALL return error",
 			code:         instructions{instruction(CALL)},
@@ -699,14 +700,13 @@ func generateTestCases() []testDeclaration {
 				mock.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{}, fmt.Errorf("error"))
 			},
 			gas: cost(
-				700 + // static
+				700 + // static cost
 					3 + // mem-expansion
 					200 + // value-transfer
 					9000 - 2300 + // value cost minus call-stipend
 					2300, // call-stipend
 			),
 		},
-
 		{
 			nameOverride: "CALL EIP-2929 warm",
 			code:         instructions{instruction(CALL)},
@@ -733,7 +733,6 @@ func generateTestCases() []testDeclaration {
 					2300, // call-stipend
 			),
 		},
-
 		{
 			nameOverride: "CALL EIP-2929 cold",
 			code:         instructions{instruction(CALL)},
@@ -760,7 +759,6 @@ func generateTestCases() []testDeclaration {
 					2300, // call-stipend
 			),
 		},
-
 		{
 			nameOverride: "CALL EIP-2929 new account",
 			code:         instructions{instruction(CALL)},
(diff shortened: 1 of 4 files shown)


go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+	tests = addTrivialTest(tests, gas(6), stackSize(1, 3), PUSH1_DUP1)
+	tests = addTrivialTest(tests, gas(6), stackSize(2, 2), DUP2_LT)
+	tests = addTrivialTest(tests, gas(6), stackSize(3, 3), SWAP2_SWAP1)
+	tests = addTrivialTest(tests, gas(6), stackSize(0, 2), PUSH1_PUSH1)
+	tests = addTrivialTest(tests, gas(10), stackSize(5, 3), POP_SWAP2_SWAP1_POP)
+	tests = addTrivialTest(tests, gas(11), stackSize(4, 3), SWAP1_POP_SWAP2_SWAP1)
+	tests = addTrivialTest(tests, gas(12), stackSize(2, 1), DUP2_MSTORE)
+	tests = addTrivialTest(tests, gas(14), stackSize(5, 3), AND_SWAP1_POP_SWAP2_SWAP1)
+
+	testWithData := func(gas gas, stack stackTest, requiredDataOps int, opCodes ...OpCode) []testDeclaration {
```
facuMH, line 192: is there any gain in merging this closure with `addTrivialTest` ?

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+
+	///////////////////////////////////
+	// Jump instructions
+
+	tests = append(tests, []testDeclaration{
+
+		{
+			nameOverride: "JUMP",
+			code:         []Instruction{{PUSH1, 2 << 8}, {JUMP, 0}, {JUMPDEST, 0}},
+			gas:          cost(3 + 8 + 1),
```
facuMH, line 216: I like very much this manner of describing costs, but I only know where they come from because we discussed it offline, do you think you could add a generic comment explaining how you come up with these sums ?

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+	tests = append(tests, testWithData(15, stackSize(0, 1), 1, PUSH1_PUSH1_PUSH1_SHL_SUB)...)
+	tests = append(tests, testWithData(9, stackSize(1, 4), 2, PUSH1_PUSH4_DUP3)...)
+
+	///////////////////////////////////
+	// Jump instructions
+
+	tests = append(tests, []testDeclaration{
+
+		{
+			nameOverride: "JUMP",
```
facuMH, line 214:
```suggestion
			nameOverride: JUMP.String(),
```
wdyt ?

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+				gas:   cost(gas(375 + (n * 375) + 8*1 + 3)),
+			}
+		}, getOpCodesInRange(LOG0, LOG4))...,
+	)
+
+	///////////////////////////////////
+	// TRIVIAL TESTS introduced in revisions
+
+	tests = append(tests, testDeclaration{
+		code:               []Instruction{{BASEFEE, 0}, {STOP, 0}},
```
facuMH, line 287: what is different between this `STOP` and the one introduced with "Special Opcodes" ?

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+		stack:              stackSize(0, 1),
+		revisionConstraint: validFrom(tosca.R10_London),
+		gas:                cost(gas(2)),
```
facuMH, line 290:
```suggestion
		gas:                cost(2),
```

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+				address,
+				callGas,
+			}, 1),
+			revisionConstraint: validOnlyIn(tosca.R07_Istanbul),
+			mockCalls: func(mock *tosca.MockRunContext) {
+				mock.EXPECT().AccountExists(gomock.Any()).Return(true)              // non-empty account
+				mock.EXPECT().GetBalance(gomock.Any()).Return(tosca.Value(callGas)) // enough balance to transfer value
+				mock.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{}, nil)
+			},
+			gas: cost(gas(
```
facuMH, line 322:
```suggestion
			gas: cost(
```

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+				callGas,
+			}, 1),
+			revisionConstraint: validFrom(tosca.R09_Berlin),
+			mockCalls: func(mock *tosca.MockRunContext) {
+				mock.EXPECT().AccountExists(gomock.Any()).Return(true)              // non-empty account
+				mock.EXPECT().GetBalance(gomock.Any()).Return(tosca.Value(callGas)) // enough balance to transfer value
+				mock.EXPECT().IsAddressInAccessList(gomock.Any()).Return(true)      // warm
+				mock.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{}, nil)
+			},
+			gas: cost(gas(
```
facuMH, line 350:
```suggestion
			gas: cost(
```

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+	for i := 0; i < numOpCodes; i++ {
+		op := OpCode(i)
+		if !re.MatchString(op.String()) {
+			continue
+		}
+
+		if _, ok := testedOps[op]; !ok {
+			t.Errorf("opcode %v is not tested", op)
+		} else {
+			// FIXME: false negatives, some ops just wont exist in older revisions
```
facuMH, line 412: these exceptions are few enough that they could be listed there and this test "fixed" ? I guess this could also be a small follow-up

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+	if rc.min != nil && revision < *rc.min {
+		return false
+	}
+	if rc.max != nil && revision >= *rc.max {
+		return false
+	}
+	return true
+}
+
+// validBefore creates a revision constraint that is valid for revisions greater equal than argument.
```
facuMH, line 517:
```suggestion
// validFrom creates a revision constraint that is valid for revisions greater equal than argument.
```

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+func validFrom(revision tosca.Revision) revisionConstraint {
+	return revisionConstraint{min: &revision}
+}
+
+// validBefore creates a revision constraint that is valid for revisions older than argument.
+func validBefore(revision tosca.Revision) revisionConstraint {
+	return revisionConstraint{max: &revision}
+}
+
+// validInRange creates a revision constraint that is valid only for revision.
```
facuMH, line 527:
```suggestion
// validOnlyIn creates a revision constraint that is valid only for revision.
```

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+	return revisionConstraint{min: &revision, max: &next}
+}
+
+// validInRange creates a revision constraint that is valid in the range [min, max]
+func validInRange(min, max tosca.Revision) revisionConstraint {
+	next := max + 1
+	return revisionConstraint{min: &min, max: &next}
+}
+
+// generateOpCodesInRange generates a slice of opcodes in the range [start, end].
```
facuMH, line 539:
```suggestion
// getOpCodesInRange generates a slice of opcodes in the range [start, end].
```

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+func toKey(value byte) tosca.Key {
+	res := tosca.Key{}
+	res[len(res)-1] = value
+	return res
+}
+
+func toWord(value int) tosca.Word {
+	v := uint256.NewInt(uint64(value))
+	return v.Bytes32()
+}
```
facuMH, line 565: idea for another PR: move this into tosca package if there are similar functions in other packages to prevent duplicated functionality

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+
+type stackTest struct {
```
facuMH, line 567: why not use a `testParameter` ?

go/interpreter/lfvm/interpreter_intergration_test.go
```diff
@@ -0,0 +1,689 @@
...
+	v := statusReturned
+	return &v
+}
+
+func expectInvalidInstruction() *status {
+	v := statusInvalidInstruction
+	return &v
+}
+
+func makeContext(
```
facuMH, line 628: I like this, we have a lot of tests where we insatiate a context and there is always the repeated block.

facuMH (changes requested):
great work !!! just a few nits and comment comments.

go/interpreter/lfvm/interpreter_intergration_test.go
LuisPH3: I rather not call functions which are not a getter get, this is a generator. mabe make...

go/interpreter/lfvm/interpreter_intergration_test.go
LuisPH3: contents are not of the same type, now before is a slice and after a size.

go/interpreter/lfvm/interpreter_intergration_test.go
LuisPH3: This can be done here because it depends on the testDeclaration type, for other tests this may be quite different.

go/interpreter/lfvm/interpreter_intergration_test.go
facuMH: I know but you could use an existing struct and just get the size of after ? it was just a thought, no need to change.

go/interpreter/lfvm/interpreter_intergration_test.go
LuisPH3: nevermind, it seems that i remaned it and you are right. the comment was out of date

LuisPH3:
This has been a proof of concept, new PR will implement a new aproach.

HerbertJordan (commented):
After our offline-discussion I assume this is not going to be the final shape of things. Please ping me if you want to discuss further or re-add me as a reviewer once a revised version is ready.

go/interpreter/lfvm/super_instructions.go
```diff
@@ -119,11 +119,14 @@ func opIsZero_Push2_Jumpi(c *context) {
...
 
 func opSwap2_Swap1_Pop_Jump(c *context) {
-	top := c.stack.pop()
-	c.stack.pop()
-	trg := c.stack.peek()
-	c.pc = int32(trg.Uint64()) - 1
-	*trg = *top
+	// A B C -> C A B
+	// C A B -> C B A
+	// C B
```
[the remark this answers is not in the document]
LuisPH3: i will have a look on that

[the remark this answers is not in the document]
LuisPH3: The unexpected behaviour came from the operation not validating jumpdest, like jump does.

LuisPH3:
This has been a proof of concept, new PR will implement a new approach.

PR #754, opened 2024-09-11
LFVM: memory methods rename part 1

facuMH:
Memory internal functions should not be public.
This PR renames `ExpansionCost` into `getExpansionCost` and `Len` to `length`. 
This PR removes `Data` and `Print` methods, since print is never used and data had a single use that can be replaced by a call to `GetSlice`. 
Follow up PRs will make the other functions in memory private and add tests for them.


simonlechner (approved):
Nice refactoring.

LuisPH3 (approved):

PR #756, opened 2024-09-11
LFVM: rename and test `EnsureCapacity`

facuMH:
`EnsureCapacity` and `EnsureCapacityWithoutGas` did not need to be public methods.
This PR:
- renames them to `expandMemory` and `expandMemoryWithoutCharging` respectively,
- removes `SetWithCapacityCheck` because it was only once and it could just call `expandMemoryWithoutCharging` plus `Set`
- removes `GetSliceWithCapacity` because it was never used.
- adds tests and documentation for both of them


go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +49,134 @@ func TestExpansionCosts(t *testing.T) {
...
+		},
+		"memory-bigger-than-size-changes-nothing": {
+			size:        32,
+			initialMem:  []byte{63: 0x0},
+			expectedMem: []byte{63: 0x0},
+		},
+		"size-zero-changes-nothing": {
+			size:        0,
+			initialMem:  []byte{31: 0x0},
+			expectedMem: []byte{31: 0x0},
```
simonlechner, line 73: This cases stresses the same thing as the case above since the memory is greater than the size. Maybe change the `initialMem` to `[]byte{}`.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +49,134 @@ func TestExpansionCosts(t *testing.T) {
...
+			err := m.expandMemory(test.offset, test.size, &ctxt)
+			memSize := m.Len()
+			if err != nil {
+				t.Fatalf("unexpected error, want: %v, got: %v", nil, err)
+			}
+			// we split memory size check in two cases because when size is zero,
+			// we don't expect the memory to be expanded, regardless of the offset
+			if test.size == 0 && m.Len() != uint64(len(test.initialMemory)) {
+				t.Errorf("memory size should not have changed, want: %d, got: %d", len(test.initialMemory), memSize)
+			} else if want := test.size + test.offset; test.size != 0 && memSize < want {
```
simonlechner, line 177: Is it really required for the memory to be greater and not just greater equal?

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +49,134 @@ func TestExpansionCosts(t *testing.T) {
...
+			ctxt.gas = 100
+
+			err := m.expandMemory(test.offset, test.size, &ctxt)
+			memSize := m.Len()
+			if err != nil {
+				t.Fatalf("unexpected error, want: %v, got: %v", nil, err)
+			}
+			// we split memory size check in two cases because when size is zero,
+			// we don't expect the memory to be expanded, regardless of the offset
+			if test.size == 0 && m.Len() != uint64(len(test.initialMemory)) {
```
simonlechner, line 175: Rather than checking size and explaining it in a comment, the test struct could be extended with a boolean `extendsMemory`.

simonlechner (commented):
I like the renaming of the `EnsureCapacity` function!

go/interpreter/lfvm/memory_test.go
facuMH: yes. because memory expands to exactly `offset+size`, or if existing memory was greater than that addition, then it should remain bigger.

LuisPH3 (approved):
As we discovered in this PR, ensureCapacity has two redundant error reporting mechanisms.
Please make sure that we have appropriate follow up in this issue.

go/interpreter/lfvm/memory_test.go
facuMH: I think adding a parameter to the struct just for this does not improve that much readability. however maybe it helps if I instead move the size check outside:
```
	if test.size == 0 {
		if want := test.size + test.offset; memSize < want {
			t.Errorf("memory size should be bigger than offset+size, want: %d, got: %d", want, memSize)
		}
	} else {
		if m.Len() != uint64(len(test.initialMemory)) {
			t.Errorf("memory size should not have changed, want: %d, got: %d", len(test.initialMemory), memSize)
		}
	}
```

LuisPH3 (dismissed):

go/interpreter/lfvm/memory.go
```diff
@@ -68,47 +68,48 @@ func (m *Memory) ExpansionCosts(size uint64) tosca.Gas {
...
-	if size <= 0 {
-		return
-	}
-	if m.Len() < size {
-		size = toValidMemorySize(size)
-		fee := m.ExpansionCosts(size)
-		m.total_memory_cost += fee
-		m.store = append(m.store, make([]byte, size-m.Len())...)
+// expandMemoryWithoutCharging expands the memory to the given size without checking gas.
+func (m *Memory) expandMemoryWithoutCharging(needed uint64, fee tosca.Gas) {
```
HerbertJordan, line 98: Accepting the expansion `fee` from an outside caller is a violation of the encapsulation principle. I would drop this parameter and re-compute the fee inside the function. The performance impact of this should be negligible.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +49,136 @@ func TestExpansionCosts(t *testing.T) {
...
+			size:        32,
+			initialMem:  []byte{63: 0x0},
+			expectedMem: []byte{63: 0x0},
+		},
+		"size-zero-changes-nothing": {
+			size:        0,
+			initialMem:  []byte{},
+			expectedMem: []byte{},
+		},
+	}
```
HerbertJordan, line 75: Maybe add a test case for `size: 41` making sure that memory is allocated in 32-byte chunks.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +49,136 @@ func TestExpansionCosts(t *testing.T) {
...
+			size:          32,
+			offset:        64,
+			initialMemory: []byte{127: 0x0},
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			ctxt := getEmptyContext()
+			m := NewMemory()
```
HerbertJordan, line 165: The initial memory is not loaded into the memory ...

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +49,136 @@ func TestExpansionCosts(t *testing.T) {
...
+		})
+	}
+}
+
+func TestMemory_expandMemory(t *testing.T) {
+
+	tests := map[string]struct {
+		size          uint64
+		offset        uint64
+		initialMemory []byte
```
HerbertJordan, line 144: The `initialMemory` field is only checked below if `size == 0`, but in all those cases the `initialMemory` field is `nil`. So it does not seem to serve a purpose.

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +49,136 @@ func TestExpansionCosts(t *testing.T) {
...
+				t.Errorf("unexpected error, want: %v, got: %v", test.expected, err)
+			}
+			if m.Len() != 0 {
+				t.Errorf("should have remained empty, got length: %d", m.Len())
+			}
+		})
+	}
+}
+
+func TestMemory_expandMemory(t *testing.T) {
```
HerbertJordan, line 139: Please rework this test case and add a name that declares what this test is supposed to test.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory_test.go
facuMH: reworked test

go/interpreter/lfvm/memory_test.go
facuMH: reworked test

LuisPH3 (dismissed):

go/interpreter/lfvm/memory.go
```diff
@@ -79,27 +83,24 @@ func (m *Memory) EnsureCapacity(offset, size uint64, c *context) error {
...
-	if size <= 0 {
-		return
-	}
-	if m.length() < size {
-		size = toValidMemorySize(size)
-		fee := m.getExpansionCosts(size)
-		m.total_memory_cost += fee
-		m.store = append(m.store, make([]byte, size-m.length())...)
+// expandMemoryWithoutCharging expands the memory to the given size without charging gas.
+func (m *Memory) expandMemoryWithoutCharging(needed uint64) {
```
LuisPH3, line 98: nit: name feels odd. 

`expandMemoryWithoutChargingGas`
or 
`expandMemoryWithoutCharge`
..
`expandMemoryWithoutCost`
`expandMemoryWithoutGasCost`

go/interpreter/lfvm/memory_test.go
LuisPH3, line 1: consistency:

sometimes you call test cases "something-something"
sometimes you use spaces. 

For the shake of readability, we should stick to one of them.

LuisPH3 (approved):

go/interpreter/lfvm/memory_test.go
```diff
@@ -43,7 +45,143 @@ func TestExpansionCosts(t *testing.T) {
...
+			size:               0,
+			offset:             32,
+			expectedMemorySize: 0,
+		},
+		"expand-memory-in-words-length": {
+			size:               13,
+			offset:             0,
+			expectedMemorySize: 32,
+		},
+		"memory-bigger-than-size+offset-does-not-expand": {
```
HerbertJordan, line 163: nit: when starting to mixing `-` and `+` it might be better to use spaces for word separation

go/interpreter/lfvm/memory_test.go
```diff
@@ -43,7 +45,143 @@ func TestExpansionCosts(t *testing.T) {
...
+		})
+	}
+}
+
+func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
+
+	tests := map[string]struct {
+		size               uint64
+		offset             uint64
+		initialMemory      []byte
```
HerbertJordan, line 149: nit: it should be enough to specify the "initialMemorySize" instead of the full memory, as it is not checked that the existing memory is not altered.

HerbertJordan (dismissed):
Looks good, thanks.

go/interpreter/lfvm/memory.go
facuMH: this is described in the documentation, so I will leave the name as is.

simonlechner (approved):

PR #757, opened 2024-09-11
Floria: bug fix, transfer value in call

simonlechner:
During the first history run using lfvm (and therefore floria for handling recursive calls), block 74 failed due to a wrong value transfer in a delegate call. An integration test stressing value transfers in all calls as well as the fix is added in this PR.


go/integration_test/processor/calls_test.go
```diff
@@ -453,6 +453,75 @@ func TestProcessor_CallingNonExistentAccountIsHandledCorrectly(t *testing.T) {
...
+					GasLimit:  sufficientGas,
+					Value:     value,
+				}
+
+				transactionContext := newScenarioContext(state)
+
+				// Run the processor
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil || !result.Success {
+					t.Errorf("execution was not successful or failed with error %v", err)
```
HerbertJordan, line 500: nit: `t.Fatalf`?

go/integration_test/processor/calls_test.go
```diff
@@ -453,6 +453,75 @@ func TestProcessor_CallingNonExistentAccountIsHandledCorrectly(t *testing.T) {
...
+				}
+				if call.callType != vm.STATICCALL && !slices.Equal(result.Output, value[:]) {
```
HerbertJordan, line 502: Should for the static call be checked that the value was indeed zero?

go/integration_test/processor/calls_test.go
```diff
@@ -453,6 +453,75 @@ func TestProcessor_CallingNonExistentAccountIsHandledCorrectly(t *testing.T) {
...
+					t.Errorf("sender balance was not updated correctly, got %v", transactionContext.GetBalance(sender0))
+				}
+				value0 := value
+				value1 := tosca.NewValue(0)
+				if call.callType == vm.CALL {
+					value0 = tosca.NewValue(0)
+					value1 = value
+				}
+				if transactionContext.GetBalance(receiver0).Cmp(value0) != 0 {
+					t.Errorf("receiver balance was not updated correctly, got %v", transactionContext.GetBalance(receiver0))
```
HerbertJordan, line 515: general remark: if yo do not print the want, I would argue the got is of little value in such cases ...

go/processor/floria/run_context.go
```diff
@@ -76,9 +76,11 @@ func (r runContext) Call(kind tosca.CallKind, parameters tosca.CallParameters) (
 	}
 
 	snapshot := r.CreateSnapshot()
-	if err := transferValue(r, parameters.Value, parameters.Sender, recipient); err != nil {
-		r.RestoreSnapshot(snapshot)
-		return tosca.CallResult{}, nil
+	if kind != tosca.DelegateCall && kind != tosca.CallCode {
```
HerbertJordan, line 79: From how I understand the semantic of `CALLCODE` it should also perform the value transfer just like CALL would do. The difference is that in the `CALLCODE` case the sender and receiver are the same. If a local transfer from a sender to itself can never fail, skipping the transfer here is fine. But if it needs to fail if there are  insufficient funds, you can not skip it here.

HerbertJordan (changes requested):
Please check whether `CALLCODE` needs to fail if there are not sufficient funds.

go/processor/floria/run_context.go
LuisPH3: @simonlechner  why is this case (callcode with insufficient funds)  not detected in the unit test. the unit test is called ?

simonlechner: @LuisPH3 there were no tests with value transfers for all call types as they are implemented in the same function. With the newly added changes to the run context different call types are now handled differently and this PR adds tests to ensure this.

HerbertJordan: This is not yet answering my question: In case of a `CallCode`, do we need to call `transferValue` to make sure there is enough balance on the account for the call?

Right now, I guess our interpreter implementations check this pro-actively (e.g. [here](https://github.com/Fantom-foundation/Tosca/blob/ce72f47093f896f9ff5528315060a2cb2df80a17/go/interpreter/lfvm/instructions.go#L1170-L1180)). But I do not think we should rely on this being implemented by all interpreters consistently.

All your tests pass if the condition is restricted to
```
if kind != tosca.DelegateCall {
```

go/processor/floria/run_context.go
```diff
@@ -170,6 +172,10 @@ func transferValue(
 		return fmt.Errorf("insufficient balance: %v < %v", senderBalance, value)
 	}
 
+	if sender == recipient {
```
HerbertJordan, line 175: Move this either to #755 or remove the modification of `transferValue` in #755 and move the test for this function into this PR.

go/integration_test/processor/transfer_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+				transactionContext := newScenarioContext(state)
+
+				// Run the processor
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil {
+					t.Fatalf("execution failed with error %v", err)
+				}
+				if result.Success != test.success {
+					t.Errorf("execution was not successful, want %v, got %v", test.success, result.Success)
```
HerbertJordan, line 199:
```suggestion
					t.Errorf("expected success flag to be %v, got %v", test.success, result.Success)
```

go/processor/floria/run_context.go
HerbertJordan, line 79: I would cover such features with unit tests to specifically target the changed feature. 

Integration tests are good for comparing between implementations and for smoke testing, but identified requirements like "the transfer needs to fail if the sender has not enough funds, even if it is the receiver" or "no value transfer for delegate calls" can be tested more to the point using unit tests.

It would also document these requirements.

HerbertJordan (changes requested):

go/processor/floria/run_context.go
simonlechner: I moved all changed related to value transfers to this PR and updated the unit test.

go/processor/floria/run_context.go
simonlechner: I reworked the integration test dealing with insufficient balance, it now also checks whether the code has been executed. The transaction should fail if a `CallCode` is executed with a value greater than the balance of the sender.

go/processor/floria/run_context.go
```diff
@@ -76,7 +76,7 @@ func (r runContext) Call(kind tosca.CallKind, parameters tosca.CallParameters) (
 	}
 
 	snapshot := r.CreateSnapshot()
-	if kind != tosca.DelegateCall && kind != tosca.CallCode {
+	if kind != tosca.DelegateCall {
```
facuMH, line 79: should this if not check as well for not static call  ?

go/integration_test/processor/transfer_test.go
```diff
@@ -0,0 +1,218 @@
...
+						t.Errorf("transaction has not been handled correctly, code was executed")
+					}
+				}
+
+				zero := tosca.NewValue(0)
+				if balance := transactionContext.GetBalance(sender0); balance.Cmp(zero) != 0 {
+					t.Errorf("sender balance was not updated correctly, want %v, got %v", zero, balance)
+				}
+				if balance := transactionContext.GetBalance(receiver0); balance.Cmp(receiverBalance) != 0 {
+					t.Errorf("receiver balance was not updated correctly, want %v, got %v", receiverBalance, balance)
```
HerbertJordan, line 162: nit: I think this error message is misleading and it through me off in understanding this test ... what you actually check here is that the receiver balance was not modified, right?

go/integration_test/processor/transfer_test.go
```diff
@@ -0,0 +1,218 @@
...
+				}
+				if balance := transactionContext.GetBalance(receiver1); balance.Cmp(zero) != 0 {
+					t.Errorf("receiver balance was not updated correctly, want %v, got %v", zero, balance)
```
HerbertJordan, line 165: same here, right?

go/integration_test/processor/transfer_test.go
```diff
@@ -0,0 +1,218 @@
...
+				// Run the processor
+				result, err := processor.Run(tosca.BlockParameters{}, transaction, transactionContext)
+				if err != nil {
+					t.Fatalf("execution failed with error %v", err)
+				}
+				if result.Success != test.success {
+					t.Errorf("expected success flag to be %v, got %v", test.success, result.Success)
+				}
+				if balance := transactionContext.GetBalance(sender); balance.Cmp(senderBalance) != 0 {
+					t.Errorf("sender balance was not updated correctly, want %v, got %v", senderBalance, balance)
```
HerbertJordan, line 213: nit: also here, there is no expectation that the balance should be updated at all, right?

HerbertJordan (approved):

go/processor/floria/run_context.go
simonlechner: The value should always be 0 in a static call that's why it is not added. I still added it for completeness and to prevent for interpreter implementation errors.

facuMH (approved):

HerbertJordan (approved):

Issue #682, opened 2024-08-26
Add comprehensive unit tests for LFVM's Memory


HerbertJordan:
In particular, make sure that

- [ ] in-variants are defined and maintained by methods
- [ ] memory usage is bound
- [ ] expansion costs are correctly computed
- [ ] capacity is managed correctly
- [ ] reading from and writing to the memory is correctly implemented

facuMH:
Memory invariants defined by functions:
- `getExpansionCosts`:
   . Can not calculate cost of more than `0x1FFFFFFFE0`.
   . Charge 3 gas per word of 32 bytes plus `(w^2)/512`.
- `expandMemory`:
   . Charge for memory before expanding it.
   . Always expand in words of 32 bytes (`toValidMemorySize`).
   . Keep tabs on total memory expansion cost (`memory.currentMemoryCost`).
   . Return errors for overflow or out of gas.
- `getSlice`:
  . Always charge for gas before expanding. 
  . Expanding memory does not modify existing data.
  . Size 0 does not provoke memory expansion.
- Since `readWord` and `set` call to `getSlice`, we get the same assurances and error checking. 

Memory usage is bound:
- This is done by max size check in `getExpansionCosts` and `useGas` in `expandMemory`.
- Tested in `TestGetExpansionCosts`, `TestMemory_expandMemory_ErrorCases`, multiple tests checking for error propagation from `readWord` and set.

Expansion costs are correctly computed:
- Done in `getExpansionCosts`.
- Checked by set fail test, read word error cases, `getSlice` error cases and getExpansionCosts test.

Capacity is managed correctly:
- This involves increasing the memory in chunks of 32 bytes (words), not increasing capacity when size is zero regardless of the offset and not increasing capacity if size+offset < current capacity, memory expansion does not overwrite existing data and newly allocated memory is zeroed.
- Handled by `expandMemory`, and `getSlice`.
[...]

PR #759, opened 2024-09-12
LFVM: `set` and `setByte` test

facuMH:
In this PR:
- `SetByte` and `SetWord` are now private functions.
- `Set` functions is removed since we are moving towards enforcing that memory can not attempt to read/write without expanding first.
- `SetWithCapacityAndGasCheck` is renamed to `set`, simplified to simply call `getSlice` from #761 and then copy from input value to the slice.
- `setByte` and `setWord` is implemented as a specialization of `setWithCapacityAndGasCheck`.
- tests are added for `set`, `setByte` and `setWord`.

This is part of #682
This PR will rework set to be a specialization+copy of `getSlice` from #761

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -210,7 +210,7 @@ func opMcopy(c *context) error {
 	if err != nil {
 		return err
 	}
-	if err := c.memory.setWithCapacityAndGasCheck(destOffset, size, data, c); err != nil {
+	if err := c.memory.set(destOffset, size, data, c); err != nil {
 		return err
 	}
 	return nil
@@ -384,7 +384,7 @@ func opCallDataCopy(c *context) error {
 		return err
 	}
 
-	return c.memory.set(memOffset64, length64, getData(c.params.Input, dataOffset64, length64))
+	return c.memory.set(memOffset64, length64, getData(c.params.Input, dataOffset64, length64), c)
 }
 
 func opAnd(c *context) {
@@ -814,7 +814,7 @@ func opCodeCopy(c *context) error {
 		return err
 	}
 	codeCopy := getData(c.params.Code, uint64CodeOffset, length.Uint64())
-	return c.memory.set(memOffset.Uint64(), length.Uint64(), codeCopy)
+	return c.memory.set(memOffset.Uint64(), length.Uint64(), codeCopy, c)
 }
 
 func opExtcodesize(c *context) error {
@@ -1009,7 +1009,7 @@ func opExtCodeCopy(c *context) error {
 		return err
 	}
 	codeCopy := getData(c.context.GetCode(address), uint64CodeOffset, length.Uint64())
-	return c.memory.set(memOffset.Uint64(), length.Uint64(), codeCopy)
+	return c.memory.set(memOffset.Uint64(), length.Uint64(), codeCopy, c)
 }
 
 func checkSizeOffsetUint64Overflow(offset, size *uint256.Int) error {
@@ -1224,7 +1224,7 @@ func opReturnDataCopy(c *context) error {
 		return errOutOfGas
 	}
 
-	return c.memory.setWithCapacityAndGasCheck(memOffset.Uint64(), length.Uint64(), c.returnData[offset64:end64], c)
+	return c.memory.set(memOffset.Uint64(), length.Uint64(), c.returnData[offset64:end64], c)
 }
 
 func opLog(c *context, size int) error {
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -380,11 +380,13 @@ func opCallDataCopy(c *context) error {
 		return err
 	}
 
-	if err := c.memory.expandMemory(memOffset64, length64, c); err != nil {
+	data, err := c.memory.getSlice(memOffset64, length64, c)
+	if err != nil {
 		return err
 	}
-
-	return c.memory.set(memOffset64, length64, getData(c.params.Input, dataOffset64, length64), c)
+	codeCopy := getData(c.params.Input, dataOffset64, length64)
+	copy(data, codeCopy)
+	return nil
 }
 
 func opAnd(c *context) {
@@ -810,11 +812,13 @@ func opCodeCopy(c *context) error {
 		return err
 	}
 
-	if err := c.memory.expandMemory(memOffset.Uint64(), length.Uint64(), c); err != nil {
+	data, err := c.memory.getSlice(memOffset.Uint64(), length.Uint64(), c)
+	if err != nil {
 		return err
 	}
 	codeCopy := getData(c.params.Code, uint64CodeOffset, length.Uint64())
-	return c.memory.set(memOffset.Uint64(), length.Uint64(), codeCopy, c)
+	copy(data, codeCopy)
+	return nil
 }
 
 func opExtcodesize(c *context) error {
@@ -1005,11 +1009,13 @@ func opExtCodeCopy(c *context) error {
 		uint64CodeOffset = math.MaxUint64
 	}
 
-	if err := c.memory.expandMemory(memOffset.Uint64(), length.Uint64(), c); err != nil {
+	data, err := c.memory.getSlice(memOffset.Uint64(), length.Uint64(), c)
+	if err != nil {
 		return err
 	}
 	codeCopy := getData(c.context.GetCode(address), uint64CodeOffset, length.Uint64())
-	return c.memory.set(memOffset.Uint64(), length.Uint64(), codeCopy, c)
+	copy(data, codeCopy)
+	return nil
 }
 
 func checkSizeOffsetUint64Overflow(offset, size *uint256.Int) error {
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -176,7 +176,7 @@ func opMstore8(c *context) error {
 	if overflow {
 		return errOverflow
 	}
-	return c.memory.SetByte(offset, byte(value.Uint64()), c)
+	return c.memory.setByte(offset, byte(value.Uint64()), c)
 }
 
 func opMcopy(c *context) error {
@@ -210,7 +210,7 @@ func opMcopy(c *context) error {
 	if err != nil {
 		return err
 	}
-	if err := c.memory.SetWithCapacityAndGasCheck(destOffset, size, data, c); err != nil {
+	if err := c.memory.setWithCapacityAndGasCheck(destOffset, size, data, c); err != nil {
 		return err
 	}
 	return nil
@@ -1224,7 +1224,7 @@ func opReturnDataCopy(c *context) error {
 		return errOutOfGas
 	}
 
-	return c.memory.SetWithCapacityAndGasCheck(memOffset.Uint64(), length.Uint64(), c.returnData[offset64:end64], c)
+	return c.memory.setWithCapacityAndGasCheck(memOffset.Uint64(), length.Uint64(), c.returnData[offset64:end64], c)
 }
 
 func opLog(c *context, size int) error {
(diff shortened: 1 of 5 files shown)


LuisPH3 (dismissed):

go/interpreter/lfvm/memory.go
```diff
@@ -171,29 +164,33 @@ func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
...
 		}
 		copy(m.store[offset:offset+size], value)
 	}
 	return nil
 }
 
-func (m *Memory) SetWithCapacityAndGasCheck(offset, size uint64, value []byte, c *context) error {
+func makeInsufficientMemoryError(memSize, size, offset uint64) error {
+	return tosca.ConstError(fmt.Sprintf("memory too small, size %d, attempted to write %d bytes at %d", memSize, size, offset))
+}
```
HerbertJordan, line 184: To be honest, I am not a fan of introducing these error-factory functions with the main purpose of allowing tests to produce the same error for checking. It adds extra code and symbols in the namespace that should not be needed.

I think in this case a constant `errInsufficientMemory` would be enough, since no where in the code anybody would ever check for the memSize, size, or offset. It would require parsing the string, which is not a great idea. And human consumers will not greatly benefit from this information either.

The constant `errInsufficientMemory` constant could also be avoided if you would check for the error string to contain some "memory too small" in the test and ignore the rest. In this case, you could use `fmt.Errorf(..)` above to create the error string with the `memSize`, `size`, and `offset` as listed here. This eliminates the `errors.Is(...)`, but I think it is a cleaner solution if the error is not carrying any additional information for the caller than something went wrong and only in the test you want to distinguish what went wrong.

go/interpreter/lfvm/memory.go
```diff
@@ -171,29 +164,33 @@ func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
...
+
+// setWithCapacityAndGasCheck sets the given value at the given offset, expands memory as needed and charges for it.
+// Returns error if insufficient gas or offset+size overflows.
+func (m *Memory) setWithCapacityAndGasCheck(offset, size uint64, value []byte, c *context) error {
```
HerbertJordan, line 188: `WitchCapacity` and `GasCheck` is a kinda redundant. If gas is charged, there can always be enough capacity, right?

How about renaming this one to `set` as this is the safe default and rename the current `set` to `trySet` or `update` ...

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +185,120 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
+			memory: baseData[:4],
+			offset: 8,
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			ctxt := getEmptyContext()
+			m := NewMemory()
+			m.store = test.memory
```
HerbertJordan, line 219: I commented on this before: when you assign the `test.memory` to `m.store` here and `test.memory` is an alias of `baseData`, the operation `m.setByte` below will write into `baseData`. Since all the tests are pointing to the same `baseData`, this makes them all inter-dependent.

If you want to check that this is true, print `baseData` before each test. It changes its value.

In this test case it is not a problem. But it would be if, for instance, you check that only this one byte got changed (which would be a nice addition).

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +185,120 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
+			ctxt.gas = tosca.Gas(3)
+			value := byte(0x12)
```
HerbertJordan, line 221: nit: In the first test, byte [0] in the memory has already the value "0x12" before the update.

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +185,120 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
+		t.Errorf("unexpected error, want: %v, got: %v", errGasUintOverflow, err)
+	}
+
+}
+
+func TestMemory_Set_SuccessfulCases(t *testing.T) {
+
+	memoryOriginalSize := uint64(8)
+	offset := uint64(1)
+	// data of size (memoryOriginalSize - offset), to ensure it would fit in memory
```
HerbertJordan, line 258: nit: left-over of former test table-driven test case?

facuMH: no, just explaining why `data` has that specific size.

go/interpreter/lfvm/memory.go
facuMH: discussed offline. I will go for the suggested `fmt.Errorf`.

HerbertJordan (dismissed):
The tests could still be written more compact and comprehensive, but I think it is good enough now. Approved.

LuisPH3 (changes requested):
This is not good naming for the `set` operation: if we did not call `get` `tryGet` we should not call `set `trySet`
Because the work done in get, the semantics here should be symmetric. `set` will always expand except for call. 
all these expand +  try could be cleaned up

go/interpreter/lfvm/instructions.go
```diff
@@ -167,7 +167,7 @@ func opMstore(c *context) {
 		c.status = statusError
 		return
 	}
-	if err := c.memory.SetWord(offset, value, c); err != nil {
+	if c.memory.SetWord(offset, value, c) != nil {
```
HerbertJordan, line 170: Something to discuss: I kinda prefer the pattern
```
if err := call(..); err != nil {
```
over
```
if call(..) != nil {
```
since it indicates to the reader that what is returned is an error. Even if the `err` is not used below.
If you do not name the result, it could be anything that is of a reference type; a pointer, am map, a list, an interface, ...

HerbertJordan (changes requested):
The change to the memory interface is nice. Tests are missing a few things.

For instance, I think the following properties are missing for testing `set`:
- the 32-step increment in size for anything that is not the first 32 bytes
- check that new memory is indeed 0
- check that during a re-size the old data is preserved
- also, I think a test that makes sure that if `size == 0` nothing happens is missing

These are all properties of the general `set` that should be `thoroughly` tested. You do not need to repeat those tests for `setByte`. For this function, tests should focus on the fact that `size` is fixed to 1 and the offset and potentiall errors are propagated correctly. For this, it should be enough to run a few examples.

go/interpreter/lfvm/instructions.go
facuMH: my motivation to delete the `err` in these kind of checks was that if we declare an error and compare it, it gives me the impression it should be propagated, but in these checks, we always do `if err := fun();  err != nil { c.signalErr(); return }` and that seemed very odd in my eyes. But I do understand that then one has to check what are all these functions returning. Either way seems like a compromise.

LuisPH3: In my work [here](https://github.com/Fantom-foundation/Tosca/compare/main...luis/lfvm_interpreter_test) I replaced everything back to `err := expresion; err != nil`. The reason being that errors continue propagating, and the if purpose is only to break execution.

IMO: an ignored error is a code smell, if we do not use errors, we could just return a boolean.

go/interpreter/lfvm/memory.go
```diff
@@ -222,22 +121,57 @@ func (m *Memory) CopyData(offset uint64, trg []byte) {
 	}
 }
 
-func (m *Memory) GetSlice(offset, size uint64) []byte {
+// getSlice obtains a slice of size bytes from the memory at the given offset.
+// Expands memory as needed and charges for it.
```
HerbertJordan, line 125: This [comment](https://github.com/Fantom-foundation/Tosca/pull/761#discussion_r1766363076) got lost.

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +192,272 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
 		})
 	}
 }
+
+func TestMemory_getSlice_ErrorCases(t *testing.T) {
+	c := getEmptyContext()
+	c.gas = 0
```
HerbertJordan, line 198: As I just realized, `getEmptyContext()` does not return an empty context. It returns a context with some `gas`. This is very misleading. Why not simpler
```suggestion
	c := &context{}
```

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +192,272 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
+				if size+offset > 32 {
+					want = uint64(32 * (increment + 2))
+				}
+				if m.length() != want {
+					t.Errorf("unexpected slice length: %d, want: %d", m.length(), 32*increment)
+				}
+			}
+		}
+	}
+}
```
HerbertJordan, line 278: I reviewed this tests also in #761 -- please get your PRs in sync and only request reviews for relevant ones.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory.go
LuisPH3, line 1: This file turned out quite elegant.  :+1:

LuisPH3 (approved):

go/interpreter/lfvm/memory_test.go
```diff
@@ -357,3 +355,143 @@ func TestToValidMemorySize_RoundsUpToNextMultipleOf32(t *testing.T) {
...
+	m := NewMemory()
+	m.store = bytes.Clone(memory)
+	value := byte(0xff)
+
+	err := m.setByte(offset, value, &ctxt)
+	if err != nil {
+		t.Errorf("unexpected error, want: %v, got: %v", nil, err)
+	}
+	if m.length() <= offset {
+		t.Errorf("unexpected memory size, want: %d, got: %d", offset, m.length())
```
HerbertJordan, line 436: This message is wrong. The memory size should not be the offset.

go/interpreter/lfvm/memory_test.go
```diff
@@ -357,3 +355,143 @@ func TestToValidMemorySize_RoundsUpToNextMultipleOf32(t *testing.T) {
...
+	}
+	for i, b := range memory {
+		if i == int(offset) {
```
HerbertJordan, line 439: `i` is never the offset ... the offset is `8` and `i` is only in the range 0-3.

HerbertJordan (changes requested):
Please improve the quality of the test cases.

go/interpreter/lfvm/memory_test.go
facuMH: right this was supposed to range in `m.store`, will fix.

LuisPH3 (dismissed):

LuisPH3 (dismissed):

go/interpreter/lfvm/memory.go
```diff
@@ -223,3 +134,30 @@ func (m *Memory) readWord(offset uint64, target *uint256.Int, c *context) error
...
+
+// set copies the given value into memory at the given offset.
+// Expands memory as needed and charges for it.
+// Returns error if insufficient gas or offset+size overflows.
+func (m *Memory) set(offset, size uint64, value []byte, c *context) error {
+	data, err := m.getSlice(offset, size, c)
+	if err != nil {
+		return err
+	}
+	copy(data, value)
```
HerbertJordan, line 146: How is this function supposed to behave if `value` is shorter than `size`? If zeros should be padded, this needs to be implemented.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory.go
facuMH: `getSlice` takes care of that.

simonlechner (approved):

HerbertJordan (approved):

LuisPH3 (approved):
:+1:

PR #760, opened 2024-09-12
LFVM: `setWord` tests

facuMH:
In this PR:
- makes `setWord` a private function
- adds tests for success and error cases for `setWord`

This is part of #682
This PR depends on #759

### DIFF
--- a/go/interpreter/lfvm/memory_test.go
+++ b/go/interpreter/lfvm/memory_test.go
@@ -128,7 +128,6 @@ func TestMemory_expandMemory_ErrorCases(t *testing.T) {
 
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
-
 			ctxt := getEmptyContext()
 			m := NewMemory()
 			ctxt.gas = test.gas


go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +50,61 @@ func TestExpansionCosts(t *testing.T) {
...
 	}
 }
+
+func TestMemory_SetWord(t *testing.T) {
+
+	tests := map[string]struct {
+		offset uint64
+		memory []byte
+	}{
+		// test for memory expansion
```
LuisPH3, line 60: these comments dont tell much, the name of the test should say what it does.

LuisPH3 (approved):

go/interpreter/lfvm/memory_test.go
facuMH: I think it will be better to use the comments as test names, as you mentioned

go/interpreter/lfvm/memory_test.go
```diff
@@ -47,3 +50,58 @@ func TestExpansionCosts(t *testing.T) {
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			ctxt := getEmptyContext()
+			m := NewMemory()
+			m.store = test.memory
+			// generate non zero bytes value
```
LuisPH3, line 73: is there any difference with `[]byte{31: 0x01} ` it will not be zeroed as well.

LuisPH3 (approved):

go/interpreter/lfvm/memory_test.go
facuMH: this is what I meant to do, yes.

facuMH: well, the idea here was to have all bytes be non zero. doing `[]byte{31: 0x01}` will create a byte slice with 31 `0x0` and then one `0x01`

LuisPH3 (approved):

LuisPH3 (approved):

go/interpreter/lfvm/memory.go
```diff
@@ -121,16 +121,14 @@ func (m *Memory) SetByte(offset uint64, value byte, c *context) error {
 	return nil
 }
 
-func (m *Memory) SetWord(offset uint64, value *uint256.Int, c *context) error {
+// setWord sets a 32-byte word at the given offset, expanding memory as needed and charging for it.
+// Returns error if insufficient gas or offset+32 overflows.
+func (m *Memory) setWord(offset uint64, value *uint256.Int, c *context) error {
```
HerbertJordan, line 126: Once #759 is merged, the body of this function could likely be simplified to something like
```
data := value.Bytes32()
return m.set(offset, 32, data[:], c)
```
Whether we should go for this depends on whether the promised speed-up statement below still holds up. Would be interesting to add a micro-benchmark comparing the 2-line version listed above with the unrolled version that is implemented here right now.

HerbertJordan: I experimented with the following benchmark:
```
func BenchmarkMemory_SetWord(b *testing.B) {
	m := NewMemory()
	m.expandMemoryWithoutCharging(32)
	ctxt := getEmptyContext()
	word := new(uint256.Int)
	for i := 0; i < b.N; i++ {
		m.SetWord(0, word, &ctxt)
	}
}
```
and the results for the current version vs. the 2-line version looks like this:
```
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/interpreter/lfvm
                  │ specialized.txt │            generic.txt             │
                  │     sec/op      │   sec/op     vs base               │
Memory_SetWord-12       15.56n ± 4%   16.99n ± 2%  +9.19% (p=0.000 n=10)
```
The difference is ~1.5 nano seconds, far from the claimed 7x improvement. I would argue that this small difference (in absolute terms) is not worth the extra complexity. Wdyt?

HerbertJordan (changes requested):
Looks good. However, I think we could remove some complexity by eliminating the special case handling for words. Wdyt?

go/interpreter/lfvm/memory.go
facuMH: I did not change it originally because of the promised speedup. I should have made the benchmark an tested it, yes. I would say it depends on how often is mload8 called. do we have this statistic ?

facuMH: the number of times this `MSTORE8` is called is not significant enough for this performance degradation to have any impact. I will do as you suggested and implment `setWord` as a specialization of `set`, just as `setByte` is in #759

PR #761, opened 2024-09-12
LFVM: CopyWord and GetSlice* test

facuMH:
This PR:
- Rename `CopyWord` to `readWord`, `GetSliceWithCapacityAndGas` to `readSliceAndExpandMemory` and `GetSlice` to `readSlice`
- Re-implement `readWord` as an especialization of `readSliceAndExpandMemory`.
- Add test for `readWord`.


go/interpreter/lfvm/instructions.go
```diff
@@ -922,7 +922,7 @@ func opCreate(c *context) {
 		}
 	}
 
-	input := c.memory.GetSlice(offset.Uint64(), size.Uint64())
+	input := c.memory.readSlice(offset.Uint64(), size.Uint64())
```
LuisPH3, line 925: See readSlice comment first.
Expansion done in line 906.

go/interpreter/lfvm/instructions.go
```diff
@@ -999,7 +999,7 @@ func opCreate2(c *context) {
...
```
LuisPH3, line 1002: See readSlice comment:
expansion done in line 977

go/interpreter/lfvm/instructions.go
```diff
@@ -1216,7 +1216,7 @@ func genericCall(c *context, kind tosca.CallKind) {
 	}
 
 	// Get arguments from the memory.
-	args := c.memory.GetSlice(inOffset.Uint64(), inSize.Uint64())
+	args := c.memory.readSlice(inOffset.Uint64(), inSize.Uint64())
```
LuisPH3, line 1219: Most probably read+expansion would have no effect here.

go/interpreter/lfvm/memory.go
```diff
@@ -224,7 +224,9 @@ func (m *Memory) CopyData(offset uint64, trg []byte) {
 	}
 }
 
-func (m *Memory) GetSlice(offset, size uint64) []byte {
+// readSlice reads size bytes from the memory at the given offset.
+// Returns nil if offset+size is out of bounds or size is 0.
+func (m *Memory) readSlice(offset, size uint64) []byte {
```
LuisPH3, line 229: I would refactor this, to enforce the difference between normal use of memory and extended introspection used by test/ct-adapter:

- During normal operation, any read must expand memory first. Having this method invites calling `readSlice` and developer has to think if expansion needs to be done or not, this is not needed. 
- During ct-adapter/testing, a method get rawMemory returning the complete memory can be used and using slice operators to scope it would make the job (or even accessing internal slice directly) 

Factoring this code inside `readSliceAndExpandMemory` would remove the need for these checks. 
- size==0 yields empty slice
- mandatory expansion removes the need to check size.

LuisPH3 (changes requested):
Please reconsider `readSlice` & `readSliceAndExpandMemory` with one of the following options:
- use only `readSliceAndExpandMemory` everywhere (call needs to be evaluated) 
- remove `readSliceAndExpandMemory` and use  `readSlice` + `expandMemory` everywhere. 

I think the first option would be easier.

go/interpreter/lfvm/memory.go
```diff
@@ -224,22 +212,30 @@ func (m *Memory) CopyData(offset uint64, trg []byte) {
 	}
 }
 
-func (m *Memory) GetSlice(offset, size uint64) []byte {
-	if size == 0 {
-		return nil
+// readSliceAndExpandMemory reads size bytes from the memory at the given offset.
+// Expands memory as needed and charges for it.
+// Returns error in case of not enough gas or offset+32 overflow.
+func (m *Memory) readSliceAndExpandMemory(offset, size uint64, c *context) ([]byte, error) {
```
HerbertJordan, line 218: I would associate a `read` prefix with a function to which I can pass the target of the read operation, like it is done for the `readWord` function below, or in Go's [Reader](https://pkg.go.dev/io#Reader) interface.

Something that returns me a part of some state is a valid `getSlice`. 

The `AndExpandMemory` could also be dropped since there is no `readSlice` or `getSlice` counterpart.

go/interpreter/lfvm/instructions.go
```diff
@@ -1216,7 +1218,11 @@ func genericCall(c *context, kind tosca.CallKind) {
 	}
 
 	// Get arguments from the memory.
-	args := c.memory.GetSlice(inOffset.Uint64(), inSize.Uint64())
+	args, err := c.memory.readSliceAndExpandMemory(inOffset.Uint64(), inSize.Uint64(), c)
```
HerbertJordan, line 1221: Could the `expandMemoryWithoutCharging(..)` of line 1195 be removed since the memory expansion is done by this call?

I guess  not, since if the balance check in line 1204 fails, the execution stops without an error but gas for the memory expansion was charged without carrying it out. Best option would be to move this call up line 1195.

go/interpreter/lfvm/memory_test.go
HerbertJordan, line 1: As discussed this morning, if you have something like
```
func doSpecialCase(..) {
   doGeneralCase(...)
}

func doGeneralCase(...) {
   ...
}
```
The bulk of the test cases should probably focus on the generic case and a few cases on the special case. Right now, all tests target the special case but one.

HerbertJordan (changes requested):

go/interpreter/lfvm/instructions.go
facuMH: I would mainly like to keep them apart since the call to `expandMemoryWithoutCharging` is expanding memory to be large enough for both input and output, but this call to `readSliceAndExpand` is only calling with the input size. Yes, this call will never trigger an expansion, but this is ok.

go/interpreter/lfvm/memory.go
LuisPH3: Sorry, I did not propose any renaming. Because the returned slice is a reference to the mem storage, I guess it is a better name than readSlice

go/interpreter/lfvm/instructions.go
```diff
@@ -1216,7 +1218,11 @@ func genericCall(c *context, kind tosca.CallKind) {
 	}
 
 	// Get arguments from the memory.
-	args := c.memory.GetSlice(inOffset.Uint64(), inSize.Uint64())
+	args, err := c.memory.getSlice(inOffset.Uint64(), inSize.Uint64(), c)
```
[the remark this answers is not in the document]
facuMH: I do not like ignoring the error, nor doing explicit handling of storage when there should be a method for it. For me the right solution would be to have a `getSliceWithoutCharging` and call that one from within `getSlice`.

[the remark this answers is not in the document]
LuisPH3: i agree

go/interpreter/lfvm/memory.go
```diff
@@ -224,22 +212,30 @@ func (m *Memory) CopyData(offset uint64, trg []byte) {
 	}
 }
 
-func (m *Memory) GetSlice(offset, size uint64) []byte {
-	if size == 0 {
-		return nil
+// getSlice reads size bytes from the memory at the given offset.
```
HerbertJordan, line 215:
```suggestion
// getSlice obtains a slice of size bytes from the memory at the given offset.
// The returned slice is backed by the memory's internal data. Updates to the
// slice will thus effect the memory states. This connection is invalidated by any
// subsequent memory operation that may change the size of the memory.
```

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +186,97 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
+	if !errors.Is(err, errOutOfGas) {
+		t.Errorf("error should be errOutOfGas, instead is: %v", err)
+	}
+	_, err = m.getSlice(math.MaxUint64-31, 32, &c)
+	if !errors.Is(err, errGasUintOverflow) {
+		t.Errorf("error should be errGasUintOverflow, instead is: %v", err)
+	}
+}
+
+func TestMemory_getSlice(t *testing.T) {
```
HerbertJordan, line 204: nit: please name the property that is covered by this test, not just the function.

go/interpreter/lfvm/memory_test.go
HerbertJordan, line 1: Things that should also be tested for `getSlice`:
- the 32-step increment in size for anything that is not the first 32 bytes
- check that new memory is indeed 0
- check that during a re-size the old data is preserved
- if `offset > length` and `size == 0` no gas is charged and memory does not grow

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +186,97 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
+	c := getEmptyContext()
+	c.gas = 0
+	m := NewMemory()
+	originalTarget := new(uint256.Int)
+	target := originalTarget.Clone()
+	err := m.readWord(math.MaxUint64-31, target, &c)
+	if !errors.Is(err, errGasUintOverflow) {
+		t.Errorf("error should be errGasUintOverflow, instead is: %v", err)
+	}
+	if target.Cmp(originalTarget) != 0 {
```
HerbertJordan, line 272: Checking for something that has not changed is not a strong test case for a `readWord` operation. If the implementation would not transfer the memory content to the target, the test would still pass.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory_test.go
facuMH: the point of this particular test is that whenever an error occurs, the corresponding error is propagated and the target memory is not modified.

go/interpreter/lfvm/memory_test.go
```diff
@@ -185,3 +192,140 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
+				if size+offset > 32 {
+					want = uint64(32 * (increment + 2))
+				}
+				if m.length() != want {
+					t.Errorf("unexpected slice length: %d, want: %d", m.length(), 32*increment)
+				}
+			}
+		}
+	}
+}
```
HerbertJordan, line 278: This would be easier to comprehend if you would directly iterate over the 3 input parameters `memSize`, `offset` and `size`. Also, factoring out the special case of `size` zero reduces complexity, and you can use `toValidMemorySize` to simplify the code:
```
func TestMemory_getSlice_ExpandsMemoryIn32ByteChunks(t *testing.T) {
	for memSize := uint64(0); memSize < 128; memSize += 32 {
		for offset := 0; offset < 128; offset++ {
			for size := 1; size < 32; size++ {
				c := &context{gas: 5000}
				m := NewMemory()
				m.store = make([]byte, memSize)
				_, err := m.getSlice(uint64(offset), uint64(size), c)
				if err != nil {
					t.Errorf("unexpected error: %v", err)
				}
				want := toValidMemorySize(uint64(offset + size))
				if want < memSize {
					want = memSize
				}
				if got, want := m.length(), want; got != want {
					t.Errorf("unexpected memory length: %d, want: %d", got, want)
				}
			}
		}
	}
}

func TestMemory_getSlice_SizeOfZeroIsNotGrowingTheMemory(t *testing.T) {
	for memSize := 0; memSize < 128; memSize += 32 {
		for offset := 0; offset < 128; offset++ {
			c := &context{gas: 5000}
			m := NewMemory()
			m.store = make([]byte, memSize)
			_, err := m.getSlice(uint64(offset), 0, c)
			if err != nil {
				t.Errorf("unexpected error: %v", err)
			}
			if c.gas != 5000 {
				t.Error("no gas should have been consumed when size is zero.")
			}
			if got, want := m.length(), uint64(memSize); got != want {
				t.Errorf("unexpected memory length: %d, want: %d", got, want)
			}
		}
	}
}
```

To make valid use of `toValidMemorySize`, this function should also get some test cases. For instance this one:
```

func TestToValidMemorySize_RoundsUpToNextMultipleOf32(t *testing.T) {
	for i := uint64(0); i < 128; i++ {
		got := toValidMemorySize(i)
		if got%32 != 0 {
			t.Errorf("result should be a multiple of 32, got: %d", got)
		}
		if got < i {
			t.Errorf("result should be greater or equal to input, got: %d", got)
		}
		if got-i >= 32 {
			t.Errorf("result should be less than 32 bytes greater than input, got: %d", got)
		}
	}
}
```

go/interpreter/lfvm/memory_test.go
HerbertJordan: In this case `zero` is also not a strong test case, since an inadvertent clearing effect would not be detected.

HerbertJordan (changes requested):

go/interpreter/lfvm/memory_test.go
```diff
@@ -17,8 +17,16 @@ import (
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/holiman/uint256"
 )
 
+func TestMemory_NewMemoryIsEmpty(t *testing.T) {
+	m := NewMemory()
+	if m.length() != 0 {
```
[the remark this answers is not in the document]
facuMH: this is covered by `TestMemory_readWord_ReadAndPadWithZeros`

go/interpreter/lfvm/memory_test.go
facuMH: right. That I can change.

LuisPH3 (approved):

go/interpreter/lfvm/memory_test.go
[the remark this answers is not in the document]
HerbertJordan: It is not. This test only checks that parts of the word that are read are zero-padded. But it does not check that other memory allocated by the operation is indeed zero.

If on a memory that is 32 bytes long you read a word from position 64-96 you might check that the bytes `[64:96)` are zero, but not that the bytes from `[32:64)` are zero.

HerbertJordan (approved):

simonlechner (approved):

PR #764, opened 2024-09-13
Zero-initialize stack values in init code size test

HerbertJordan:
This PR fixes a lack of stack-value initialization in the test `TestCreateShanghaiInitCodeSize`.

The test assumed implicitly that a newly created stack is filled with zeros. However, due to stack-recycling, this is not the case. The content is actually undefined. This lead the given test to be flaky, since one of the stack values assumed to be zero leads the following issue if it is not zero:
```
--- FAIL: TestCreateShanghaiDeploymentCost (0.00s)
    instructions.go:888: Unexpected call to *tosca.MockRunContext.GetBalance([0x0100000000000000000000000000000000000000]) at /home/herbert/coding/fantom/tosca/go/interpreter/lfvm/instructions.go:888 because: there are no expected calls of the method "GetBalance" for that receiver
    controller.go:98: missing call(s) to *tosca.MockRunContext.Call(is equal to create (tosca.CallKind), is anything) /home/herbert/coding/fantom/tosca/go/interpreter/lfvm/instructions_test.go:720
    controller.go:98: aborting test due to missing call(s)
```

The issue could be reproduced using
```
go test --race ./interpreter/lfvm/... -count 100
```

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -646,8 +646,9 @@ func TestCreateShanghaiInitCodeSize(t *testing.T) {
 			}
 
 			// Prepare stack arguments.
-			ctxt.stack.stackPointer = 3
-			ctxt.stack.data[0].Set(uint256.NewInt(test.init_code_size))
+			ctxt.stack.push(uint256.NewInt(test.init_code_size))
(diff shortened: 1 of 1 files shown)


facuMH (approved):
nice catch !! will keep it in mind for future testing

PR #765, opened 2024-09-13
Replace interpreter instances with interpreter factories in registry

HerbertJordan:
This PR updates the interpreter registry to contain factories instead of instantiated interpreters. This provides the following benefits:
- EVM configurations do not consume resources if they are not used
- when creating an EVM, client-code defined properties (e.g. cache sizes) can be passed the construction process of the interpreter instances (prepared but not implemented by this PR)

**Note to reviewers:** start with the changes in the Tosca package, which is the main part of this PR. All the rest is just API adaptations.

TODO:
- [x] check backward compatibility in Aida
- [x] check backward compatibility in Sonic

### DIFF
--- a/go/integration_test/interpreter/vm_test.go
+++ b/go/integration_test/interpreter/vm_test.go
@@ -41,7 +41,10 @@ var (
 func TestExamples_ComputesCorrectResult(t *testing.T) {
 	for _, example := range testExamples {
 		for _, variant := range getAllInterpreterVariantsForTests() {
-			vm := tosca.GetInterpreter(variant)
+			vm, err := tosca.NewInterpreter(variant)
+			if err != nil {
+				t.Fatalf("failed to load %s interpreter: %v", variant, err)
+			}
 			for i := 0; i < 10; i++ {
 				t.Run(fmt.Sprintf("%s-%s-%d", example.Name, variant, i), func(t *testing.T) {
 					want := example.RunReference(i)
@@ -61,9 +64,15 @@ func TestExamples_ComputesCorrectResult(t *testing.T) {
 func TestExamples_ComputesCorrectGasPrice(t *testing.T) {
 	for _, example := range testExamples {
 		for _, revision := range revisions {
-			reference := tosca.GetInterpreter("geth")
+			reference, err := tosca.NewInterpreter("geth")
+			if err != nil {
+				t.Fatalf("failed to load geth interpreter: %v", err)
+			}
 			for _, variant := range getAllInterpreterVariantsForTests() {
-				vm := tosca.GetInterpreter(variant)
+				vm, err := tosca.NewInterpreter(variant)
+				if err != nil {
+					t.Fatalf("failed to load %s interpreter: %v", variant, err)
+				}
 				for i := 0; i < 10; i++ {
 					t.Run(fmt.Sprintf("%s-%s-%s-%d", example.Name, revision, variant, i), func(t *testing.T) {
 						want, err := example.RunOn(reference, i)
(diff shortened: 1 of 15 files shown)


facuMH (dismissed):

go/tosca/interpreter_registry.go
```diff
@@ -30,14 +30,47 @@ import (
 // GetInterpreter performs a lookup for the given name (case-insensitive) in
 // the registry. The result is nil if no interpreter was registered under the
 // given name.
+//
+// Deprecated: Use NewInterpreter instead.
```
LuisPH3, line 34: I could not find any use, can we remove it?

go/tosca/interpreter_registry.go
```diff
@@ -30,14 +30,47 @@ import (
...
+	if len(config) > 1 {
+		return nil, fmt.Errorf("invalid configuration: too many arguments")
+	}
+	factory := GetInterpreterFactory(name)
+	if factory == nil {
+		return nil, fmt.Errorf("interpreter not found: %s", name)
+	}
+	c := any(nil)
+	if len(config) > 0 {
+		c = config[0]
```
LuisPH3, line 58: Since it is an any value. It would be possible to pass all the variadic arguments as a slice. 
Nevertheless, this feature is never used. Dead code?

LuisPH3 (approved):
I really like this change. The registry may be unavoidable, but at least now there are no silent objects created in the background.

LuisPH3:
@LorenzSchueler FYI, your code may have to migrate instance initialization because of this change.

go/tosca/interpreter_registry.go
HerbertJordan: Aida is using it.

go/tosca/interpreter_registry.go
HerbertJordan: As said in the description, this feature introduces the option to pass configuration parameters like cache sizes to the interpreters. Not used know, but when we need it, we do not have to break the existing API.

go/integration_test/interpreter/vm_test.go
```diff
@@ -41,7 +41,10 @@ var (
 func TestExamples_ComputesCorrectResult(t *testing.T) {
 	for _, example := range testExamples {
 		for _, variant := range getAllInterpreterVariantsForTests() {
-			vm := tosca.GetInterpreter(variant)
+			vm, err := tosca.NewInterpreter(variant)
+			if err != nil {
+				t.Fatalf("failed to load %s interpreter: %v", variant, err)
```
facuMH, line 46: nit: this  error message is used in multiple files, it could be oursourced to `makeLoadError(variant, err)`

go/tosca/interpreter_registry.go
```diff
@@ -48,22 +81,43 @@ func GetAllRegisteredInterpreters() map[string]Interpreter {
...
+	})
+	if err != nil {
+		panic(err)
+	}
+}
+
+// RegisterInterpreterFactory registers a new Interpreter implementation
+// to be exported for general use in the binary. The name is not case-sensitive,
+// and a panic is triggered if a factory was bound to the same name before, or
+// the factory is nil. This function is mainly intended to be used by package
```
facuMH, line 98: if it is meant as package initialization, should it be public ? should it not be in `func init()` ?

facuMH (approved):

go/tosca/interpreter_registry.go
HerbertJordan: This is meant to be a function to be **used** by package initialization code. It is not initializing this package.

go/integration_test/interpreter/vm_test.go
HerbertJordan: This is an error message of a unit test, not an error. I think it is fine to be explicit there and not to obfuscate error messages through level of indirection.

LuisPH3 (approved):

PR #767, opened 2024-09-13
Implement evmc interfaces and add ffi functions

LorenzSchueler:
- add ffi functions which get called by Go
- implement `EvmcVm` and `SteppableEvmcVm`


LorenzSchueler:
@LuisPH3 Jenkins failed again. Can you please have a look at this?

LuisPH3:
> and because `evmc_declare` seems to be no longer maintained.
- Can you provide a link where I can see why is it no longer maintained?

rust/src/ffi/evmc_vm.rs
```diff
@@ -0,0 +1,211 @@
...
+    },
+    EvmcContainer, EvmcVm, ExecutionContext, ExecutionMessage, ExecutionResult, SetOptionError,
+};
+
+use crate::evmc::EvmRs;
+
+static EVM_RS_NAME: &str = "evmrs\0";
+static EVM_RS_VERSION: &str = "0.1.0\0";
+
+pub const EVMC_CAPABILITY: evmc_capabilities = evmc_capabilities::EVMC_CAPABILITY_EVM1;
```
LuisPH3, line 20: A comment explaining why is EVM1 would be nice to have.

rust/src/ffi/evmc_vm.rs
```diff
@@ -0,0 +1,211 @@
...
+        // code is not null and code size > 0
+        unsafe { slice::from_raw_parts(code, code_size) }
+    };
+
+    let container = unsafe {
+        // Acquire ownership from EVMC.
+        EvmcContainer::<EvmRs>::from_ffi_pointer(instance)
+    };
+
+    let result = panic::catch_unwind(|| {
```
LuisPH3, line 138: I have been wondering about a similar mechanism for go as well:
We could spawn a goroutine and check for panics, but currently it is not done. 
I am not sure about the reasons behind, it may have been discarded.

It would be interesting to study what are the implications of doing this, is there any performance penalty?

rust/src/ffi/evmc_vm.rs
```diff
@@ -0,0 +1,211 @@
...
+    code: *const u8,
+    code_size: usize,
+) -> evmc_result {
+    if instance.is_null()
+        || (host.is_null() && EVMC_CAPABILITY != evmc_capabilities::EVMC_CAPABILITY_PRECOMPILES)
+        || message.is_null()
+        || (code.is_null() && code_size != 0)
+    {
+        // These are irrecoverable errors that violate the EVMC spec.
+        std::process::abort();
```
LuisPH3, line 115: I am not sure about this.
On the one hand, I understand the validity of rejecting ill-formed context. On the other hand, I see two problems with this: 

- The conditions checked in this have different criticality:
   - this code is indeed odd, but see second paragraph about abort. 
   - capability is an explicit check: that should be done here
   - null instance could be the sign of an ill-formed program. See second paragraph about this.
   - null host could be ignored until callbacks are needed, but it may be better to check it here nevertheless. Execution error could be returned. 
   - I am not sure about message. I am not sure what does it contain, and a default value could mask the error completely and return a successful execution. 

- We have no guarantee at all about the correctness or nature of the environment calling this library. In production would be pretty bad if this crashes, and we have no means to test if every input will be always valid.  In other words: murphy's law, if you do not write an abort here, it cannot abort here. 
    - Would it be fundamentally wrong to return an error? EVMC_FAILURE or EVMC_INTERNAL_ERROR ? 

I see that the abort comes from the original implementation: https://docs.rs/evmc-declare/latest/src/evmc_declare/lib.rs.html#350
Nevertheless, It may be sensible to avoid killing the process.

rust/src/ffi/steppable_evmc_vm.rs
```diff
@@ -0,0 +1,181 @@
...
+    let container = SteppableEvmcContainer::<EvmRs>::new(new_instance);
+
+    // Release ownership to EVMC.
+    unsafe { SteppableEvmcContainer::into_ffi_pointer(container) }
+}
+
+extern "C" fn __evmc_steppable_destroy(instance: *mut evmc_vm_steppable) {
+    if instance.is_null() {
+        // This is an irrecoverable error that violates the EVMC spec.
+        std::process::abort();
```
LuisPH3, line 33: Not sure about this neither, I see where it comes from, but I guess that at this point it could be ignored (just don't drop it if null)

rust/src/ffi/steppable_evmc_vm.rs
```diff
@@ -0,0 +1,181 @@
...
+    last_call_result_data_size: usize,
+    steps: i32,
+) -> evmc_step_result {
+    if instance.is_null()
+        || (host.is_null() && EVMC_CAPABILITY != evmc_capabilities::EVMC_CAPABILITY_PRECOMPILES)
+        || message.is_null()
+        || (code.is_null() && code_size > 0)
+        || (stack.is_null() && stack_size > 0)
+        || (memory.is_null() && memory_size > 0)
+        || (last_call_result_data.is_null() && last_call_result_data_size > 0)
```
LuisPH3, line 67: see comment for __evmc_execute

rust/src/interpreter/run_result.rs
```diff
@@ -66,3 +66,14 @@ impl<'a> From<RunResult<'a>> for StepResult {
         )
     }
 }
+
+impl<'a> From<RunResult<'a>> for ExecutionResult {
```
LuisPH3, line 70: I was reviewing this code and I had to expand the file. 
```rust
impl<'a> From<StatusCode> for RunResult<'a> {
    fn from(status_code: StatusCode) -> Self {
        let step_status_code = match status_code {
            StatusCode::EVMC_SUCCESS => StepStatusCode::EVMC_STEP_RUNNING,
            StatusCode::EVMC_REVERT => StepStatusCode::EVMC_STEP_REVERTED,
            _ => StepStatusCode::EVMC_STEP_FAILED,
        };
        Self {
            step_status_code,
            status_code,
            revision: Revision::EVMC_FRONTIER,
            code_state: CodeState::new(&[], 0),
            gas_left: 0,
            gas_refund: 0,
            output: None,
            stack: Stack::new(Vec::new()),
            memory: Memory::new(Vec::new()),
            last_call_return_data: None,
        }
    }
}
```

This is fishy, we turn status code into EVMC_FRONTIER, but I don't see why we need revision in the result. Now we are casting `ExecutionResult` into `ExecutionResult`. do we even need that type?

LorenzSchueler: In an upcomming refactor PR I removed the RunResult completely but did not want to make this PR even larger. So maybe ignore this for now?

rust/src/ffi/steppable_evmc_vm.rs
LorenzSchueler: Here I would not have an issue with ignoring it. Will change this.

rust/src/interpreter/run_state.rs
```diff
@@ -24,19 +24,17 @@ impl<'a> RunState<'a> {
         revision: Revision,
         message: &'a ExecutionMessage,
         context: &'a mut ExecutionContext<'a>,
-        step_status_code: StepStatusCode,
         code: &'a [u8],
-        gas_refund: i64,
     ) -> Self {
         Self {
-            step_status_code,
+            step_status_code: StepStatusCode::EVMC_STEP_RUNNING,
```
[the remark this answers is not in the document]
LorenzSchueler: In the refactoring of the last PR I simply overlooked that `step_status_code` and `gas_refund` are not available in the non steppable interface.

rust/src/ffi/evmc_vm.rs
LorenzSchueler: Sure

rust/src/ffi/evmc_vm.rs
LorenzSchueler: That is certainly true. I don't know the performance implications and would have to read up on this. I kept it like in the macro but I was wondering why not everything in this function in `catch_unwind`. After all there is (unsafe) FFI code outside which if implemented incorrectly would me most likely to panic.

rust/src/ffi/evmc_vm.rs
LorenzSchueler: I understand that you would like to avoid aborting. If there are sensible defaults for ill formed values that could be used instead, I would be fine with that. And if we can return an error that would be even better but here I'm unsure about the semantics. However I would like to avoid deferring check until first use because then we have those checks all over the place and not (like currently) in a single place.

LorenzSchueler: Unless @HerbertJordan has a definite opinion on what to do here we could also just open an issue to not forget about it and merge it as is. This way all the other stuff can get merged more quickly and changing this later on does not influence the other parts of the code anyway.

LuisPH3: If you are unsure about the semantics, we leave as it is. After all, this is a copy of the original macro expansion, and not a redo.  
The one option that we could explore in the future, is what happens if we return an error here instead of abort.

rust/src/ffi/evmc_vm.rs
LuisPH3: You can take a note about this. And you could investigate this later on.

rust/src/interpreter/run_result.rs
LuisPH3: looking forward.

LuisPH3 (approved):

PR #769, opened 2024-09-13
Update go-ethereum-sonic dependency

HerbertJordan:
As a preparation for the integration of Tosca into Sonic, the Tosca extensions in the go-ethereum-sonic repository got revised. This PR updates the dependency to to the latest version and applies necessary interface adaptations.


HerbertJordan:
Replaced by #770

PR #772, opened 2024-09-16
LFVM: unify error status

facuMH:
Different error status were handled similarly but inconsistently. Functions were also inconsistent when propagating or generating errors regarding when they would be responsible for reporting `statusError`, because of this there were also a lot of code smells of the type 
```
if err != nil {
    return
}
```
Hiding error handling which seems very error prone.

This PR:
- unifies the 3 erroneous status (`statusOutOfGas`, `statusInvalidInstruction` and `statusError`) in a single `statusError`
- adds documentation for functions whose errors should be handled by caller
- consistently handles error propagating and handling.

A CT run was made to corroborate that correctness of known cases has not been affected.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -99,12 +99,11 @@ func opPush(c *context, n int) {
 }
 
 func opPush0(c *context) {
-	if c.isAtLeast(tosca.R12_Shanghai) {
-		z := c.stack.pushUndefined()
-		z[3], z[2], z[1], z[0] = 0, 0, 0, 0
-	} else {
-		c.status = statusInvalidInstruction
+	if !c.isAtLeast(tosca.R12_Shanghai) {
+		c.signalError()
 	}
+	z := c.stack.pushUndefined()
+	z[3], z[2], z[1], z[0] = 0, 0, 0, 0
 }
 
 func opPush1(c *context) {
@@ -164,10 +163,10 @@ func opMstore(c *context) {
 
 	offset, overflow := addr.Uint64WithOverflow()
 	if overflow {
-		c.status = statusError
+		c.signalError()
 		return
 	}
-	if err := c.memory.SetWord(offset, value, c); err != nil {
+	if c.memory.SetWord(offset, value, c) != nil {
 		c.signalError()
 	}
 }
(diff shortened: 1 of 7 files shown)


go/interpreter/lfvm/interpreter.go
```diff
@@ -604,15 +603,13 @@ func steps(c *context, oneStepOnly bool) {
 	}
 }
 
+// satisfiesStackRequirements checks if the current stack size satisfies the
```
LuisPH3, line 606: this sentence does not say much, does it? 
```suggestion
// satisfiesStackRequirements checks that the opCode will not make an out of bounds access with the current stack size. 
```

go/interpreter/lfvm/interpreter.go
```diff
@@ -604,15 +603,13 @@ func steps(c *context, oneStepOnly bool) {
...
+// stack requirements of the given OpCode.
+// caller should handle false return as an error.
 func satisfiesStackRequirements(c *context, op OpCode) bool {
```
LuisPH3, line 609: Now that the function does not modify the context, we could pass only the stack size as an argument instead of the full mutable context

LuisPH3 (approved):
The PR changes the interpreter so that errors are only set at the instruction level. 
it cleans many error variables declared and dropped in the same line.

LuisPH3 (dismissed):

go/interpreter/lfvm/interpreter.go
```diff
@@ -604,15 +603,12 @@ func steps(c *context, oneStepOnly bool) {
 	}
 }
 
-func satisfiesStackRequirements(c *context, op OpCode) bool {
+// satisfiesStackRequirements checks that the opCode will not make an out of
+// bounds access with the current stack size.
+// Caller should handle false return as an error.
```
simonlechner, line 608: Why don't you directly return an error?

go/interpreter/lfvm/instructions.go
```diff
@@ -862,6 +872,10 @@ func opExtcodehash(c *context) {
 	}
 }
 
+// checkInitCodeSize checks the size of the init code.
+// if size is greater than MaxInitCodeSize, gas is consumed and returns false.
+// if not enough gas is available, returns false.
+// caller should handle false return value by stopping the execution.
```
simonlechner, line 878: Why don't you directly return an error?

simonlechner (commented):
Maybe I overseeing other calls of these functions but I would rather have a function directly returning an error than a comment that says it should be handled as an error.

go/interpreter/lfvm/interpreter.go
facuMH: because I wanted to keep changes to a minimum. Also with the function name starting with `satisfies` seems reasonable it returns true/false.

go/interpreter/lfvm/instructions.go
facuMH: this is done in #786

HerbertJordan (approved):

Issue #766, opened 2024-09-13
Cleanup Tosca integration into Sonic


HerbertJordan:
The integration of Tosca into Sonic is based on an early prototype for intercepting the interpreter. This should be reviewed and improved as necessary.

Issue #687, opened 2024-08-26
Isolate and/or eliminate experimental features


HerbertJordan:
To simplify the scope of the released code base, there should be a clean separation between release-ready and experimental features in the code base. Experimental features may include super instructions, profiling modes or execution modes collecting statistics.

Issue #676, opened 2024-08-26
Define LFVM configuration to be released


HerbertJordan:
The LFVM implementation offers a range of optional features that can be configured when creating an instance, before being registered in the interpreter registry. The available options include the caching of code conversion results, the caching of hash results, the enabling of special features like shadow, profiling, or logging modes, etc...

In order to focus the development toward the release, a target-configuration is to be defined.

PR #773, opened 2024-09-16
Provide public LFVM interpreter creator function

HerbertJordan:
This PR performs a clean-up of the LFVM instantiation code. It does so by
- providing a new `NewInterpreter(Config)` function where `Config` is supposed to be extended in the future with LFVM specific properties (e.g. cache sizes); the implementation of these properties is beyond the scope of this PR, but since this interface will be accessed by down-stream components (e.g. the Sonic client), adding the `Config` only once needed will be a breaking change;
- the package internal pre-existing `Config` was made private
- the old factory function `NewVm` was made private

This is part of #676, #687, #774, and #766

### DIFF
--- a/go/interpreter/lfvm/lfvm.go
+++ b/go/interpreter/lfvm/lfvm.go
@@ -16,10 +16,26 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 )
 
+// Config provides a set of user-definable options for the LFVM interpreter.
+type Config struct {
+	// TODO: add options for code cache size and other configuration options
+}
+
+// NewInterpreter creates a new LFVM interpreter instance with the official
+// configuration for production purposes.
+func NewInterpreter(Config) (*lfvm, error) {
+	return newVm(config{
+		ConversionConfig: ConversionConfig{
+			WithSuperInstructions: false,
+		},
+		WithShaCache: true,
+	})
+}
+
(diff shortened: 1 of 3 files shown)


go/interpreter/lfvm/lfvm.go
```diff
@@ -16,10 +16,26 @@ import (
...
 )
 
+// Config provides a set of user-definable options for the LFVM interpreter.
+type Config struct {
+	// TODO: add options for code cache size and other configuration options
+}
+
+// NewInterpreter creates a new LFVM interpreter instance with the official
+// configuration for production purposes.
+func NewInterpreter(Config) (*lfvm, error) {
```
facuMH, line 26: since this is not an interface and it will be called in a context where other interpreters will also be instantiated, what do you think of `NewLfvmInstance`  instead ?

facuMH (approved):

go/interpreter/lfvm/lfvm.go
HerbertJordan: It is in the package `lfvm` so it is most likely called as `lfvm.NewInterpreter(lfvm.Config{...})`. So the term `lfvm` already shows up twice. An example can be found [here](https://github.com/Fantom-foundation/Sonic/blob/8e5ede0624be3df10324f235e9058b6eef63d2ee/opera/rules.go#L32).

LuisPH3 (approved):

PR #775, opened 2024-09-17
LFVM: change return type of useGas to error

facuMH:
`useGas` used to return `bool`, to inform whether it was successful or not, in Go this should be done using `error`, for error propagation. 
This PR does just that, changes the return type from `bool` to `error` and checks whether this error is nil or not in all the call places. 
This PR should be carefully rebased with #772 since they are both about errors, but this change seemed big enough, conceptually, to be its own PR.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -889,7 +889,7 @@ func checkInitCodeSize(c *context, size *uint256.Int) error {
 	if !size.IsUint64() || size.Uint64() > MaxInitCodeSize {
 		c.useGas(c.gas)
 		c.signalError()
-		return errOutOfGas
+		return errInitCodeTooLarge
 	}
 	return c.useGas(tosca.Gas(InitCodeWordGas * tosca.SizeInWords(size.Uint64())))
 }
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -211,7 +211,7 @@ func opMcopy(c *context) {
 
 	size := sizeU256.Uint64()
 	price := tosca.Gas(3 * tosca.SizeInWords(size))
-	if !c.useGas(price) {
+	if err := c.useGas(price); err != nil {
 		c.signalError()
 		return
 	}
@@ -271,7 +271,7 @@ func opSstore(c *context) {
 	storageStatus := c.context.SetStorage(c.params.Recipient, key, value)
 
 	cost += getDynamicCostsForSstore(c.params.Revision, storageStatus)
-	if !c.useGas(cost) {
+	if err := c.useGas(cost); err != nil {
 		c.signalError()
 		return
 	}
@@ -290,7 +290,7 @@ func opSload(c *context) {
 		if c.context.AccessStorage(addr, slot) == tosca.ColdAccess {
 			costs = 2100
 		}
-		if !c.useGas(costs) {
+		if err := c.useGas(costs); err != nil {
 			c.signalError()
 			return
 		}
@@ -394,7 +394,7 @@ func opCallDataCopy(c *context) {
 	// Charge for the copy costs
 	words := tosca.SizeInWords(length64)
 	price := tosca.Gas(3 * words)
-	if !c.useGas(price) {
+	if err := c.useGas(price); err != nil {
 		c.signalError()
 		return
 	}
@@ -596,7 +596,7 @@ func opSMod(c *context) {
 
 func opExp(c *context) {
 	base, exponent := c.stack.pop(), c.stack.peek()
-	if !c.useGas(tosca.Gas(50 * exponent.ByteLen())) {
+	if err := c.useGas(tosca.Gas(50 * exponent.ByteLen())); err != nil {
 		c.signalError()
 		return
 	}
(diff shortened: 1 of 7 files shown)


go/interpreter/lfvm/instructions.go
```diff
@@ -862,27 +862,26 @@ func opExtcodehash(c *context) {
...
 	}
 	if !size.IsUint64() || size.Uint64() > MaxInitCodeSize {
 		c.useGas(c.gas)
 		c.signalError()
-		return false
-	}
-	if !c.useGas(tosca.Gas(InitCodeWordGas * tosca.SizeInWords(size.Uint64()))) {
-		c.status = statusOutOfGas
-		return false
+		return errOutOfGas
```
simonlechner, line 882: This error is due to the size being greater than the MaxInitCodeSize, not due to out of gas.

simonlechner (dismissed):

go/interpreter/lfvm/instructions.go
```diff
@@ -862,27 +862,26 @@ func opExtcodehash(c *context) {
 	}
 }
 
-func checkInitCodeSize(c *context, size *uint256.Int) bool {
+// checkInitCodeSize checks the size of the init code.
+// An error is returned if size is greater than MaxInitCodeSize, or
+// if not enough gas is available.
+// Caller should handle error return and stop the execution.
+func checkInitCodeSize(c *context, size *uint256.Int) error {
```
HerbertJordan, line 869: For a follow-up: Since the function is charging gas it should not just be called "check".

This function actually does a lot of stuff that should be divided up:
- check for maximum length depending on the revision
- if the size is larger than the maximum limit it consumes all the remaining gas (which it should probably not do; using all remaining gas in an error case should be handled in some central place)
- it charges gas for the init code size

This function looks to me like a prime candidate for being inlined into its call-side which after #781 should only be a single place. Wdyt?

go/interpreter/lfvm/interpreter.go
```diff
@@ -247,7 +247,7 @@ func steps(c *context, oneStepOnly bool) {
 		}
 
 		// Consume static gas price for instruction before execution
-		if !c.useGas(staticGasPrices.get(op)) {
+		if c.useGas(staticGasPrices.get(op)) != nil {
```
HerbertJordan, line 250: Can you run a benchmark like fib before and after this change? Just to be safe that replacing a boolean return value with a interface return type is not adding significant overhead here by losing some compiler optimizations.

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -53,18 +53,22 @@ func TestContext_useGas_HandlesTerminationIfOutOfGas(t *testing.T) {
...
 			// Check that the result of UseGas indicates whether there was
 			// enough gas.
-			want := test.required >= 0 && test.available >= test.required
-			if want != success {
-				t.Errorf("expected UseGas to return %v, got %v", want, success)
+			want := error(nil)
+			// TODO: remove status check once #772 is merged
+			if test.required < 0 || test.available < test.required {
+				want = errOutOfGas
+			}
```
HerbertJordan, line 64: nit: you do not have to deal with error types if you define
```
success := err != nil
```
and keep using the old "want != success" check.

HerbertJordan (changes requested):
Looks good, but it would be interesting to see whether the static-gas consumption in the interpreter loop is getting significantly slowed down by this change.

go/interpreter/lfvm/instructions.go
facuMH: gas usage can definitely be inlined in genericCreate, but I like having the `MaxInitCodeSize` check  here. As for consuming all gas, I think it is an exceptional handling. But with proper error propagation it should easily be handled in `genericCreate` instead of here.  But yes, I like the little refactor

go/interpreter/lfvm/interpreter.go
facuMH:
```
         │  old.benc   │              new.benc              │
Fib10-16   88.84µ ± 1%   88.17µ ± 2%  -0.76% (p=0.019 n=50)
```
I think we are safe.

simonlechner (dismissed):

LuisPH3 (dismissed):
I like the overal idea, IMO seting  `c.status = statusOutOfGas` in UseGas AND return error is till redundant (and somethimes overrided) 
I am working on removing this [here](https://github.com/Fantom-foundation/Tosca/compare/main...luis/lfvm_interpreter_test)
But I can use this change.

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -53,18 +53,19 @@ func TestContext_useGas_HandlesTerminationIfOutOfGas(t *testing.T) {
...
-			success := ctx.useGas(test.required)
+			err := ctx.useGas(test.required)
 
 			// Check that the result of UseGas indicates whether there was
 			// enough gas.
 			want := test.required >= 0 && test.available >= test.required
+			success := err == nil
 			if want != success {
-				t.Errorf("expected UseGas to return %v, got %v", want, success)
+				t.Errorf("expected UseGas to return %v, got %v", want, err)
```
HerbertJordan, line 63:
```suggestion
				t.Errorf("expected UseGas to return %v, got %v", want, success)
```

HerbertJordan (approved):

LuisPH3 (dismissed):
As discussed, `if c.useGas(price) != nil {` is a no-pattern.
I have a fix in the making, If my PR does not make we will have to clean these  ifs.

HerbertJordan (approved):

simonlechner (approved):

PR #777, opened 2024-09-17
Enable more lints & refactor

LorenzSchueler:
- enable additional lints and fix them
- depend on `Fantom-foundation/evmc`
- rename `CodeState` -> `CodeReader`
- rename `RunState` -> `Interpreter`
- make `run` and `create/call*` methods of `Interpreter`
- remove `RunResult`
- move `memory.rs`, `stack.rs`, `code_reader.rs` into `types` module


rust/Cargo.toml
```diff
@@ -5,8 +5,45 @@ edition = "2021"
...
 
 [lib]
 crate-type = ["cdylib"]
+
+[lints.rust]
+macro_use_extern_crate = "warn"
+unused_crate_dependencies = "warn"
+unused_unsafe = "warn"
+
+[lints.clippy]
```
LuisPH3, line 19: Can you explain to us why we need to list the lints ?

Are there reduced in criticality? error to warning. or they are they turned on for the first time? 

I think this list is quite difficult to maintain, and I would rather prefer to have curated groups or defaults.
For example, can we add full grops from these categories?  https://doc.rust-lang.org/clippy/lints.html

The maintenance of this list requires an of investment, and it may not be so easy to find time for this in the future. For this reason this list may end up unmaintained, out of date, and inducing confusion.

rust/src/evmc.rs
```diff
@@ -34,10 +34,10 @@ impl EvmcVm for EvmRs {
...
             process::abort();
         };
-        let run_state = RunState::new(revision, message, context, code);
-        interpreter::run(run_state)
-            .unwrap_or_else(Into::into)
-            .into()
+        Interpreter::new(revision, message, context, code)
+            .run()
+            .map(Into::into)
+            .unwrap_or_else(|status_code| ExecutionResult::new(status_code, 0, 0, None))
```
LuisPH3, line 40: I got quite confused the other day by this statement. But I get it now. 
The problem that I see here is that `Run` returns a type `Result<Self, StatusCode>` and `StatusCode` is odd type to be the error type since it can be a successful value.

This is not very rustastic, since the purpose of Result is to have Success/Error as disjoint types, and status_code is a union of both cases. 

The problem comes from evmc, the interpretation they do of all those types comes from a naive translation of the C types. I guess this is a necessary evil. But we may want to consider in the future having disjoint types inside the rust vm

rust/src/types/amount.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::multiple_inherent_impl)]
```
LuisPH3, line 1: We already talked about this, the order of declaring the structs. 
Now that clippy is on my side, I am going to ask you to place together member implementations. 

I am not a fan of writing extra lines to suppress extra tooling.

rust/src/ffi/steppable_evmc_vm.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::undocumented_unsafe_blocks)]
```
LuisPH3, line 1: we do have document lines inside of unsafe code in this file. please add the missing ones and remove the exception.

rust/src/ffi/evmc_vm.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::undocumented_unsafe_blocks)]
```
LuisPH3, line 1: idem. document and remove.

rust/src/interpreter.rs
LuisPH3, line 1: Where does this file come from?  i thought we already reviewed this code, why is it a new file on this PR?

rust/src/types/code_reader.rs
```diff
@@ -5,13 +5,13 @@ use evmc_vm::StatusCode;
 use crate::types::{code_byte_type, u256, CodeByteType, Opcode};
 
 #[derive(Debug)]
-pub struct CodeState<'a> {
+pub struct CodeReader<'a> {
```
LuisPH3, line 8: :+1:

LuisPH3 (changes requested):
please see comments bellow.
Additionally, check interpreter.rs, this change is too large to be merged alongside with the lints.

rust/Cargo.toml
LorenzSchueler: Those are all enabled for the first time. Most of them are in the `pendantic` lints group.

docs:
The clippy::pedantic group makes Clippy even more pedantic. You can enable the whole group with #![warn(clippy::pedantic)] in the lib.rs/main.rs of your crate. This lint group is for Clippy power users that want an in depth check of their code.

Note: Instead of enabling the whole group (like Clippy itself does), you may want to cherry-pick lints out of the pedantic group.

If you enable this group, expect to also use #[allow] attributes generously throughout your code. Lints in this group are designed to be pedantic and false positives sometimes are intentional in order to prevent false negatives.

If at any point we no longer want to use them we can just remove them and everthing keeps working. What should we do?

rust/src/interpreter.rs
LorenzSchueler: `interpreter/mod.rs`

rust/Cargo.toml
LuisPH3: ok, let's keep it then

LuisPH3 (approved):

PR #781, opened 2024-09-18
LFVM: unify create and create2

facuMH:
`CREATE` and `CREATE2` are two extremely similar operations with a little variations for the latter.
This PR unifies their implementation making maintainability a little bit easier for the future (and also bugfixing if ever needed) and reducing code duplication.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -905,10 +905,10 @@ func genericCreate(c *context, kind tosca.CallKind) {
 		value  = c.stack.pop()
 		offset = c.stack.pop()
 		size   = c.stack.pop()
-		salt   = uint256.NewInt(0) // salt us zero for Create
+		salt   = tosca.Hash{}
 	)
 	if kind == tosca.Create2 {
-		salt = c.stack.pop() // pop salt value for Create2
+		salt = c.stack.pop().Bytes32() // pop salt value for Create2
 	}
 
 	if checkSizeOffsetUint64Overflow(offset, size) != nil {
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/instructions.go
```diff
@@ -982,10 +924,13 @@ func opCreate2(c *context) {
 		return
 	}
 
-	// Charge for the code size
-	words := tosca.SizeInWords(size.Uint64())
-	if !c.useGas(tosca.Gas(6 * words)) {
-		return
+	if kind == tosca.Create2 {
+		// Charge for the code size
```
HerbertJordan, line 928: nit: 
```suggestion
		// Charge for hashing the init code to compute the target address.
```

go/interpreter/lfvm/instructions.go
```diff
@@ -967,9 +905,13 @@ func opCreate2(c *context) {
 		value  = c.stack.pop()
 		offset = c.stack.pop()
 		size   = c.stack.pop()
-		salt   = c.stack.pop()
+		salt   = uint256.NewInt(0) // salt us zero for Create
```
HerbertJordan, line 908: nit: declaring the salt as a 
```suggestion
		salt   tosca.Hash
```
and assigning the salt in line 911 below as
```
salt = c.stack.pop().Bytes32()
```
removes a memory allocation.

HerbertJordan (dismissed):
Looks good, thanks!

simonlechner (approved):

go/interpreter/lfvm/instructions.go
```diff
@@ -886,78 +886,16 @@ func checkInitCodeSize(c *context, size *uint256.Int) bool {
 }
 
 func opCreate(c *context) {
```
LuisPH3, line 888: nit, just like in push, swap, log. The call to genericCreate could be done directly from the interpreter switch.

LuisPH3 (approved):

go/interpreter/lfvm/instructions.go
facuMH: unless it would bring a big performance performance, I wouldn't do that. maybe in another PR we add a benchmarka and  do the change you suggest.

PR #782, opened 2024-09-18
LFVM: `genericCall` memory expansion

facuMH:
We want to move towards the memory methods being secure (aka only modifying the memory once it has been expanded and charged for). But there were the exception cases in the call operations, where we needed to know how much the memory expansion would cost before using it.
This PR moves up the memory expansions within the `genericCall` function keeping a reference for the return slice, hence removing the need of a custom read/write of memory that would not expand it. 

CT run for the calls was executed and passed.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -1044,15 +1044,11 @@ func checkSizeOffsetUint64Overflow(offset, size *uint256.Int) error {
 	return nil
 }
 
-func neededMemorySize(c *context, offset, size *uint256.Int) (uint64, error) {
+func checkAndReturnUint64(offset, size *uint256.Int) (uint64, uint64, error) {
 	if err := checkSizeOffsetUint64Overflow(offset, size); err != nil {
-		c.signalError()
-		return 0, err
+		return 0, 0, err
 	}
-	if size.IsZero() {
-		return 0, nil
-	}
-	return offset.Uint64() + size.Uint64(), nil
+	return offset.Uint64(), size.Uint64(), nil
 }
 
 func getAccessCost(accessStatus tosca.AccessStatus) tosca.Gas {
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/instructions.go
```diff
@@ -1098,15 +1098,11 @@ func checkSizeOffsetUint64Overflow(offset, size *uint256.Int) error {
 	return nil
 }
 
-func neededMemorySize(c *context, offset, size *uint256.Int) (uint64, error) {
+func checkAndReturnUint64(offset, size *uint256.Int) (uint64, uint64, error) {
```
LuisPH3, line 1101: Check what? Returns 2 ints.

LuisPH3 (approved):

go/interpreter/lfvm/instructions.go
facuMH: functioned removed. used another existing one for the checks.

simonlechner (approved):

go/interpreter/lfvm/instructions.go
```diff
@@ -1081,22 +1070,25 @@ func genericCall(c *context, kind tosca.CallKind) {
...
+	args, err := c.memory.GetSliceWithCapacityAndGas(inOffset.Uint64(), inSize.Uint64(), c)
 	if err != nil {
+		c.signalError()
 		return
 	}
-
-	needed_memory_size := arg_memory_size
-	if ret_memory_size > arg_memory_size {
-		needed_memory_size = ret_memory_size
+	output, err := c.memory.GetSliceWithCapacityAndGas(retOffset.Uint64(), retSize.Uint64(), c)
```
HerbertJordan, line 1085: I assume `GetSliceWithCapacityAndGas` will be renamed by some other PR into something simpler like `getSlice`.

HerbertJordan (approved):
Nice improvement, thanks!

go/interpreter/lfvm/instructions.go
facuMH: exactly

PR #783, opened 2024-09-18
Add new linter to detect unassigned error returns

LuisPH3:
This PR makes more difficult to ignore error returns. 
The problem why this was not working earlier is explained [here](https://www.dolthub.com/blog/2024-07-24-static-analysis/)

Found cases have been fixed with usual methods. 
Only the close call, between an error generation and its use, seems odd. This error in particular is discussed in the [go-PR](https://github.com/golang/go/issues/20148#issuecomment-2148205291) that attempted to enable this check in vet but was rejected.

### DIFF
--- a/go/interpreter/lfvm/interpreter_test.go
+++ b/go/interpreter/lfvm/interpreter_test.go
@@ -371,7 +371,7 @@ func TestRunWithLogging(t *testing.T) {
 		runner: loggingRunner{},
 	}, params, code)
 	// read the output
-	w.Close()
+	_ = w.Close() // ignore error in test
 	out, _ := io.ReadAll(r)
 	os.Stdout = old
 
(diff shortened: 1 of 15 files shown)


go/interpreter/lfvm/interpreter_test.go
```diff
@@ -375,7 +375,7 @@ func TestRunWithLogging(t *testing.T) {
 		runner: loggingRunner{},
 	}, params, code)
 	// read the output
-	w.Close()
+	_ = w.Close() // ignore error in test
```
HerbertJordan, line 378: Why not do the usual in tests - check for an error and fail if there is one? 
There is no need for 100% code coverage in tests.

HerbertJordan (dismissed):
Thanks for adding this checks, looks good!

go/interpreter/lfvm/interpreter_test.go
LuisPH3: This is a funny scenario:
the close happens between `_, err := run(interpreterConfig{` and `if err != nil` (for the run err) 

This is the precise function call that made the decision not to enable this check inside of Vet. 

Because `err:= run` `err2:= close`  is pretty odd, I decided to ignore it for now. Real use cases should add a `defer` and deal with releasing resources properly.

facuMH (approved):

PR #786, opened 2024-09-19
LFVM: refactor max init code size check

facuMH:
the previous `checkInitCodeSize` also did modifications to gas, which was not documented nor in the function name. Since we trying to maintain all `useGas` calls in `opX` functions, this is moved to `genericCreate` along with the revision check, simplifying greatly this function.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -940,6 +940,8 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 	return nil
 }
 
+// checkInitCodeSize checks the size of the init code.
+// An error is returned if size is greater than MaxInitCodeSize.
 func checkMaxInitCodeSize(size uint64) error {
 	const (
 		maxCodeSize     = 24576           // Maximum bytecode to permit for a contract
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -841,23 +841,6 @@ func opExtcodehash(c *context) error {
 	return nil
 }
 
-// checkInitCodeSize checks the size of the init code.
-func checkInitCodeSize(c *context, size *uint256.Int) error {
-	const (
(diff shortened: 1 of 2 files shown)


go/interpreter/lfvm/instructions.go
```diff
@@ -924,19 +903,30 @@ func genericCreate(c *context, kind tosca.CallKind) {
...
 		c.signalError()
 		return
 	}
 
-	if err := checkInitCodeSize(c, size); err != nil {
-		c.signalError()
-		return
+	if c.isAtLeast(tosca.R12_Shanghai) {
+		if checkMaxInitCodeSize(sizeU64) != nil {
+			c.useGas(c.gas)
```
LuisPH3, line 914: Use gas is not a responsibility here. 
Any out of gas error must lead to all gas being consumed, and deal with in a single point of responsibility manner. 
This is line is redundant, having it may hide severe errors somewhere else.

go/interpreter/lfvm/instructions.go
```diff
@@ -989,6 +979,19 @@ func genericCreate(c *context, kind tosca.CallKind) {
 	c.refund += res.GasRefund
 }
 
+// checkInitCodeSize checks the size of the init code.
+// An error is returned if size is greater than MaxInitCodeSize.
+func checkMaxInitCodeSize(size uint64) error {
```
LuisPH3, line 984: This is ok, but it would be much better if this function offloads some work from the caller side: 

i.e.  `computeCodeSizeCost(...) (gas,error)`
Where the function is responsible for all the semantics related to code size: revision check, cost computations, overflows during computation.

This would streamline the op code and centralize all the side-effects related to init code .

LuisPH3 (changes requested):

LuisPH3 (approved):

LuisPH3 (approved):

simonlechner (approved):

LuisPH3 (approved):

simonlechner (approved):

Issue #684, opened 2024-08-26
Add comprehensive unit tests for LFVM's interpreter


HerbertJordan:
This issue is closely related to #751 
Refined responsibilities for the interpreter tests are:
- different configurations (normal mode, step-wise, statistics, logging), evaluate the following: 
   - [x]  `TestRunWithLogging`
   - [x] `TestRunBasic` -> `TestInterpreter_Vanilla_RunsWithoutOutput`

- Stop conditions: end of code, stopping codes, error triggered 
    - [x] `TestRunReturnsEmptyResultOnEmptyCode` rewrite to test reaching end of code (and case for empty code)    
- stack bounds violations
    - [x] `TestInterpreter_step_DetectsLowerStackLimitViolation` 
    - [x] `TestInterpreter_step_DetectsUpperStackLimitViolation` 
- static gas costs
    - [x] `TestContext_useGas_HandlesTerminationIfOutOfGas` use gas handles no termination any more 
    - [x] `TestStepsFailsOnTooLittleGas` rewrite to tests static gas costs.
    - [x] `TestInstructionsGasConsumption` move to instructions tests
- Detects non-executable code
    - [x] `TestStepsDetectsNonExecutableCode` 
- error/state propagation

Add/check unit tests for context utilities:
- useGas
- stack limits check
- getOutput
- `generateResult` 

Responsibilities which are to be handled in #751 (therefore not in interpreter tests):
- minimum required revision
- static-mode violations
- jumps are correctly processed (JUMPDEST)
- context calls are issued
    
Tests to be migrated to instruction tests: 
- `TestStepsProperlyHandlesJUMP_TO`
- `TestStepsDoesNotExecuteCodeIfStatic`

PR #788, opened 2024-09-19
LFVM: fix todo in interpreter tests

simonlechner:
The `OpcodeTest` still used a revision based an boolean types, this is updated to tosca's revision implementation. Some cosmetic changes have also been added to the file.

### DIFF
--- a/go/interpreter/lfvm/interpreter_test.go
+++ b/go/interpreter/lfvm/interpreter_test.go
@@ -94,9 +94,9 @@ func TestContext_isAtLeast_RespectsOrderOfRevisions(t *testing.T) {
 			},
 		}
 
-		for _, trg := range revisions {
-			if want, got := is >= trg, context.isAtLeast(trg); want != got {
-				t.Errorf("revision %v should be at least %v: %t, got %t", is, trg, want, got)
+		for _, target := range revisions {
+			if want, got := is >= target, context.isAtLeast(target); want != got {
+				t.Errorf("revision %v should be at least %v: %t, got %t", is, target, want, got)
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/interpreter_test.go
```diff
@@ -268,8 +267,8 @@ func TestInstructionsGasConsumption(t *testing.T) {
 	opCodes = append(opCodes, attachGasTo(10, POP_SWAP2_SWAP1_POP)...)
 	opCodes = append(opCodes, attachGasTo(4, POP_POP)...)
 	opCodes = append(opCodes, attachGasTo(6, generateOpCodesInRange(PUSH1_SHL, PUSH1_DUP1)...)...)
-	// opCodes = append(opCodes, applyGasTo(11, PUSH2_JUMP)...) // FIXME: this seems to be broken
-	opCodes = append(opCodes, attachGasTo(13, PUSH2_JUMPI)...)
+	//opCodes = append(opCodes, attachGasTo(11, PUSH2_JUMP)...)  // This test fails due to jump dest analysis
+	opCodes = append(opCodes, attachGasTo(13, PUSH2_JUMPI)...) // Jump dest analysis is skipped because argument is zero
```
HerbertJordan, line 271: Is this hiding the "FIXME"?

I guess these tests will become all obsolete if we go for Luis' testing approach. But if not, we would miss this one. 

If this test is never going to work, please state something like "deliberately skipped because ...".

HerbertJordan (dismissed):
Thanks for the clean-ups.

go/interpreter/lfvm/interpreter_test.go
simonlechner: Yes this got lost, I re-added it.

LuisPH3 (changes requested):
When I first engaged with interpreter/instructions testing, I rewrote this test with two goals in mind:
- completeness, there is no check that verifies that every OpCode is tested,
- concise, remove lots of boilerplate to make tests easier to write
- multiline, the definitions in this test are unreadable.

The rewrite was closed to prioritize other testing: https://github.com/Fantom-foundation/Tosca/pull/753/files

Neverthless I marked this function to be migrated to instructions, as this list is more useful to test explicit properties of each opCode instead of the interpreter itself. 
Please check #684 and then we can go back to see if this PR makes sense.

simonlechner:
This PR changes code which is going to be rewritten in the near future. Therefore, this PR is no longer required and closed.

PR #789, opened 2024-09-19
Remove accidentally committed fuzzing test data

HerbertJordan:
This data was accidentally added by #784.


LuisPH3:
I don't dislike this data, in addition to fast-forwarding the fuzzing corpus set to more sensible inputs, it will be run as normal unit tests with each test run, serving as a regression. 
It is true that it is a cryptic, maybe a readme file in the folder explaining where do they come from.

LuisPH3 (approved):
Although these files serve as unit tests against regressions, I don't think these are issues we must guard against and can be safely deleted.

PR #794, opened 2024-09-23
LFVM: Add test for every instruction static gas check

LuisPH3:
part of #684


go/interpreter/lfvm/interpreter.go
```diff
@@ -259,7 +259,7 @@ func steps(c *context, oneStepOnly bool) (status, error) {
 
 		// Consume static gas price for instruction before execution
 		if err := c.useGas(staticGasPrices.get(op)); err != nil {
-			return statusError, err
+			return statusError, errNotEnoughStaticGas
```
facuMH, line 262: `useGas` can only fail with an `errOutOfGas`, do we care if this is static or dynamic ?

LuisPH3: I am not sure. wdyt? 
The reason why I renamed it, is because there are many situations where out of gas can be produced, and I wanted to be specific about this check being the one failing.

facuMH: why do you want to be specific about this one case failing ?

facuMH (approved):

simonlechner (dismissed):

facuMH (approved):

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -626,28 +624,57 @@ func TestSteps_StaticContextViolation(t *testing.T) {
...
+				Static: false,
+				Code:   []byte{0x0},
+			}
+			ctxt.code = []Instruction{{op, 0}}
+			ctxt.stack.stackPointer = 50
+			ctxt.gas = staticGasPrices.get(op) - 1
+
+			if ctxt.gas < 0 {
+				t.Fatalf("invalid test setup: gas is negative")
+			}
```
HerbertJordan, line 659: Maybe, instead of filtering out explicit instructions with zero-cost here, just check the `staticGasPrices` for zero here and return early if so. Then you are automatically sure that you got the right instructions (unless wanting to verify this is part of the test's intention)

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -626,28 +624,57 @@ func TestSteps_StaticContextViolation(t *testing.T) {
...
 }
 
+var invalidOpCodeString = regexp.MustCompile(`^op\(0x[0-9a-fA-F]+\)`)
+
+func isValidOpCode(op OpCode) bool {
+	return !invalidOpCodeString.MatchString(op.String())
+}
+func isInvalidOpCode(op OpCode) bool {
+	return invalidOpCodeString.MatchString(op.String())
+}
```
HerbertJordan, line 676: I would like to avoid re-introducing the "valid" OpCode property, which does not have a satisfying, clean definition. I think you do not need it here. You can just run all op-codes and skip those where the static gas cost is 0.

HerbertJordan (changes requested):
Looks good, just minor restructuring requests.

go/interpreter/lfvm/interpreter_test.go
LuisPH3: In a later change, we figured out that it was important to check gas for each revision and zero was then more relevant, as it is the static cost of some "valid" ops after Berlin. 
The new code skips OP/Revision test when the expected static gas cost is zero.

facuMH (approved):

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -562,34 +562,37 @@ func TestStepsProperlyHandlesJUMP_TO(t *testing.T) {
...
+		t.Run(opCode.String(), func(t *testing.T) {
+			ctxt := getEmptyContext()
+			ctxt.params = tosca.Parameters{
+				Input:  []byte{},
+				Static: false,
+				Code:   []byte{0x0},
+			}
+			ctxt.code = []Instruction{{opCode, 0}}
+			ctxt.stack.stackPointer = 50
+			ctxt.gas = 1 << 32
```
HerbertJordan, line 589: `params`, `stack` size and `gas` are actually irrelevant for this test
```suggestion
			ctxt := getEmptyContext()
			ctxt.code = []Instruction{{opCode, 0}}
```

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -652,34 +655,37 @@ func TestSteps_StaticContextViolation(t *testing.T) {
...
+			forEachRevision(t, op, func(t *testing.T, revision tosca.Revision) {
+
+				expectedGas := getStaticGasPrices(revision).get(op)
+
+				ctxt := getEmptyContext()
+				ctxt.params = tosca.Parameters{
+					Input:  []byte{},
+					Static: false,
+					Code:   []byte{0x0},
+				}
```
HerbertJordan, line 671: Parameters are not consumed by this test, since operations are not getting executed.
```suggestion
```

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -652,34 +655,37 @@ func TestSteps_StaticContextViolation(t *testing.T) {
...
+				ctxt.code = []Instruction{{op, 0}}
+				ctxt.stack.stackPointer = 50
+				ctxt.gas = expectedGas - 1
+
+				if expectedGas == 0 {
```
HerbertJordan, line 676: nit: move up to where `expectedGas` was defined

HerbertJordan (approved):
Looks good, thanks!

facuMH (dismissed):

HerbertJordan (approved):

simonlechner (approved):

PR #796, opened 2024-09-23
Evmrs refactor

LorenzSchueler:
- add `u256::try_into_u64`
- call `code_reader.next()` after Opcode match
- do not check if address is sender or coinbase address

### DIFF
--- a/rust/src/ffi/evmc_vm.rs
+++ b/rust/src/ffi/evmc_vm.rs
@@ -79,7 +79,7 @@ extern "C" fn __evmc_set_option(
 }
 
 #[no_mangle]
-pub extern "C" fn evmc_create_evmrs() -> *const evmc_vm_t {
+pub(super) extern "C" fn evmc_create_evmrs() -> *const evmc_vm_t {
     let new_instance = evmc_vm_t {
         abi_version: EVMC_ABI_VERSION as i32,
         destroy: Some(__evmc_destroy),
--- a/rust/src/ffi/steppable_evmc_vm.rs
+++ b/rust/src/ffi/steppable_evmc_vm.rs
@@ -42,7 +42,6 @@ extern "C" fn __evmc_steppable_destroy(instance: *mut evmc_vm_steppable) {
     }
 }
 
-#[no_mangle]
 extern "C" fn __evmc_step_n(
     instance: *mut evmc_vm_steppable,
     host: *const evmc_host_interface,
--- a/rust/src/interpreter.rs
+++ b/rust/src/interpreter.rs
@@ -113,55 +113,46 @@ impl<'a> Interpreter<'a> {
                     consume_gas(&mut self.gas_left, 3)?;
                     let [value1, value2] = self.stack.pop()?;
                     self.stack.push(value1 + value2)?;
-                    self.code_reader.next();
                 }
                 Opcode::Mul => {
                     consume_gas(&mut self.gas_left, 5)?;
                     let [fac1, fac2] = self.stack.pop()?;
                     self.stack.push(fac1 * fac2)?;
-                    self.code_reader.next();
                 }
                 Opcode::Sub => {
                     consume_gas(&mut self.gas_left, 3)?;
                     let [value1, value2] = self.stack.pop()?;
                     self.stack.push(value1 - value2)?;
-                    self.code_reader.next();
                 }
                 Opcode::Div => {
                     consume_gas(&mut self.gas_left, 5)?;
                     let [value, denominator] = self.stack.pop()?;
                     self.stack.push(value / denominator)?;
-                    self.code_reader.next();
                 }
                 Opcode::SDiv => {
                     consume_gas(&mut self.gas_left, 5)?;
                     let [value, denominator] = self.stack.pop()?;
                     self.stack.push(value.sdiv(denominator))?;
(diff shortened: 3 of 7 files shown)


rust/src/ffi/evmc_vm.rs
LuisPH3, line 1: improve ffi is not enought to explain what was changed here. 
what was the problem?

rust/src/ffi/steppable_evmc_vm.rs
LuisPH3, line 1: idem, what happen here?

rust/src/interpreter.rs
```diff
@@ -796,10 +722,16 @@ impl<'a> Interpreter<'a> {
...
 
                     self.step_status_code = StepStatusCode::EVMC_STEP_STOPPED;
-                    self.code_reader.next();
                     break;
                 }
             }
+
+            if !(Opcode::Push1 as u8..=Opcode::Push32 as u8).contains(&(op as u8))
+                && op != Opcode::Jump
+                && op != Opcode::JumpI
```
LuisPH3, line 731: I am not sure was this implemented, but lfvm does an indiscriminate pc+1 every time. 
Could we clean this up?  

What I don't like is the semantics being separated. One would think that to introduce a new op implementation, we only need to implement the op code, but we have to look at this code a well. 
what do you think about this code.

rust/src/interpreter.rs
```diff
@@ -796,10 +722,16 @@ impl<'a> Interpreter<'a> {
                     }
 
                     self.step_status_code = StepStatusCode::EVMC_STEP_STOPPED;
-                    self.code_reader.next();
                     break;
                 }
             }
+
+            if !(Opcode::Push1 as u8..=Opcode::Push32 as u8).contains(&(op as u8))
```
LuisPH3, line 729: it is pretty safe to assume that push1-32 are sequential op code values. if you want to turn this into a comparison. 
Nevertheless I am curious, how good can rustc optimize this code ?

rust/src/types/opcode.rs
```diff
@@ -148,7 +148,7 @@ const REVERT: u8 = 0xFD;
 const INVALID: u8 = 0xFE;
 const SELFDESTRUCT: u8 = 0xFF;
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
LuisPH3, line 151: I don't have any problem with making opcode copy.  passing it by value should never be a problem

LuisPH3 (approved):
Removing pc increment from op implementation is a nice cleanup.
Some changes need a little background though.

rust/src/ffi/evmc_vm.rs
LorenzSchueler: `evmc_create_evmrs` is needed only in `steppable_evmc_vm.rs` so is is enough to have it `pub(super)`.

rust/src/ffi/steppable_evmc_vm.rs
LorenzSchueler: The callback does not need to be `#[no_mangle]` and coverage only works properly when it is removed.

rust/src/types/amount.rs
```diff
@@ -370,10 +370,22 @@ impl Shr for u256 {
...
 }
 
+#[derive(Debug, PartialEq)]
+pub struct U64Overflow;
+
 impl u256 {
     pub const ZERO: Self = Self(Uint256 { bytes: [0; 32] });
     pub const MAX: Self = Self(Uint256 { bytes: [0xff; 32] });
 
+    pub fn try_into_u64(self) -> Result<u64, U64Overflow> {
```
LuisPH3, line 380: This would be better modelled using [try_into](https://doc.rust-lang.org/beta/std/convert/trait.TryInto.html)
or try_from for uin64 (is this even possible?)

Is it there any literature against?

rust/src/types/amount.rs
```diff
@@ -370,10 +370,22 @@ impl Shr for u256 {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub struct U64Overflow;
```
LuisPH3, line 374: Error code could be directly StatusCode. To reduce error casting artifacts. 

If error management turns more complicated, we should consider using a library like thiserror or anyhow instead of polluting the namespace with more symbols. I think we won't be needing neither of these two in the context of the VM .

LuisPH3 (changes requested):
Sorry, I was missing the cast.
Can you consider specializing TryInto instead of the member function?

rust/src/interpreter.rs
LorenzSchueler: Well for most ops we indeed only have to implement only the op. The only exception are those ops which cause the pc to not be advanced by 1. I don't see any way around this check except for doing the `code_reader.next()` in ever op.

rust/src/types/amount.rs
LorenzSchueler: That does not work because in the error case the error gets mapped to different StatusCodes depending on the context.

rust/src/interpreter.rs
LorenzSchueler: I just checked. It's optimized away...

PR #798, opened 2024-09-23
CT driver. print error and generate regression file on panic

LuisPH3:
When a panic is triggered when running the CT, a fail report will be printed and a regression file generated.


go/ct/driver/run.go
```diff
@@ -97,6 +97,14 @@ func doRun(context *cli.Context) error {
 	}
 
 	opRun := func(state *st.State) rlz.ConsumerResult {
+		result := rlz.ConsumeContinue
+		defer func() {
+			if r := recover(); r != nil {
+				result = rlz.ConsumeAbort
+				issuesCollector.AddIssue(state, fmt.Errorf("failed to process input state %v: %w", state, err))
```
facuMH, line 104: maybe change the error msg to differentiate it from the usual "failures" to something that reflects it faild due to a panic or is `err` the msg from the panic ?

go/ct/driver/run.go
```diff
@@ -97,6 +97,14 @@ func doRun(context *cli.Context) error {
...
+			}
```
facuMH, line 105: maybe also at the end of this block you could return result to ensure that no further execution is done ??

go/ct/driver/run.go
```diff
@@ -97,6 +97,14 @@ func doRun(context *cli.Context) error {
 	}
 
 	opRun := func(state *st.State) rlz.ConsumerResult {
+		result := rlz.ConsumeContinue
+		defer func() {
+			if r := recover(); r != nil {
+				result = rlz.ConsumeAbort
```
HerbertJordan, line 103: I think this result over-ride only works for named return parameters. Right?

So you would have to change the signature to
```
func(state *st.State) (result rlz.ConsumerResult) {
```

HerbertJordan (approved):

go/ct/driver/run.go
simonlechner: The execution is already terminated by returning `ConsumeAbort`.

LuisPH3: The need for `issuesCollector` is that if not used, the program terminates without any output.
I will change the string, to emphasize that the execution panicked.

go/ct/driver/run.go
LuisPH3: I will change the string, to emphasize that the execution panicked.

facuMH (approved):

simonlechner (approved):

PR #799, opened 2024-09-23
evmrs: Test Interpreter::run

LorenzSchueler:
- add trait `ExecutionContextTrait` in order to mock `ExecutionContext`
- make all functions which took `ExecutionContext` as an argument generic over `ExecutionContextTrait`
- add `MockExecutionMessage`
- (partially) test `Interpreter::run`

### DIFF
--- a/rust/src/types/execution_context.rs
+++ b/rust/src/types/execution_context.rs
@@ -1,0 +1,121 @@
+use evmc_vm::{
+    AccessStatus, Address, ExecutionContext, ExecutionMessage, ExecutionResult, ExecutionTxContext,
+    StorageStatus, Uint256,
+};
+
+#[cfg_attr(test, mockall::automock)]
+pub trait ExecutionContextTrait {
+    /// Retrieve the transaction context.
+    fn get_tx_context(&self) -> &ExecutionTxContext;
+
+    /// Check if an account exists.
+    fn account_exists(&self, address: &Address) -> bool;
+
+    /// Read from a storage key.
+    fn get_storage(&self, address: &Address, key: &Uint256) -> Uint256;
+
+    // Set value of a storage key.
+    fn set_storage(&mut self, address: &Address, key: &Uint256, value: &Uint256) -> StorageStatus;
+
+    /// Get balance of an account.
+    fn get_balance(&self, address: &Address) -> Uint256;
+
+    /// Get code size of an account.
+    fn get_code_size(&self, address: &Address) -> usize;
+
+    /// Get code hash of an account.
+    fn get_code_hash(&self, address: &Address) -> Uint256;
+
+    /// Copy code of an account.
+    fn copy_code(&self, address: &Address, code_offset: usize, buffer: &mut [u8]) -> usize;
+
+    /// Self-destruct the current account.
+    fn selfdestruct(&mut self, address: &Address, beneficiary: &Address) -> bool;
+
+    /// Call to another account.
+    fn call(&mut self, message: &ExecutionMessage) -> ExecutionResult;
+
+    /// Get block hash of an account.
+    fn get_block_hash(&self, num: i64) -> Uint256;
+
+    /// Emit a log.
+    fn emit_log(&mut self, address: &Address, data: &[u8], topics: &[Uint256]);
+
+    /// Access an account.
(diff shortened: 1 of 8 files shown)


rust/src/types/execution_context.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn access_storage(&mut self, address: &Address, key: &Uint256) -> AccessStatus;
+
+    /// Read from a transient storage key.
+    fn get_transient_storage(&self, address: &Address, key: &Uint256) -> Uint256;
+
+    /// Set value of a transient storage key.
+    fn set_transient_storage(&mut self, address: &Address, key: &Uint256, value: &Uint256);
+}
+
+impl<'a> ExecutionContextTrait for ExecutionContext<'a> {
```
LuisPH3, line 57: This implementation seems a little superfluous.
Would it be possible to implement `ExecutionContextTrait` for `ExecutionContext` where the old methods are declared and avoid the forwarding?

rust/src/types/mock_execution_message.rs
```diff
@@ -0,0 +1,57 @@
+use evmc_vm::{Address, ExecutionMessage, MessageKind, Uint256};
+
+use crate::types::u256;
+
+pub const DEFAULT_INIT_GAS: u64 = 1_000_000;
```
LuisPH3, line 5: is this used anywhere else?  Otherwise it is polluting the namespace.

rust/src/types/mod.rs
```diff
@@ -1,13 +1,19 @@
...
 mod opcode;
 mod stack;
 mod tx_context;
 
 pub use amount::u256;
 pub use code_reader::{CodeReader, GetOpcodeError};
+pub use execution_context::*;
 pub use memory::Memory;
+#[cfg(test)]
+pub use mock_execution_message::{MockExecutionMessage, DEFAULT_INIT_GAS};
```
LuisPH3, line 16: I guess there will be some day we have to reconsider include habits. In the meanwhile we forward symbols for visibility. I don't think we need to forward this one.

rust/src/types/execution_context.rs
```diff
@@ -0,0 +1,121 @@
+use evmc_vm::{
+    AccessStatus, Address, ExecutionContext, ExecutionMessage, ExecutionResult, ExecutionTxContext,
+    StorageStatus, Uint256,
+};
+
+#[cfg_attr(test, mockall::automock)]
```
LuisPH3, line 6: Does this macro define `MockExecutionContextTrait`? 

does it define it for `test` configuration or for all of them ?

rust/src/types/execution_context.rs
```diff
@@ -0,0 +1,121 @@
...
+pub trait ExecutionContextTrait {
```
LuisPH3, line 7: I am not a big fan of Naming types by its type. 
This is an execution context. The derived types are execution context instances, each with its own flavor. 

-  `MockExecutionContextTrait` is not a good name 
- the old `ExecutionContext<'a>` should be `EvmcExectutionContext<'a>`

LuisPH3 (changes requested):
This is a nice adition.

I think the main misunderstanding is that the mock trait is some kind of testing tooling, and the intent should be that the trait is a functional cornerstone that gets mocked. having a code that extends functionality feels odd.

rust/src/types/execution_context.rs
LorenzSchueler: yet it does but only if the mock feature is enabled (which is the case for tests)

rust/src/types/execution_context.rs
LorenzSchueler: no because ExecutionContext comes from evmc

rust/src/types/execution_context.rs
LorenzSchueler: ExecutionContext is a type from evmc. The naming of ExecutionContextTrait is indeed not ideal but at the end it is a trait with the same methods as the struct ExecutionContext.

LuisPH3: i see :(

LuisPH3 (approved):

PR #800, opened 2024-09-24
evmrs: Add bench and tests scripts

LorenzSchueler:
- add test bash script
- add bench bash script

### DIFF
--- a/rust/scripts/bench.sh
+++ b/rust/scripts/bench.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# Run Go interpreter benchmarks for evmrs, evmzero, lfvm and geth in branch origin/main and evmrs in branch origin/evmrs-main.
+# This script must be called from the parent directory of Tosca.
+
+mkdir -p benches
+
+cd Tosca
+git fetch --all
+
+RUNS=20
+TIMEOUT=1h
+REF_BRANCH="origin/main"
+NEW_BRANCH="origin/evmrs-main"
+
+BRANCH=$REF_BRANCH
+git checkout ${BRANCH}
+make clean-rust
+make
+
+GIT_REF=$(git show-ref --hash=7 ${BRANCH})
+INTERPRETERS=("evmrs" "evmzero" "lfvm" "geth")
+for INTERPRETER in "${INTERPRETERS[@]}"; do
+    BENCH_NAME=$(basename ${BRANCH})-${GIT_REF}-${INTERPRETER}-${RUNS}
+    echo running ${BENCH_NAME}
+    go test ./go/integration_test/interpreter \
+        --run none -bench ^Benchmark[a-zA-Z]+/./${INTERPRETER}$ \
+        -timeout $TIMEOUT -count $RUNS | tee ../benches/${BENCH_NAME}
+    sed -i "s/${INTERPRETER}-//g" ../benches/${BENCH_NAME}
+done
+
+BRANCH=$NEW_BRANCH
+git checkout ${BRANCH}
+make clean-rust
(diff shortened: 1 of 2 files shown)


rust/scripts/test.sh
LuisPH3, line 1: please document how to run (from which folder and other needed stuff) 

It is not clear to me what does this script do

rust/scripts/bench.sh
LuisPH3, line 1: please document how to run (from which folder and other needed stuff)

rust/scripts/bench.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+mkdir -p ~/benches
+
+cd Tosca
+git fetch --all
+
+RUNS=20
+TIMEOUT=1h
+REF_BRANCH="origin/main"
```
LuisPH3, line 10: note: because you checkout this, this could be a commit as well.  or a tag. any git ref would do

LuisPH3 (changes requested):
Please document use of these files.

rust/scripts/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# Run Go tests and CT for evmrs on branch origin/evmrs-main.
+# This script must be called from the parent directory of Tosca.
+
+mkdir -p ~/tests
```
LuisPH3, line 6: please dont create folders in users home.

rust/scripts/test.sh
```diff
@@ -0,0 +1,21 @@
...
+
+git checkout ${BRANCH}
+make clean-rust
+make tosca-rust
+
+GIT_REF=$(git show-ref --hash=7 ${BRANCH})
+
+export RUST_BACKTRACE=full
+go test ./go/... | tee ~/tests/ct-full-${GIT_REF}
+go run ./go/ct/driver run --full-mode evmrs | tee ~/logs/go-test-${GIT_REF}
```
LuisPH3, line 21: if logs do not exist, this will fail. 
please dont create folders in users home.

rust/scripts/test.sh
LuisPH3, line 1: I don't think we need this file. I see the value, but It creates a little of noise on how/where to place output files.
Furthermore, I would recommend that you program this as a VS Code task for your use. 
Validation runs will be scripted in jenkins and the rustvm will be tested on a nightly plan.

LuisPH3 (changes requested):

rust/scripts/test.sh
LorenzSchueler: This is mainly intended to be used in a cloud instance...

LuisPH3: aditionally, makefile has coverage for evmzero as a target:
```makefile
ct-coverage-evmzero: tosca-cpp-coverage
ct-coverage-evmzero: 
        go run ./go/ct/driver run evmzero ; \
        echo "Coverage report generated in cpp/build/coverage/index.html"
        @cd cpp/build ; \
        cmake --build .  --target coverage 
```
Maybe we could enable rust coverage in an analogous manner. (note, reports end up in a folder inside of the build folder )

rust/scripts/test.sh
LorenzSchueler: will just replace `~` with `..`

rust/scripts/test.sh
LorenzSchueler: For coverage this would make sense but not for comparing benchmarks. And even for the tests it does not make to much sense because CT in full mode is also not included for the other vms in the makefile.

LuisPH3 (approved):

Issue #751, opened 2024-09-10
Comprehensive unit testing of LFVM operations


HerbertJordan:
This task focuses on testing properties of VM instructions, in particular code in `instructions.go` and `super_instructions.go` 

Common properties to opcodes are: 
- minimum required version
- static mode violation
- dynamic cost computation, leading to out of gas or overflow
- context calls are issued.

Properties common to certain ops are: 
- JUMPDEST check for any jump operation (except for JUMP_TO) 
    - [x] TestInstructions_JumpInstructionsCheckJUMPDEST
- EIP-2929 COLD/WARM dynamic cost introduced in Berlin
     - SLOAD, 
     - SSTORE
     - *CALL, 
     - BALANCE, 
     - EXT*
     - SELFDESTRUCT 
- Memory operations
...

LuisPH3:
Attempt to refactor existing end-to-end test  #753

LuisPH3:
- [ ] minimum required version
- [ ] static mode violation 
- [ ] dynamic cost
- [ ] context calls are issued.

LuisPH3:
note: opSwap2_Swap1_Pop_Jump never checked for jumpdest and is never tested for it.

facuMH:
After the merge of #837 #839 #841 #838 there should be still need to test specific branches of
- Calldataload
- opExp
- opExtcodehash
- opExtCodeCopy
- returndatacopy

facuMH:
- opLog also needs a test for memory expansion

LuisPH3:
Regarding context calls, It would be interesting not only to test that calls are issued, but that the arguments passed to external context calls are correct as well:

- [x] AccessAccount
    - TestSelfDestruct_ExistingAccountToNewBeneficiary
- [x] Balance
    - TestSelfDestruct_ExistingAccountToNewBeneficiary
- [ ] GetCodeSize
- [ ] GetCode
- [x] AccountExists
     - TestSelfDestruct_ProperlyReportsNotEnoughGas
- [ ] AccessStorage
- [ ] GetStorage
- [ ] SetStorage
- [ ] Call
- [ ] EmitLog #856 
- [x] SelfDestruct
     - TestSelfDestruct_ExistingAccountToNewBeneficiary
- [ ] GetTransientStorage
- [ ] SetTransientStorage

PR #802, opened 2024-09-24
LFVM: Test interpreter dispatcher

LuisPH3:
part of #684

This PR adds a unit test to check that every opcode can get dispatched by the interpreter. 
Every opcode gets a context where they can sucessfuly execute. The interpreter will dispatch the opCode to the appropiate handling function. 

Note: 
This change removes an old test, because of the change of paradigm, we are testing properties and not the bulk on instructions here. The one single problem is that the test for STORE and SLOAD will be deleted and such instructions are not yet tested. A plan to test WARM/COLD instructions is in the making here #751

### DIFF
--- a/go/interpreter/lfvm/interpreter_test.go
+++ b/go/interpreter/lfvm/interpreter_test.go
@@ -679,32 +679,34 @@ func isJump(op OpCode) bool {
 	return idx != -1
 }
 
-var _introducedInRevision = newOpCodePropertyMap(func(op OpCode) tosca.Revision {
-	switch op {
-	case BASEFEE:
-		return tosca.R10_London
-	case PUSH0:
-		return tosca.R12_Shanghai
-	case BLOBHASH:
-		return tosca.R13_Cancun
-	case BLOBBASEFEE:
-		return tosca.R13_Cancun
-	case TLOAD:
-		return tosca.R13_Cancun
-	case TSTORE:
-		return tosca.R13_Cancun
-	case MCOPY:
-		return tosca.R13_Cancun
-	}
-	return tosca.R07_Istanbul
-})
-
+// forEachRevision runs a test for each revision starting from the revision
+// where the operation was introduced.
 func forEachRevision(
 	t *testing.T, op OpCode,
 	f func(t *testing.T, revision tosca.Revision)) {
 
+	introducedIn := newOpCodePropertyMap(func(op OpCode) tosca.Revision {
+		switch op {
+		case BASEFEE:
+			return tosca.R10_London
+		case PUSH0:
+			return tosca.R12_Shanghai
+		case BLOBHASH:
+			return tosca.R13_Cancun
+		case BLOBBASEFEE:
+			return tosca.R13_Cancun
+		case TLOAD:
+			return tosca.R13_Cancun
+		case TSTORE:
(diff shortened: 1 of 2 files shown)


go/interpreter/lfvm/interpreter_test.go
```diff
@@ -699,6 +628,152 @@ func BenchmarkSatisfiesStackRequirements(b *testing.B) {
...
+			stack.data[i] = *uint256.NewInt(uint64(counter))
+		}
+	}
+
+	return nil
+}
+
+var _isUndefinedOpCodeRegex = regexp.MustCompile(`^op\(0x[0-9A-Fa-f]+\)$`)
+
+func isExecutable(op OpCode) bool {
```
facuMH, line 726: would it be terrible if these 2 are inlined ?

facuMH (dismissed):

go/interpreter/lfvm/interpreter_test.go
LuisPH3: The reason why I did it this way is that I would not like to have a regex compiled on each loop iteration. Although it is test code. 
I refactored it, and now it is part of the function.

facuMH: I thought of not instantiating regex, but my thought as well was "this is not a performance critical function"

facuMH (dismissed):

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -183,126 +184,68 @@ func TestInterpreter_step_DetectsUpperStackLimitViolation(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("unexpected creating stack: %v", err)
+				}
+
+				_, err = vanillaRunner{}.run(&ctx)
+
+				var expectedErr error
+				if !isExecutable(op) {
+					expectedErr = errInvalidOpCode
+				}
```
HerbertJordan, line 243: A thought: since the check for the detection of invalid instructions was already covered by other tests (if I remember correctly I saw that in one of your other PRs), you could accept either an `errInvalidOpCode` or a `nil` error and avoid the need to identify invalid instructions in this test too.

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -699,6 +611,143 @@ func BenchmarkSatisfiesStackRequirements(b *testing.B) {
...
+
+	return nil
+}
+
+func isJump(op OpCode) bool {
+	ops := append(op.decompose(), op)
+	idx := slices.IndexFunc(ops, func(op OpCode) bool {
+		return op == JUMP || op == JUMPI
+	})
+	return idx != -1
```
HerbertJordan, line 712:
```suggestion
	return slices.ContainsFunc(ops, func(op OpCode) bool {
		return op == JUMP || op == JUMPI
	})
```

HerbertJordan (approved):
Nice test tests!

go/interpreter/lfvm/interpreter_test.go
LuisPH3: Later change deleted that test:  the reason being naming, I could not name this test dispachesAllInstructions and start filtering non-executable ones.
wdyt?

facuMH (approved):

PR #803, opened 2024-09-24
LFVM: Check jump instructions JUMPDEST test

LuisPH3:
part of #751 

This test verifies that every jump operation does JUMPDEST check.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1336,18 +1336,8 @@ func TestInstructions_JumpInstructionsCheckJUMPDEST(t *testing.T) {
 
 	for op, test := range tests {
 		t.Run(op.String(), func(t *testing.T) {
-
-			ctxt := context{
-				params: tosca.Parameters{
-					BlockParameters: tosca.BlockParameters{
-						Revision: tosca.R13_Cancun,
-					},
-				},
-				stack:  NewStack(),
-				memory: NewMemory(),
-				gas:    1 << 32,
-				code:   Code{{op, 0}},
-			}
+			ctxt := getEmptyContext()
+			ctxt.code = Code{{op, 0}}
 			for _, v := range test.stack {
 				ctxt.stack.push(uint256.NewInt(v))
 			}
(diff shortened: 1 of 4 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -1173,3 +1173,71 @@ func TestOpEndWithResult_ReportOverflow(t *testing.T) {
...
+
+			ctxt := context{
+				params: tosca.Parameters{
+					BlockParameters: tosca.BlockParameters{
+						Revision: tosca.R13_Cancun,
+					},
+				},
+				stack:  NewStack(),
+				memory: NewMemory(),
+				gas:    1 << 32,
```
facuMH, line 1230: unnecessary specifications.

facuMH (approved):

go/interpreter/lfvm/instructions_test.go
LuisPH3: gas is needed, but memory not.

facuMH: sorry, yes, gas does not need to be so big, but it's fine as is

facuMH (dismissed):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1291,3 +1291,61 @@ func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
 		}
 	}
 }
+
+func TestInstructions_JumpInstructionsCheckJUMPDEST(t *testing.T) {
```
HerbertJordan, line 1295: For a follow-up: a related test that should be added for conditional jumps: if the condition is false, it must not test the jump target for pointing to a JUMPDEST; invalid target destinations must be accepted if the branch is not taken.

HerbertJordan (approved):
Nice, thanks!

go/interpreter/lfvm/instructions_test.go
LuisPH3: I think this is not that far off. I extend the scope of this PR.

facuMH (approved):

PR #795, opened 2024-09-23
LFVM: Refactor Run method to return tuple (status, error)

LuisPH3:
This refactor facilitates #684 

The change removes `context.status` and `statusError` in favor of a `(status, error)` return type for the complete execution of the lfvm.

- A extra change in the interpreter was needed to correct program counter position after error. This was not a problem in main so far because `statusError` was returned. Since it does not exist anymore, the program counter was read at the wrong position. 
- Benchmarks have been reordered, to follow order of relevance. 

The most delicate change is in [go/interpreter/lfvm/ct.go](https://github.com/Fantom-foundation/Tosca/pull/795/files#diff-8732597f25fb78bbf7cc4df47e41abef29441a1e8690763ea444bc4c37c15f45), where the (status, error) pair needs to be converted into (*st.State, error). This has a semantic difference because Error inside the VM is a valid state (outOfGas, staticModeViolation...) but error in the ct adapter is a conversion error, where unsupported states are translated to ct code. For this reason both `state` and `error` are collected and errors are translated into a `failed state`

TODO:

- [x] ct
- [x] benchmarks

PR #804, opened 2024-09-24
LFVM: Return/Revert

facuMH:
In evm.codes it states that both `RETURN` and `REVERT` much charge for the memory expansion cost. However in our previous implementation this was done as part of getting the data from the context to build the `tosca.Result` at the end of a `run`. 
This PR:
- moves that memory expansion into the corresponding operation implementations, adding a function for the common parts and testing it. 
- CT conversion used to rely on a previous function called `getOutput` to also get the `ReturnData`, this now has it's own implementation.

### DIFF
--- a/go/interpreter/lfvm/ct.go
+++ b/go/interpreter/lfvm/ct.go
@@ -96,11 +96,6 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 		return nil, err
 	}
 
-	result, err := getOutput(status, ctxt)
-	if err != nil {
-		executionErr = err
-	}
-
 	if executionErr == nil {
 		state.Pc = pcMap.lfvmToEvm[ctxt.pc]
 	}
@@ -114,7 +109,9 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 	state.Stack = convertLfvmStackToCtStack(ctxt.stack, state.Stack)
 	state.Memory = convertLfvmMemoryToCtMemory(ctxt.memory)
 	state.LastCallReturnData = common.NewBytes(ctxt.returnData)
-	state.ReturnData = common.NewBytes(result)
+	if status == statusReturned || status == statusReverted {
+		state.ReturnData = common.NewBytes(ctxt.returnData)
+	}
 
 	return state, nil
 }
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -22,15 +22,7 @@ func opStop() status {
 	return statusStopped
 }
 
-func opRevert(c *context) (status, error) {
-	return genericReturnRevert(c, statusReverted)
-}
-
-func opReturn(c *context) (status, error) {
-	return genericReturnRevert(c, statusReturned)
-}
-
-func genericReturnRevert(c *context, status status) (status, error) {
+func opReturningOp(c *context, status status) (status, error) {
 	offset := *c.stack.pop()
 	size := *c.stack.pop()
 	if err := checkSizeOffsetUint64Overflow(&offset, &size); err != nil {
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -22,16 +22,23 @@ func opStop() status {
 	return statusStopped
 }
 
-func opRevert(c *context) status {
-	c.resultOffset = *c.stack.pop()
-	c.resultSize = *c.stack.pop()
-	return statusReverted
+func opRevert(c *context) (status, error) {
+	return genericReturnRevert(c, statusReverted)
 }
 
-func opReturn(c *context) status {
-	c.resultOffset = *c.stack.pop()
-	c.resultSize = *c.stack.pop()
-	return statusReturned
+func opReturn(c *context) (status, error) {
+	return genericReturnRevert(c, statusReturned)
+}
+
+func genericReturnRevert(c *context, status status) (status, error) {
+	offset := *c.stack.pop()
+	size := *c.stack.pop()
+	if err := checkSizeOffsetUint64Overflow(&offset, &size); err != nil {
+		return status, err
+	}
+	var err error
+	c.returnData, err = c.memory.getSlice(offset.Uint64(), size.Uint64(), c)
+	return status, err
 }
 
 func opPc(c *context) {
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -22,15 +22,15 @@ func opStop() status {
 	return statusStopped
 }
 
-func opReturningOp(c *context, status status) (status, error) {
+func opEndWithResult(c *context) error {
 	offset := *c.stack.pop()
 	size := *c.stack.pop()
 	if err := checkSizeOffsetUint64Overflow(&offset, &size); err != nil {
-		return status, err
+		return err
 	}
 	var err error
 	c.returnData, err = c.memory.getSlice(offset.Uint64(), size.Uint64(), c)
-	return status, err
+	return err
 }
 
 func opPc(c *context) {
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1147,32 +1147,28 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
 	}
 }
 
-func TestRevertReturn_ReturnsExpectedState(t *testing.T) {
+func TestOpEndWithResult_ReturnsExpectedState(t *testing.T) {
 	c := getEmptyContext()
 	c.stack.push(uint256.NewInt(1))
 	c.stack.push(uint256.NewInt(1))
 	c.memory.store = []byte{0x1, 0xff, 0x2}
-	wantStatus := statusReverted
 
-	status, err := opReturningOp(&c, wantStatus)
+	err := opEndWithResult(&c)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
-	}
-	if status != wantStatus {
-		t.Errorf("unexpected status, wanted %v, got %v", wantStatus, status)
 	}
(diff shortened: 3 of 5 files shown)


go/interpreter/lfvm/instructions.go
```diff
@@ -22,16 +22,23 @@ func opStop() status {
 	return statusStopped
 }
 
-func opRevert(c *context) status {
-	c.resultOffset = *c.stack.pop()
-	c.resultSize = *c.stack.pop()
-	return statusReverted
+func opRevert(c *context) (status, error) {
```
LuisPH3, line 25: I have seen this pattern before with call and create, it produces symbols and names that we may not need. 

Better rename  `genericReturnRevert` into `opReturningOp`  and call it in the switch:
```
      case RETURN:
		status, err = opReturningOp(c, statusReturned)
      case REVERT:
		status, err = opReturningOp(c, statusReverted)			
```
This change saves us from one line functions that do not contribute

go/interpreter/lfvm/ct.go
```diff
@@ -105,11 +100,19 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 	state.Stack = convertLfvmStackToCtStack(ctxt.stack, state.Stack)
 	state.Memory = convertLfvmMemoryToCtMemory(ctxt.memory)
 	state.LastCallReturnData = common.NewBytes(ctxt.returnData)
-	state.ReturnData = common.NewBytes(result)
+	state.ReturnData = common.NewBytes(getOutput(ctxt))
 
 	return state, nil
 }
 
+func getOutput(ctxt *context) []byte {
```
LuisPH3, line 108: nit: this could be a lambda in stepn, to avoid having more symbol names.

LuisPH3 (changes requested):
This worked out pretty well.
I have some concerns about unnecessary symbols. And the overall naming of functions and tests once fixed:

LuisPH3 (approved):
:+1:

go/interpreter/lfvm/ct.go
```diff
@@ -100,12 +95,20 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
...
+		return res
+	}
+
 	state.Gas = ctxt.gas
 	state.GasRefund = ctxt.refund
 	state.Stack = convertLfvmStackToCtStack(ctxt.stack, state.Stack)
 	state.Memory = convertLfvmMemoryToCtMemory(ctxt.memory)
 	state.LastCallReturnData = common.NewBytes(ctxt.returnData)
-	state.ReturnData = common.NewBytes(result)
+	state.ReturnData = common.NewBytes(getOutput(ctxt))
```
HerbertJordan, line 111: This would be less code and easier to read if inlined.

go/interpreter/lfvm/instructions.go
```diff
@@ -22,16 +22,15 @@ func opStop() status {
...
-	c.resultOffset = *c.stack.pop()
-	c.resultSize = *c.stack.pop()
-	return statusReverted
-}
-
-func opReturn(c *context) status {
-	c.resultOffset = *c.stack.pop()
-	c.resultSize = *c.stack.pop()
-	return statusReturned
+func opReturningOp(c *context, status status) (status, error) {
```
HerbertJordan, line 25: "op" is present twice in this function name .. how about
```suggestion
func opEndWithResult(c *context, status status) (status, error) {
```

HerbertJordan, line 25: The status argument is only passed in to be returned. This could be dropped. The caller knows the result state already.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1147,3 +1147,34 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
 		})
 	}
 }
+
+func TestRevertReturn_ReturnsExpectedState(t *testing.T) {
```
HerbertJordan, line 1151: nit: please align the test name prefix with the function under test

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -721,17 +703,7 @@ func benchmarkFib(b *testing.B, arg int, with_super_instructions bool) {
...
-			b.Fatalf("unexpected length of end; wanted 32, got %d", size.Uint64())
-		}
-		res := make([]byte, size.Uint64())
-		offset := ctxt.resultOffset
-
-		data, err := ctxt.memory.getSlice(offset.Uint64(), size.Uint64(), &ctxt)
-		if err != nil {
-			b.Fatalf("unexpected error: %v", err)
-		}
+		res := make([]byte, len(ctxt.returnData))
```
HerbertJordan, line 706: Why make a new buffer and copy the `ctxt.returnData` instead of just using the `ctxt.returnData`?

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -417,45 +416,69 @@ func TestRunGenerateResult(t *testing.T) {
...
-		"getOutput fail": {func(ctx *context) {
-			ctx.status = statusReturned
-			ctx.resultSize = uint256.Int{1, 1}
-		}, nil, tosca.Result{Success: false}},
+		"max init code": {
+			status:         statusError,
+			expectedResult: tosca.Result{Success: false}},
+		"error": {
+			status:         statusError,
+			expectedResult: tosca.Result{Success: false}},
```
HerbertJordan, line 430: These two test cases are identical ...

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -417,45 +416,69 @@ func TestRunGenerateResult(t *testing.T) {
...
+				GasRefund: baseRefund},
+		},
+		"unknown status": {
+			status:         statusError + 1,
+			expectedErr:    fmt.Errorf("unexpected error in interpreter, unknown status: %v", statusError+1),
+			expectedResult: tosca.Result{}},
+	}
+
+	getCtxt := func() context {
+		ctxt := getEmptyContext()
```
HerbertJordan, line 470: As recently learned, `getEmptyContext` is not that empty as it seems. Maybe it should be replaced with a &context{..} initializer.

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -417,45 +416,69 @@ func TestRunGenerateResult(t *testing.T) {
...
+				GasLeft:   baseGas,
+				GasRefund: baseRefund},
+		},
+		"unknown status": {
+			status:         statusError + 1,
+			expectedErr:    fmt.Errorf("unexpected error in interpreter, unknown status: %v", statusError+1),
+			expectedResult: tosca.Result{}},
+	}
+
+	getCtxt := func() context {
```
HerbertJordan, line 469: this is only used once and small enough => inline

HerbertJordan (changes requested):
Good improvement, but needs improvement.

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -417,45 +416,69 @@ func TestRunGenerateResult(t *testing.T) {
...
+			status:         statusError,
+			expectedResult: tosca.Result{Success: false}},
+		"returned": {
+			status:     statusReturned,
+			returnData: baseOutput,
+			expectedResult: tosca.Result{
+				Success:   true,
+				Output:    baseOutput,
+				GasLeft:   baseGas,
+				GasRefund: baseRefund},
```
HerbertJordan, line 438: nit: should end with a line break

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -417,45 +416,69 @@ func TestRunGenerateResult(t *testing.T) {
...
+			expectedResult: tosca.Result{
+				Success:   true,
+				GasLeft:   baseGas,
+				GasRefund: baseRefund},
+		},
+		"unknown status": {
+			status:         statusError + 1,
+			expectedErr:    fmt.Errorf("unexpected error in interpreter, unknown status: %v", statusError+1),
+			expectedResult: tosca.Result{}},
+	}
```
HerbertJordan, line 467: #795 has a cleaner version of this refactor.

HerbertJordan (changes requested):

go/interpreter/lfvm/instructions.go
facuMH: right, this was to forward the right status, and error, but as you mention it can be avoided.

go/interpreter/lfvm/interpreter_test.go
facuMH: maybe in a different PR we should rename that function, although sometimes it is useful to get a context with randomly generated data.

go/interpreter/lfvm/interpreter_test.go
facuMH: ah, yes, agree. I'll imitate that refactor.

go/interpreter/lfvm/interpreter_test.go
facuMH: will delete max init code, since it is covered in error case as well.

LuisPH3 (approved):

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -468,23 +456,15 @@ func TestRunGenerateResult(t *testing.T) {
...
 		t.Run(fmt.Sprintf("%v", name), func(t *testing.T) {
 
-			ctxt := getCtxt()
-			if test.setup != nil {
-				test.setup(&ctxt)
-			}
+			ctxt := context{}
+			ctxt.refund = baseRefund
+			ctxt.gas = baseGas
+			ctxt.returnData = bytes.Clone(baseOutput)
```
LuisPH3, line 467: this is a nice detail. I guess it is important to check that a different instance must have the same values.

LuisPH3 (approved):

simonlechner (approved):

LuisPH3:
@HerbertJordan refactor to merge

HerbertJordan (approved):

PR #806, opened 2024-09-25
evmrs: add tests and remove some unsafe code

LorenzSchueler:
- remove two unsafe blocks and replace them with code which looks like it allocated but (at least in compiler explorer) is optimized away
- switch parameter order of `consume_gas` for better readability and consistency
- add `Gas` struct and make all functions in gas.rs methods of `Gas`
- add (more) tests for memory, stack, gas and helpers

### DIFF
--- a/rust/src/types/memory.rs
+++ b/rust/src/types/memory.rs
@@ -4,7 +4,7 @@ use evmc_vm::StatusCode;
 
 use crate::{
     types::u256,
-    utils::{consume_copy_cost, consume_gas, word_size},
+    utils::{word_size, Gas},
 };
 
 #[derive(Debug)]
@@ -23,7 +23,7 @@ impl Memory {
         self.0.len() as u64
     }
 
-    fn expand(&mut self, new_len_bytes: u64, gas_left: &mut u64) -> Result<(), StatusCode> {
+    fn expand(&mut self, new_len_bytes: u64, gas_left: &mut Gas) -> Result<(), StatusCode> {
         let current_len = self.0.len() as u64;
         let new_len = word_size(new_len_bytes)? * 32; // word_size just did a division by 32 so * will not overflow
         if new_len > current_len {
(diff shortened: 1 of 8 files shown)


rust/src/utils/gas.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
 };
 
 #[inline(always)]
-pub fn consume_gas(gas_left: &mut u64, gas: u64) -> Result<(), StatusCode> {
+pub fn consume_gas(gas: u64, gas_left: &mut u64) -> Result<(), StatusCode> {
```
LuisPH3, line 10: Because you decided to change the order of these two parameters, I realized that this is slightly awkward. 
You still need to add `&mut` in the caller side, which makes a little more clear which of the two parameters is the cost and which one is the accumulator. 
I still think that the best would be to have a member function on the gas or context. wdyt?

rust/src/utils/gas.rs
```diff
@@ -27,35 +27,202 @@ pub fn consume_positive_value_cost(value: &u256, gas_left: &mut u64) -> Result<(
...
     }
 }
 
-/// consume 3 * minimum_word_size
 #[inline(always)]
-pub fn consume_copy_cost(gas_left: &mut u64, len: u64) -> Result<(), StatusCode> {
-    let (cost, cost_overflow) = word_size(len)?.overflowing_mul(3);
-    if cost_overflow {
-        return Err(StatusCode::EVMC_OUT_OF_GAS);
+pub fn consume_copy_cost(len: u64, gas_left: &mut u64) -> Result<(), StatusCode> {
```
LuisPH3, line 54: I am not sure if we need a function at all. It would not be too large if you write:
` consume_gas(word_size(len)? * 3, &mut gas)`,

rust/src/types/memory.rs
```diff
@@ -129,7 +129,63 @@ mod tests {
...
+            memory.expand(u64::MAX, &mut 1_000),
+            Err(StatusCode::EVMC_OUT_OF_GAS)
+        );
+    }
+
+    #[test]
+    fn consume_expansion_cost() {
+        let memory = Memory::new(Vec::new());
+        let mut gas_left = 0;
+        assert_eq!(memory.consume_expansion_cost(&mut gas_left, 0), Ok(()));
```
LuisPH3, line 160: `consume_expansion_cost` did not change order, still has `(gas_left, cost)` format, the other functions have inverted them,

LorenzSchueler: The second parameter is in this case not the cost but the length to which the memory should be expanded. But I agree that the order should be switched.

rust/src/utils/gas.rs
LorenzSchueler: I find it better this way because it is self documenting.

LuisPH3 (approved):

Issue #690, opened 2024-08-26
Resolve remaining TODOs in the LFVM code base


HerbertJordan:
There are several remaining TODOs in the LFVM code base that need to be resolved. A list can be obtained [here](https://github.com/search?q=repo%3AFantom-foundation%2FTosca+path%3A%2F%5Ego%5C%2Finterpreter%5C%2Flfvm%5C%2F%2F+TODO&type=code). All of these should be processed and resolved.

PR #808, opened 2024-09-25
LFVM: test EIP2929 warm/cold acceses charging

facuMH:
Part of #751

This PR fixes and tests many of the changes introduced in EIP-2929 (https://eips.ethereum.org/EIPS/eip-2929) :
- Fixes static costs of BALANCE, EXTCODECOPY, EXTCODESIZE and EXTCODEHASH to zero. All gas for these operations should be dynamically calculated. 
- removes `gasEip2929AccountCheck` because it was duplicated functionality from `getAccessCost`, which is already tested.
- adds test checking for the extra charging of warm/cold accesses in Berlin and after, in the relevant opeations.

NOTE: SelfDestruct is a special case and has its own pre-existing tests.

- [x] ct

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -249,7 +249,10 @@ func opSstore(c *context) error {
 	cost := tosca.Gas(0)
 	if c.isAtLeast(tosca.R09_Berlin) &&
 		c.context.AccessStorage(c.params.Recipient, key) == tosca.ColdAccess {
-		cost += 2100
+		if err := c.useGas(ColdSloadCostEIP2929); err != nil {
+			// TODO: this error can never be triggered, because if we get here, we already have 2300.
+			return err
+		}
 	}
 
 	storageStatus := c.context.SetStorage(c.params.Recipient, key, value)
@@ -652,8 +655,10 @@ func opGasPrice(c *context) {
 func opBalance(c *context) error {
 	slot := c.stack.peek()
 	address := tosca.Address(slot.Bytes20())
-	if err := gasEip2929AccountCheck(c, address); err != nil {
-		return err
+	if c.isAtLeast(tosca.R09_Berlin) {
+		if err := c.useGas(getAccessCost(c.context.AccessAccount(address))); err != nil {
+			return err
+		}
 	}
 	balance := c.context.GetBalance(address)
 	slot.SetBytes32(balance[:])
@@ -817,19 +822,23 @@ func opCodeCopy(c *context) error {
 
 func opExtcodesize(c *context) error {
 	top := c.stack.peek()
-	addr := tosca.Address(top.Bytes20())
-	if err := gasEip2929AccountCheck(c, addr); err != nil {
-		return err
+	address := tosca.Address(top.Bytes20())
+	if c.isAtLeast(tosca.R09_Berlin) {
+		if err := c.useGas(getAccessCost(c.context.AccessAccount(address))); err != nil {
+			return err
+		}
 	}
-	top.SetUint64(uint64(c.context.GetCodeSize(addr)))
+	top.SetUint64(uint64(c.context.GetCodeSize(address)))
 	return nil
 }
 
 func opExtcodehash(c *context) error {
 	slot := c.stack.peek()
 	address := tosca.Address(slot.Bytes20())
-	if err := gasEip2929AccountCheck(c, address); err != nil {
-		return err
+	if c.isAtLeast(tosca.R09_Berlin) {
+		if err := c.useGas(getAccessCost(c.context.AccessAccount(address))); err != nil {
+			return err
+		}
 	}
 	if !c.context.AccountExists(address) {
 		slot.Clear()
@@ -986,9 +995,11 @@ func opExtCodeCopy(c *context) error {
 		return err
 	}
 
-	addr := tosca.Address(a.Bytes20())
-	if err := gasEip2929AccountCheck(c, addr); err != nil {
-		return err
+	address := tosca.Address(a.Bytes20())
+	if c.isAtLeast(tosca.R09_Berlin) {
+		if err := c.useGas(getAccessCost(c.context.AccessAccount(address))); err != nil {
+			return err
+		}
 	}
 	var uint64CodeOffset uint64
 	if codeOffset.IsUint64() {
@@ -1000,7 +1011,7 @@ func opExtCodeCopy(c *context) error {
 	if err := c.memory.expandMemory(memOffset.Uint64(), length.Uint64(), c); err != nil {
 		return err
 	}
-	codeCopy := getData(c.context.GetCode(addr), uint64CodeOffset, length.Uint64())
+	codeCopy := getData(c.context.GetCode(address), uint64CodeOffset, length.Uint64())
 	return c.memory.Set(memOffset.Uint64(), length.Uint64(), codeCopy)
 }
 
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -249,10 +249,7 @@ func opSstore(c *context) error {
 	cost := tosca.Gas(0)
 	if c.isAtLeast(tosca.R09_Berlin) &&
 		c.context.AccessStorage(c.params.Recipient, key) == tosca.ColdAccess {
-		if err := c.useGas(ColdSloadCostEIP2929); err != nil {
-			// TODO: this error can never be triggered, because if we get here, we already have 2300.
-			return err
-		}
+		cost += 2100
 	}
 
 	storageStatus := c.context.SetStorage(c.params.Recipient, key, value)
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -249,6 +249,7 @@ func opSstore(c *context) error {
 	cost := tosca.Gas(0)
 	if c.isAtLeast(tosca.R09_Berlin) &&
 		c.context.AccessStorage(c.params.Recipient, key) == tosca.ColdAccess {
+		// we do not call useGas here because it is already checked that there is at least 2300 gas available
 		cost += 2100
 	}
 
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -249,7 +249,6 @@ func opSstore(c *context) error {
 	cost := tosca.Gas(0)
 	if c.isAtLeast(tosca.R09_Berlin) &&
 		c.context.AccessStorage(c.params.Recipient, key) == tosca.ColdAccess {
-		// we do not call useGas here because it is already checked that there is at least 2300 gas available
 		cost += 2100
 	}
 
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1199,39 +1199,19 @@ func TestInstructions_EIP2929_dynamicGasCostReportsOutOfGas(t *testing.T) {
 		return accessCost{warm: 100, cold: 2600}
 	})
 
-	tests := map[OpCode]struct {
-		implementation func(*context) error
-	}{
-		BALANCE: {
-			implementation: opBalance,
-		},
-		EXTCODECOPY: {
-			implementation: opExtCodeCopy,
-		},
-		EXTCODEHASH: {
-			implementation: opExtcodehash,
-		},
-		EXTCODESIZE: {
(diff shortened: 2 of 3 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -1147,3 +1147,151 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+					if err != errOutOfGas {
+						t.Errorf("unexpected error: %v", err)
+					}
+				})
+			}
+		}
+	}
+}
+
+func TestInstructions_EIP2929_SSTOREReportsOutOfGasAfterBerlin(t *testing.T) {
```
LuisPH3, line 1260:
```suggestion
func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
```
EIP2929 already says Berlin or later

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1147,3 +1147,151 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+func TestInstructions_EIP2929_staticGasCostForBerlinAndAfterIsZero(t *testing.T) {
+	ops := []OpCode{BALANCE, EXTCODECOPY, EXTCODEHASH, EXTCODESIZE, CALL, CALLCODE, DELEGATECALL, STATICCALL}
+	for _, op := range ops {
+		if getBerlinGasPriceInternal(op) != 0 {
+			t.Errorf("expected zero gas cost for %v", op)
+		}
+	}
+}
+
+func TestInstructions_EIP2929_dynamicGasCostForBerlinAndAfterReportsOutOfGas(t *testing.T) {
```
LuisPH3, line 1160:
```suggestion
func TestInstructions_EIP2929_ReturnsErrorWhenNotEnoughGasForDinamicCost(t *testing.T) {
```
EIP2929 already says Berlin or later

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1147,3 +1147,151 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
 		})
 	}
 }
+
+func TestInstructions_EIP2929_staticGasCostForBerlinAndAfterIsZero(t *testing.T) {
```
LuisPH3, line 1151:
```suggestion
func TestInstructions_EIP2929_StaticGasCostIsZero(t *testing.T) {
```
EIP2929 already says Berlin or later

go/interpreter/lfvm/instructions.go
```diff
@@ -250,7 +250,10 @@ func opSstore(c *context) error {
 	cost := tosca.Gas(0)
 	if c.isAtLeast(tosca.R09_Berlin) &&
 		c.context.AccessStorage(c.params.Recipient, key) == tosca.ColdAccess {
-		cost += 2100
+		if err := c.useGas(ColdSloadCostEIP2929); err != nil {
```
LuisPH3, line 253: The code before the refactor may have been already aware of this issue. 
please revert this lines and document why cost is accumulated but not substracted yet.

LuisPH3 (changes requested):
I like the tests, they seem targeting what we listed.
Nevertheless have a look at the following comments:

LuisPH3 (approved):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,151 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+	}
+}
+
+func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
+	// SSTORE needs to be tested on its own because it demands that at least 2300 gas are available.
+	// Hence we cannot take the same testing approach as for the other operations in EIP-2929.
+
+	// SSTORE demands at least 2300 gas to be available
+	// 2301 gas is not enough to afford tosca.StorageDeleted.
+	for _, availableGas := range []tosca.Gas{2300, 2301} {
```
simonlechner, line 1265: Why two very similar numbers that both fail?

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,151 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+					if err != errOutOfGas {
+						t.Errorf("unexpected error: %v", err)
+					}
+				})
+			}
+		}
+	}
+}
+
+func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
```
simonlechner, line 1259: Since only storage operation `tosca.StorageDeleted` is tested, the name should be changed or also other storage operations should be added to the test.

go/interpreter/lfvm/instructions_test.go
facuMH: because the implementation of sstore checks for  `c.gas <= 2300` hence the first value should exit there, and the second value is big enough to pass that check, but big enough to afford any type of `tosca.StorageStatus`, hence triggering failure of `c.useGas` after `getDynamicCostsForSstore`

go/interpreter/lfvm/instructions_test.go
facuMH: actually comments in the test should be tested, and the return of `SetStorage`, so that we get cheapest dynamic cost. I don't think all storage status cases need to be tested in this test, since also `getDynamicCostsForSstore` is exhaustively tested.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1221,7 +1221,7 @@ func TestInstructions_EIP2929_dynamicGasCostReportsOutOfGas(t *testing.T) {
 					ctxt := context{
 						params: tosca.Parameters{
 							BlockParameters: tosca.BlockParameters{
-								Revision: tosca.R13_Cancun,
+								Revision: revision,
```
LuisPH3, line 1224: :eyes: :open_mouth:

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1261,35 +1261,38 @@ func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
...
-	for _, availableGas := range []tosca.Gas{2300, 2301} {
-		for revision := tosca.R09_Berlin; revision <= newestSupportedRevision; revision++ {
-			for _, access := range []tosca.AccessStatus{tosca.WarmAccess, tosca.ColdAccess} {
-				t.Run(fmt.Sprintf("%v/%v/%v", SSTORE, revision, access), func(t *testing.T) {
-
-					ctxt := context{
-						params: tosca.Parameters{
-							BlockParameters: tosca.BlockParameters{
-								Revision: tosca.R13_Cancun,
+	// 2301 gas is not enough to afford StorageAdded, StorageModified, StorageDeleted.
```
LuisPH3, line 1264: This comment was difficult to read before, the new ordering is not much better:
shall we try something like this?
```go

tests := []tosca.Gas{
    2300, //< not enough for minimum gas requirement
    2301, //< not enough to afford StorageAdded, StorageModified, or StorageDeleted.
}
for _, availableGas := range tests{
```

Is the new for loop with storageStatus really needed?

LuisPH3 (changes requested):
Later changes made comments difficult to read

go/interpreter/lfvm/instructions.go
```diff
@@ -250,6 +250,7 @@ func opSstore(c *context) error {
 	cost := tosca.Gas(0)
 	if c.isAtLeast(tosca.R09_Berlin) &&
 		c.context.AccessStorage(c.params.Recipient, key) == tosca.ColdAccess {
+		// we do not call useGas here because it is already checked that there is at least 2300 gas available
```
HerbertJordan, line 253: This does not really make sense, because `useGas` is more than checking whether there is enough gas. It also consumes gas. The actual reason why `useGas` is not called here is that several cost factors are summed up and `useGas` is called for the aggregated value once, in line 260.
```suggestion
```

go/interpreter/lfvm/instructions.go
```diff
@@ -653,8 +654,10 @@ func opGasPrice(c *context) {
 func opBalance(c *context) error {
 	slot := c.stack.peek()
 	address := tosca.Address(slot.Bytes20())
-	if err := gasEip2929AccountCheck(c, address); err != nil {
-		return err
+	if c.isAtLeast(tosca.R09_Berlin) {
+		if err := c.useGas(getAccessCost(c.context.AccessAccount(address))); err != nil {
```
HerbertJordan, line 658: Maybe for a follow-up: Since these two checks always come in pairs, maybe the revision check should move inside the `getAccessCost` function. In pre-Berlin cases, a cost of `0` would then be returned.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,159 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+				t.Run(fmt.Sprintf("%v/%v/%v", op, revision, access), func(t *testing.T) {
+					ctxt := context{
+						params: tosca.Parameters{
+							BlockParameters: tosca.BlockParameters{
+								Revision: revision,
+							},
+						},
+						stack:  NewStack(),
+						memory: NewMemory(),
+						code:   Code{{op, 0}},
```
HerbertJordan, line 1229: nit: `code` field is irrelevant for the test
```suggestion
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,159 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+					ctxt.gas = accessCosts.warm - 1
+					if access == tosca.ColdAccess {
+						ctxt.gas = accessCosts.cold - 1
+					}
+					mockRunContext := tosca.NewMockRunContext(gomock.NewController(t))
+					if op == SLOAD {
+						mockRunContext.EXPECT().AccessStorage(gomock.Any(), gomock.Any()).Return(access).AnyTimes()
+					} else {
+						mockRunContext.EXPECT().AccessAccount(gomock.Any()).Return(access)
+					}
```
HerbertJordan, line 1243: simpler:
```suggestion
mockRunContext.EXPECT().AccessStorage(gomock.Any(), gomock.Any()).Return(access).AnyTimes()
mockRunContext.EXPECT().AccessAccount(gomock.Any()).Return(access).AnyTimes()
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,159 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+					ctxt.context = mockRunContext
+					for _, v := range test.stack {
+						ctxt.stack.push(uint256.NewInt(v))
+					}
```
HerbertJordan, line 1247: Why is it important to set the address to be accessed to `1` instead of `0`?

This seems to work equally well, but you can eliminate the `stack` specification in the table:
```
					for ctxt.stack.len() < 50 {
						ctxt.stack.push(uint256.NewInt(0))
					}
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,159 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+
+						ctxt := context{
+							params: tosca.Parameters{
+								BlockParameters: tosca.BlockParameters{
+									Revision: revision,
+								},
+							},
+							stack:  NewStack(),
+							memory: NewMemory(),
+							code:   Code{{SSTORE, 0}},
```
HerbertJordan, line 1285: not needed for this test
```suggestion
```

HerbertJordan (approved):

go/interpreter/lfvm/instructions.go
facuMH: while refactoring we decided to not aggregate costs and simply call `c.useGas`. But after that, looking at coverage, we realised of what is explained in this comment. So this is mostly a note to future devs to not try to use `c.useGas` here.

LuisPH3 (approved):

LuisPH3:
part of #690

PR #810, opened 2024-09-26
rustrs: Add rust driver crate

LorenzSchueler:
Add `driver` crate. This crate calls `evmrs` via the C FFI interface. It will be used for integration testing and for benchmarks.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,15 +1,24 @@
+[workspace]
+members = ["driver"]
+
 [package]
 name = "evmrs"
 version = "0.1.0"
 edition = "2021"
 
+[features]
+# workaround for enabling mock feature also in integration tests
+mock = ["dep:mockall"]
+
 [dependencies]
 bnum = "0.11.0"
 evmc-vm = { git = "https://github.com/Fantom-foundation/evmc", branch = "tosca-extensions" }
+mockall = { version = "0.13.0", optional = true }
 sha3 = "0.10.8"
 
 [dev-dependencies]
-mockall = "0.13.0"
+# workaround for enabling mock feature also in integration tests
+evmrs = { path = ".", features = ["mock"] }
 
 [lib]
 crate-type = ["cdylib"]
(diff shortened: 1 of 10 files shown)


LorenzSchueler:
Note: The setup where the dev-dependencies of `evmrs` require `evmrs` is somewhat ugly but it is the only way I could find to enable additional features of the lib for integration tests because for integration tests the lib crate is compiled without the test feature.

rust/Cargo.toml
```diff
@@ -1,15 +1,24 @@
+[workspace]
+members = ["driver"]
+
 [package]
 name = "evmrs"
 version = "0.1.0"
 edition = "2021"
 
+[features]
+# workaround for enabling mock feature also in integration tests
```
LuisPH3, line 10: This comment could be formulated in a more descriptive way, for non-versed people to understand better
```suggestion
# Flag mock enables code generation for test mocking. This allows generating mocks for other build configuration but "test". 
```

rust/Cargo.toml
```diff
@@ -1,15 +1,24 @@
...
+mock = ["dep:mockall"]
+
 [dependencies]
 bnum = "0.11.0"
 evmc-vm = { git = "https://github.com/Fantom-foundation/evmc", branch = "tosca-extensions" }
+mockall = { version = "0.13.0", optional = true }
 sha3 = "0.10.8"
 
 [dev-dependencies]
```
LuisPH3, line 19: integration tests use dev-dependecies? how does this work?

rust/src/types/mock_execution_message.rs
```diff
@@ -20,13 +18,17 @@ pub struct MockExecutionMessage<'a> {
     pub code: Option<&'a [u8]>,
 }
 
+impl<'a> MockExecutionMessage<'a> {
+    pub const DEFAULT_INIT_GAS: u64 = 1_000_000;
```
LuisPH3, line 22: :+1:  
better scoped, yes

LuisPH3 (approved):

rust/Cargo.toml
LorenzSchueler: yes. tests and benchmarks use additionally to the normal dependencies also dev-dependecies.

LuisPH3 (approved):

PR #811, opened 2024-09-26
evmrs: add ffi tests

LorenzSchueler:


LorenzSchueler:
Note: the `#![allow(unused_crate_dependencies)]` is due to the setup where the tests depend on evmrs with feature mock and because the unit tests don't need the driver.

LorenzSchueler:
Not sure how you feel about `#![allow(clippy::undocumented_unsafe_blocks)]` but since this are only tests I think it should be fine.

rust/tests/ffi.rs
```diff
@@ -0,0 +1,184 @@
...
+            &mut [],
+            1,
+        )
+    };
+    assert_eq!(result.status_code, StatusCode::EVMC_SUCCESS);
+    assert_eq!(result.step_status_code, StepStatusCode::EVMC_STEP_STOPPED);
+}
+
+#[test]
+fn run_steppable_with_only_tx_context() {
```
LuisPH3, line 131: What are we testing here? 
please rewrite test name to better reflect property tested:

rust/tests/ffi.rs
```diff
@@ -0,0 +1,184 @@
...
+            &mut [0],
+            1,
+        )
+    };
+    assert_eq!(result.status_code, StatusCode::EVMC_SUCCESS);
+    assert_eq!(result.step_status_code, StepStatusCode::EVMC_STEP_RUNNING);
+}
+
+#[test]
+fn run_steppable_without_values() {
```
LuisPH3, line 103: What are we testing here?

Same with the second one. The property tested here is that the VM stops when reaching the end of code. Not needing values (values? )  is a side effect that does not really need testing.

rust/tests/ffi.rs
```diff
@@ -0,0 +1,184 @@
...
+    let context = ptr::null_mut();
+    let revision = Revision::EVMC_CANCUN;
+    let message = driver::to_evmc_message(&MockExecutionMessage::default());
+    let code = &[Opcode::Add as u8]; // this will error because the stack is empty
+    let result = unsafe { driver::run(&host, context, revision, &message, code) };
+    assert_eq!(result.status_code, StatusCode::EVMC_STACK_UNDERFLOW);
+}
+
+#[test]
+fn run_steppable_with_mock_context() {
```
LuisPH3, line 72: Not sure what property of the VM are we testing here. 

This test overlaps with run_steppable_with_only_tx_context

rust/tests/ffi.rs
```diff
@@ -0,0 +1,184 @@
...
+    let context = ptr::null_mut();
+    let revision = Revision::EVMC_CANCUN;
+    let message = driver::to_evmc_message(&MockExecutionMessage::default());
+    let code = &[Opcode::Push0 as u8];
+    let result = unsafe { driver::run(&host, context, revision, &message, code) };
+    assert_eq!(result.status_code, StatusCode::EVMC_SUCCESS);
+}
+
+#[test]
+fn run_no_code() {
```
LuisPH3, line 48: overlaps with run_steppable_without_values

rust/Cargo.toml
```diff
@@ -19,9 +19,10 @@ sha3 = "0.10.8"
 [dev-dependencies]
 # workaround for enabling mock feature also in integration tests
 evmrs = { path = ".", features = ["mock"] }
+driver = { path = "driver", features = ["mock"] }
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "rlib"]
```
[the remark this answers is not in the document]
LorenzSchueler: `cdylib` is needed to create a dynamic C lib which can be called from Go. The tests need `rlib` because they need type definitions from the main library.

rust/tests/ffi.rs
```diff
@@ -0,0 +1,184 @@
+#![allow(clippy::undocumented_unsafe_blocks, unused_crate_dependencies)]
```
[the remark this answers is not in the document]
LorenzSchueler: This is not possible because we want to be able to pass raw pointers which could be dereferenced. see [clippy](https://rust-lang.github.io/rust-clippy/master/index.html#/not_unsafe_ptr_arg_deref)

rust/tests/ffi.rs
LorenzSchueler: No because run_no_code goes though the non-steppable interface and run_steppable_without_values goes through the steppable interface.

LorenzSchueler:
My main goal with these tests is not to test the correctness of the interpreter. That is (partially) done with unit tests. These tests are testing the ffi module.
here is what they are intended to test:
- run_with_mock_context: test that the non-steppable interface can be called with a mock as context
- run_with_only_tx_context: test that the non-steppable interface can be called with a hard coded get_tx_context function
- run_no_code: test that the non-steppable interface handles empty code correctly
- run_stack_underflow: test that the non-steppable interface handles errors correctly
- run_steppable_with_mock_context: test that the steppable interface can be called with a mock as context
- run_steppable_with_only_tx_context: test that the steppable interface can be called with a hard coded get_tx_context function
- run_steppable_without_values: test that the steppable interface handles empty code, stack, memory and last_call_result_data correctly
- run_steppable_stack_underflow: test that the steppable interface handles errors correctly

rust/tests/ffi.rs
LorenzSchueler: see comment below

rust/tests/ffi.rs
LorenzSchueler: see comment below

rust/tests/ffi.rs
LorenzSchueler: see comment below

LuisPH3:
I am inviting @HerbertJordan here, because I would like to have insights about how we consider tests and what is to be tested where.

rust/tests/ffi.rs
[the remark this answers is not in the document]
LorenzSchueler: I now found a way to make it safe. There are now two wrapper functions. One takes a reference which makes it safe and the other one takes no context and passes a null pointer to the *_raw functions.

HerbertJordan (commented):
Looks okay to me, but I would not say that I understand a lot of the Rust details ...

About coding style, I would argue that the test case names need improvement. A common pattern is
```
<test_target>_<condition>_<expected_behavior>
```
or where there are no conditions, just
```
<test_target>_<property>
```
Examples would be `run_can_be_called_with_mocks_and_callbacks_are_forwarded` instead of `run_with_mock_context`.
Another might be `run_steppable_empty_code_terminates_successfully` instead `run_no_code'.

The snake-case convention of Rust makes separating the parts of the test name a bit harder. Maybe double-underscores could  be used? Something like `run_steppable__empty_code__terminates_successfully`.

In Go, we use a mix of underscores and camel case, like `TestRunSteppable_EmptyCode_TerminatesSuccessfully`.

By using names like this, this table of yours would be implicitly documented:
> ```
> run_with_mock_context: test that the non-steppable interface can be called with a mock as context
> run_with_only_tx_context: test that the non-steppable interface can be called with a hard coded get_tx_context function
> run_no_code: test that the non-steppable interface handles empty code correctly
> run_stack_underflow: test that the non-steppable interface handles errors correctly
> run_steppable_with_mock_context: test that the steppable interface can be called with a mock as context
> run_steppable_with_only_tx_context: test that the steppable interface can be called with a hard coded get_tx_context function
> run_steppable_without_values: test that the steppable interface handles empty code, stack, memory and last_call_result_data correctly
> run_steppable_stack_underflow: test that the steppable interface handles errors correctly
> ```

LuisPH3 (approved):
This PR adds testing for the ffi interface, adds guarantees that data makes the trip through.

PR #812, opened 2024-09-26
LFVM: genericCall errors test

facuMH:
To ensure that each early returned is behaving as expected, we test it.
This PR adds:
- test for `checkSizeOffsetUint64Overflow` all 4 cases.
- removes `checkGas` from `genericCall` and replaces it for calls to `c.useGas` this way we do not need to check for overflow when adding each expense. 
- inlines function `CallGas`  and removes its test.
- adds a test for all the error cases of `genericCall`
- adds test for static context propagation
- adds test for internal call result propagation in the stack

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -1096,23 +1096,25 @@ func genericCall(c *context, kind tosca.CallKind) error {
 		}
 	}
 
-	// The cost of gas was changed during the homestead price change HF.
-	// As part of EIP 150 (TangerineWhistle), the returned gas is gas - base * 63 / 64.
-	cost := tosca.Gas(c.gas - c.gas/64)
-	if provided_gas.IsUint64() && (cost >= tosca.Gas(provided_gas.Uint64())) {
-		cost = tosca.Gas(provided_gas.Uint64())
+	// The Homestead hard-fork introduced a limit on the amount of gas that can be
+	// forwarded to recursive calls. EIP-150 (https://eips.ethereum.org/EIPS/eip-150)
+	// defines that at but one 64th of the available gas in one scope may be passed
+	// to a nested call.
+	endowment := tosca.Gas(c.gas - c.gas/64)
+	if provided_gas.IsUint64() && (endowment >= tosca.Gas(provided_gas.Uint64())) {
+		endowment = tosca.Gas(provided_gas.Uint64())
 	}
-	if err := c.useGas(cost); err != nil {
-		// this usage can never fail because:
-		// - original cost is less than available gas.
-		// - if cost is greater than provided_gas, then provided_gas is also less than available gas
+	if err := c.useGas(endowment); err != nil {
+		// this usage can never fail because, since the endowment is at most
+		// 63/64 of the current gas level.
 		return err
 	}
 
+	nestedCallGas := endowment
 	// first use static and dynamic gas cost and then resize the memory
 	// when out of gas is happening, then mem should not be resized
 	if !value.IsZero() {
-		cost += CallStipend
+		nestedCallGas += CallStipend
 	}
 
 	// Check that the caller has enough balance to transfer the requested value.
@@ -1122,7 +1124,7 @@ func genericCall(c *context, kind tosca.CallKind) error {
 		if balanceU256.Lt(value) {
 			c.stack.pushUndefined().Clear()
 			c.returnData = nil
-			c.gas += cost // the gas send to the nested contract is returned
+			c.gas += nestedCallGas // the gas send to the nested contract is returned
 			return nil
 		}
 	}
@@ -1138,7 +1140,7 @@ func genericCall(c *context, kind tosca.CallKind) error {
 	// Prepare arguments, depending on call kind
 	callParams := tosca.CallParameters{
 		Input: args,
-		Gas:   cost,
+		Gas:   nestedCallGas,
 		Value: tosca.Value(value.Bytes32()),
 	}
 
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -1098,19 +1098,18 @@ func genericCall(c *context, kind tosca.CallKind) error {
 
 	// The Homestead hard-fork introduced a limit on the amount of gas that can be
 	// forwarded to recursive calls. EIP-150 (https://eips.ethereum.org/EIPS/eip-150)
-	// defines that at but one 64th of the available gas in one scope may be passed
+	// defines that at all but one 64th of the available gas in one scope may be passed
 	// to a nested call.
-	endowment := tosca.Gas(c.gas - c.gas/64)
-	if provided_gas.IsUint64() && (endowment >= tosca.Gas(provided_gas.Uint64())) {
-		endowment = tosca.Gas(provided_gas.Uint64())
+	nestedCallGas := tosca.Gas(c.gas - c.gas/64)
+	if provided_gas.IsUint64() && (nestedCallGas >= tosca.Gas(provided_gas.Uint64())) {
+		nestedCallGas = tosca.Gas(provided_gas.Uint64())
 	}
-	if err := c.useGas(endowment); err != nil {
-		// this usage can never fail because, since the endowment is at most
+	if err := c.useGas(nestedCallGas); err != nil {
+		// this usage can never fail because the endowment is at most
 		// 63/64 of the current gas level.
 		return err
 	}
 
-	nestedCallGas := endowment
 	// first use static and dynamic gas cost and then resize the memory
 	// when out of gas is happening, then mem should not be resized
 	if !value.IsZero() {
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -1055,8 +1055,11 @@ func genericCall(c *context, kind tosca.CallKind) error {
 	// and not doing it would be identified by the replay tool as an error.
 	toAddr := tosca.Address(addr.Bytes20())
 
-	if checkSizeOffsetUint64Overflow(inOffset, inSize) != nil ||
-		checkSizeOffsetUint64Overflow(retOffset, retSize) != nil {
+	if checkSizeOffsetUint64Overflow(inOffset, inSize) != nil {
+		return errOverflow
+	}
+
+	if checkSizeOffsetUint64Overflow(retOffset, retSize) != nil {
 		return errOverflow
 	}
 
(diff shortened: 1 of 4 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,103 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckSizeOFfsetUintOverflow_ReturnsAsExpected(t *testing.T) {
```
LuisPH3, line 1150: I think we are missing a case:
`offset.Uint64()+size.Uint64() < offset.Uint64() `
Are we?

LuisPH3 (dismissed):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1146,3 +1146,107 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+		},
+		"return Size overflow": {
+			stack:         []*uint256.Int{u64overflow, zero, zero, zero, zero, zero, zero},
+			expectedError: errOverflow,
+		},
+		"input memory too big": {
+			stack:         []*uint256.Int{zero, zero, uint256.NewInt(maxMemoryExpansionSize + 1), zero, zero, zero, zero},
+			expectedError: errMaxMemoryExpansionSize,
+		},
+		"not enough gas for output memory expansion": {
```
HerbertJordan, line 1206: Should there be a similar test for the input memory expansion?

HerbertJordan (changes requested):

go/interpreter/lfvm/instructions.go
```diff
@@ -1056,38 +1059,39 @@ func genericCall(c *context, kind tosca.CallKind) error {
...
+	// The cost of gas was changed during the homestead price change HF.
+	// As part of EIP 150 (TangerineWhistle), the returned gas is gas - base * 63 / 64.
+	cost := tosca.Gas(c.gas - c.gas/64)
+	if provided_gas.IsUint64() && (cost >= tosca.Gas(provided_gas.Uint64())) {
+		cost = tosca.Gas(provided_gas.Uint64())
+	}
+	if err := c.useGas(cost); err != nil {
+		// this usage can never fail because:
+		// - original cost is less than available gas.
+		// - if cost is greater than provided_gas, then provided_gas is also less than available gas
```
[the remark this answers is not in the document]
facuMH: "available gas" refers to `c.gas`, maybe I should have just used that to refer to it instead of "available"

go/interpreter/lfvm/instructions_test.go
facuMH: since `getSlice` has already been tested I did not want to make an exhaustive list here as well, but rather check that errors can be caught on both calls to `getSlice`, checking one error for input and another for output. If you think all cases should be duplicated for both input/output, I can add it as well.

go/interpreter/lfvm/instructions.go
```diff
@@ -1055,45 +1058,48 @@ func genericCall(c *context, kind tosca.CallKind) error {
...
+		if err := c.useGas(CallNewAccountGas); err != nil {
+			return err
+		}
 	}
 
-	cost := callGas(c.gas, baseGas, provided_gas)
-	if err := c.useGas(baseGas + cost); err != nil {
+	// The Homestead hard-fork introduced a limit on the amount of gas that can be
+	// forwarded to recursive calls. EIP-150 (https://eips.ethereum.org/EIPS/eip-150)
+	// defines that at but one 64th of the available gas in one scope may be passed
```
HerbertJordan, line 1086:
```suggestion
	// defines that at all but one 64th of the available gas in one scope may be passed
```

go/interpreter/lfvm/instructions.go
```diff
@@ -1055,45 +1058,48 @@ func genericCall(c *context, kind tosca.CallKind) error {
...
+	// to a nested call.
+	endowment := tosca.Gas(c.gas - c.gas/64)
+	if provided_gas.IsUint64() && (endowment >= tosca.Gas(provided_gas.Uint64())) {
+		endowment = tosca.Gas(provided_gas.Uint64())
+	}
+	if err := c.useGas(endowment); err != nil {
+		// this usage can never fail because, since the endowment is at most
```
HerbertJordan, line 1093:
```suggestion
		// this usage can never fail because the endowment is at most
```

go/interpreter/lfvm/instructions.go
```diff
@@ -1055,45 +1058,48 @@ func genericCall(c *context, kind tosca.CallKind) error {
...
+		// 63/64 of the current gas level.
 		return err
 	}
 
+	nestedCallGas := endowment
```
HerbertJordan, line 1098: Should this renaming here be removed and let it be called either `nestedCallGas` or `endowment` throughout the function?

go/interpreter/lfvm/instructions_test.go
HerbertJordan: There are four different error branches, all should be reached by at least one of those cases: line 1044, 1048, 1054, and 1058.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1173,3 +1173,137 @@ func TestOpEndWithResult_ReportOverflow(t *testing.T) {
...
+			runContext.EXPECT().AccessAccount(gomock.Any()).Return(tosca.WarmAccess).AnyTimes()
+			runContext.EXPECT().AccountExists(gomock.Any()).Return(false).AnyTimes()
+			runContext.EXPECT().Call(tosca.Call, gomock.Any()).Return(tosca.CallResult{}, nil).AnyTimes()
+
+			ctxt := getEmptyContext()
+			ctxt.context = runContext
+			ctxt.params.Revision = tosca.R13_Cancun
+			ctxt.gas = test.gas
+			// retSize, retOffset, inSize, inOffset, value, address, provided_gas
+			for i := 0; i < 7; i++ {
```
HerbertJordan, line 1267: please simlify

HerbertJordan (changes requested):

go/interpreter/lfvm/instructions_test.go
facuMH: with the test cases as they are at this point, all 4 of those lines are reached.

LuisPH3 (approved):

HerbertJordan (approved):

LuisPH3 (dismissed):

simonlechner (dismissed):

LuisPH3 (approved):

LuisPH3 (changes requested):
I am checking coverage: can we add tests for: 

```
	if c.params.Static && kind == tosca.Call {
		kind = tosca.StaticCall
	}
	```
	and:
	```
		if err != nil || !ret.Success {
		success.Clear()
	} else {
		success.SetOne()
	}
	c.gas += ret.GasLeft
	
	```
	
	
	It makes sense to me that they are part of this PR instead of another one, but if the test needs to be modified dramatically, we can do it in a different test or PR.

File: go/interpreter/lfvm/instructions_test.go [hunk 8]
```diff
@@ -1497,3 +1497,35 @@ func TestGenericCall_ProperlyReportsErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestGenericCall_CallKindPropagatesStaticMode(t *testing.T) {
```
LuisPH3, line 1501: I am afraid this test does not test the property.

Calling CALL in static mode is promoted to StaticCall 
I guess this can be checked through the parameters of the mock. 
Test only needs to test CALL

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1497,3 +1497,35 @@ func TestGenericCall_ProperlyReportsErrors(t *testing.T) {
...
+
+		ctxt := getEmptyContext()
+		ctxt.context = runContext
+		ctxt.params.Static = callKind == tosca.StaticCall
+		ctxt.stack.stackPointer = 7
+		_ = genericCall(&ctxt, tosca.Call)
+	}
+}
+
+func TestGeneralCall_SetsCorrectResultOnStack(t *testing.T) {
```
LuisPH3, line 1514: :+1:

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1497,3 +1497,35 @@ func TestGenericCall_ProperlyReportsErrors(t *testing.T) {
...
+func TestGenericCall_CallKindPropagatesStaticMode(t *testing.T) {
+	for _, callKind := range []tosca.CallKind{tosca.Call, tosca.StaticCall} {
+		runContext := tosca.NewMockRunContext(gomock.NewController(t))
+		runContext.EXPECT().Call(callKind, gomock.Any()).Return(tosca.CallResult{}, nil)
+
+		ctxt := getEmptyContext()
+		ctxt.context = runContext
+		ctxt.params.Static = callKind == tosca.StaticCall
+		ctxt.stack.stackPointer = 7
+		_ = genericCall(&ctxt, tosca.Call)
```
LuisPH3, line 1510: plz checkerror result

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1419,3 +1419,148 @@ func TestGetData(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckSizeOffsetUintOverflow_ReturnsAsExpected(t *testing.T) {
```
LuisPH3, line 1423: Nit: 
The name here does not say much. I understand that these are the tests for the function properties:

An improvement would be to make the function TestCheckSizeOffsetUint64Overflow and define the different properties in a table test:
- IgnoresOffsetWithSizeZero
- ReturnsErrorOnSizeOverflow
- ReturnsErrorOnOffsetOverflow
- ReturnsErrorAfterSumOverflow
- ReturnsOkWhenNoOverflow

(I have no preference for CamelCase or normal sentences, I just wish we could unify this some day)

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1419,3 +1419,148 @@ func TestGetData(t *testing.T) {
...
+	ctxt.context = runContext
+	ctxt.params.Static = true
+	ctxt.stack.stackPointer = 7
+	err := genericCall(&ctxt, tosca.Call)
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+}
+
+func TestGeneralCall_SetsCorrectResultOnStack(t *testing.T) {
```
LuisPH3, line 1549: nit
```suggestion
func TestGeneralCall_ResultIsWrittenToStack(t *testing.T) {
```

LuisPH3 (approved):

go/interpreter/lfvm/instructions_test.go
facuMH: ok will work on the rename

LuisPH3 (approved):

PR #816, opened 2024-09-27
Change jenkins agent

wsodsong:
As an attempt to resolve a waiting time in a queue, this PR changes an agent type which doesn't use the same instance pool and the other regular tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,11 +9,12 @@
 // this software will be governed by the GNU Lesser General Public License v3.
 
 pipeline {
-    agent { label 'quick' }
+    agent { label 'pr' }
 
     options {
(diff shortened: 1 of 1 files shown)


LuisPH3:
Hi @wsodsong,
I noticed this PR, and it comes right in time, we have been having some congestion in the CI lately. 
Please let me know If I can help you with it.

LuisPH3 (approved):
Thanks for the dedicated pool, we can use this change.

LuisPH3:
FYI @facuMH @simonlechner @HerbertJordan 

PRs are going to use dedicated VMs to prevent competition with longer test runs.

simonlechner (approved):

PR #817, opened 2024-09-27
LFVM: Remove use of gas on use gas failure.

LuisPH3:
The normal handling of stop conditions in the VM is to return an error and stop execution. Errors triggered by the VM are equally critical and results on all the gas being consumed. There is no reason to consume the gas for Out of gas scenarios but not for any of the other stopping conditions, therefore this logic is somewhere else and doing it at this point in execution is redundant and error-prone.

The real place where gas gets consumed is [here](https://github.com/Fantom-foundation/Tosca/blob/db1b6de32735886de9f8e7796f5830ddb27f3229/go/interpreter/lfvm/interpreter.go#L115)

### DIFF
--- a/go/interpreter/lfvm/interpreter.go
+++ b/go/interpreter/lfvm/interpreter.go
@@ -57,7 +57,6 @@ type context struct {
 // false otherwise.
 func (c *context) useGas(amount tosca.Gas) error {
 	if c.gas < 0 || amount < 0 || c.gas < amount {
-		c.gas = 0
 		return errOutOfGas
 	}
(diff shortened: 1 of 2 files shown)


LuisPH3:
ct ran successfully.

facuMH (approved):
should maybe add a comment in the actual place where gas is "depleted", since it is done in an "implicit" manner, this can be easily overlooked. 
that aside, nice change

HerbertJordan (approved):
Nice clean-up, thanks!

Issue #758, opened 2024-09-11
Remove internal redundancy of error propagation and status


facuMH:
in memory expansion function status is changed according to the corresponding error, and also an error is returned, callers of this function then have nothing to do with said error since changing status could override the previously assigned status. 
This should be modified so that all errors are only propagated (not modifying context stastus) in internal functions, and only high level functions deal with context status based on propagated errors.

facuMH:
PRs that contribute to this issue are: #795 #797 #790 #786 #783 #779 #772 #775

LuisPH3:
- [ ] fix problem detected by logger changes:  differentiate logic errors / runtime errors  (the valid failure states vs the vm runtime errors)

PR #818, opened 2024-09-27
LFVM: refactor loggerRunner to accept io.Writer instead of printing to stdout. 

facuMH:
We want to add the flexibility of choosing whether logging should be done to a file or stdout/err.
For this we change the `loggerRunner` to take an `io.Writter` and log in there. If no writter is provided then nothing will be logged.

### DIFF
--- a/go/interpreter/lfvm/instruction_logger.go
+++ b/go/interpreter/lfvm/instruction_logger.go
@@ -1,0 +1,51 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package lfvm
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+// loggingRunner is a runner that logs the execution of the contract code to
+// stdout. It is used for debugging purposes.
+type loggingRunner struct {
+	log io.Writer
+}
+
+// NewLoggingRunner creates a new logging runner.
+func NewLoggingRunner(writer io.Writer) loggingRunner {
+	return loggingRunner{log: writer}
+}
+
+func (l loggingRunner) run(c *context) (status, error) {
+	if l.log == nil {
+		l.log = os.Stderr
+	}
+	status := statusRunning
+	var err error
+	for status == statusRunning {
+		// log format: <op>, <gas>, <top-of-stack>\n
+		if int(c.pc) < len(c.code) {
+			top := "-empty-"
+			if c.stack.len() > 0 {
+				top = c.stack.peek().ToBig().String()
+			}
+			l.log.Write([]byte(fmt.Sprintf("%v, %d, %v\n", c.code[c.pc].opcode, c.gas, top)))
+		}
+		status, err = step(c)
+		if err != nil {
+			return status, err
+		}
+	}
+	return status, nil
+}
--- a/go/interpreter/lfvm/instruction_logger.go
+++ b/go/interpreter/lfvm/instruction_logger.go
@@ -41,6 +41,7 @@ func (l loggingRunner) run(c *context) (status, error) {
 			if l.log != nil {
 				_, err = l.log.Write([]byte(fmt.Sprintf("%v, %d, %v\n", c.code[c.pc].opcode, c.gas, top)))
 				if err != nil {
+					// TODO: rework this error to be handled differently than step errors
 					return status, err
 				}
 			}
--- a/go/interpreter/lfvm/instruction_logger.go
+++ b/go/interpreter/lfvm/instruction_logger.go
@@ -16,7 +16,7 @@ import (
 )
 
 // loggingRunner is a runner that logs the execution of the contract code to an io.Writer.
-// If no writter is provided with NewLoggingRunner, nothing will be logged.
+// If the output log is nil, nothing is logged.
 type loggingRunner struct {
 	log io.Writer
 }
--- a/go/interpreter/lfvm/instruction_logger.go
+++ b/go/interpreter/lfvm/instruction_logger.go
@@ -15,9 +15,8 @@ import (
 	"io"
 )
 
-// loggingRunner is a runner that logs the execution of the contract code to
-// an io.Writer. If no writter is provided with NewLoggingRunner, the log will
-// be written to os.Stderr.
+// loggingRunner is a runner that logs the execution of the contract code to an io.Writer.
+// If no writter is provided with NewLoggingRunner, nothing will be logged.
 type loggingRunner struct {
 	log io.Writer
 }
--- a/go/interpreter/lfvm/instruction_logger.go
+++ b/go/interpreter/lfvm/instruction_logger.go
@@ -13,24 +13,22 @@ package lfvm
 import (
 	"fmt"
 	"io"
-	"os"
 )
 
 // loggingRunner is a runner that logs the execution of the contract code to
-// stdout. It is used for debugging purposes.
+// an io.Writer. If no writter is provided with NewLoggingRunner, the log will
+// be written to os.Stderr.
 type loggingRunner struct {
 	log io.Writer
 }
 
-// NewLoggingRunner creates a new logging runner.
-func NewLoggingRunner(writer io.Writer) loggingRunner {
+// newLogger creates a new logging runner that writes to the provided
+// io.Writer.
+func newLogger(writer io.Writer) loggingRunner {
 	return loggingRunner{log: writer}
 }
 
 func (l loggingRunner) run(c *context) (status, error) {
-	if l.log == nil {
-		l.log = os.Stderr
-	}
 	status := statusRunning
 	var err error
 	for status == statusRunning {
@@ -40,7 +38,12 @@ func (l loggingRunner) run(c *context) (status, error) {
 			if c.stack.len() > 0 {
 				top = c.stack.peek().ToBig().String()
 			}
-			l.log.Write([]byte(fmt.Sprintf("%v, %d, %v\n", c.code[c.pc].opcode, c.gas, top)))
+			if l.log != nil {
+				_, err = l.log.Write([]byte(fmt.Sprintf("%v, %d, %v\n", c.code[c.pc].opcode, c.gas, top)))
+				if err != nil {
+					return status, err
+				}
+			}
 		}
 		status, err = step(c)
 		if err != nil {
--- a/go/interpreter/lfvm/instruction_logger_test.go
+++ b/go/interpreter/lfvm/instruction_logger_test.go
@@ -12,6 +12,8 @@ package lfvm
 
 import (
 	"bytes"
+	"io"
+	"os"
 	"strings"
 	"testing"
 
@@ -47,7 +49,7 @@ func TestLogger_ExecutesCodeAndLogs(t *testing.T) {
 			}
 			code := test.code
 			buffer := bytes.NewBuffer([]byte{})
-			logger := NewLoggingRunner(buffer)
+			logger := newLogger(buffer)
 			config := interpreterConfig{
 				runner: logger,
 			}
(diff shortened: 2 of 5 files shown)


go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,111 @@
...
+	os.Stdout = oldOut
+	_ = wErr.Close() // ignore error in test
+	outErr, _ := io.ReadAll(rErr)
+	os.Stderr = oldErr
+
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+
+	if strings.Compare(string(outOut), "") != 0 {
```
LuisPH3, line 105: Nit: len(0) should have been enough

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+
+			if strings.Compare(string(buffer.String()), test.want) != 0 {
+				t.Errorf("unexpected log: want %v, got %v", test.want, buffer.String())
+			}
+		})
+	}
+}
+
+func TestLogger_IfNoWritterIsProvidedStdErrAndStdOutAreNotUsed(t *testing.T) {
```
LuisPH3, line 68: nit: for symetry equivalent test in vanilla is called `TestInterpreter_Vanilla_RunsWithoutOutput`

Do we test both STDERR and STDOUT in vanilla?

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}{
+		"empty": {},
+		"stop": {
+			code: []Instruction{{STOP, 0}},
+			want: "STOP, 10, -empty-\n",
+		},
+		"multiple codes": {
+			code: []Instruction{{PUSH4, 0}, {DATA, 1}, {STOP, 0}},
+			want: "PUSH4, 10, -empty-\nSTOP, 7, 1\n",
+		},
```
LuisPH3, line 37: can we add a test for an error? 
this would add coverage for line 50 in instruction_logger.go and we wont have to add it later on

go/interpreter/lfvm/instruction_logger.go
```diff
@@ -0,0 +1,54 @@
...
+		// log format: <op>, <gas>, <top-of-stack>\n
+		if int(c.pc) < len(c.code) {
+			top := "-empty-"
+			if c.stack.len() > 0 {
+				top = c.stack.peek().ToBig().String()
+			}
+			if l.log != nil {
+				_, err = l.log.Write([]byte(fmt.Sprintf("%v, %d, %v\n", c.code[c.pc].opcode, c.gas, top)))
+				if err != nil {
+					return status, err
```
LuisPH3, line 44: I am conflicted here, and I think we may have a problem because of the error refactoring. 
there is no statusError anymore, and we assume status is an error if there is an error. This is logically correct because all the errors propagated by the vm are runtime errors product of the input. 

The error in this line is not, i.e. this can happen when the filesystem is full. The code calling this error treats the error as an execution violation, sets the result status to Reverted and consumes all the gas. The user will never see that the execution stopped and the error will be lost. 
How can we solve this?

go/interpreter/lfvm/instruction_logger_test.go
facuMH: yes, but I think `strings.Compare(string(outOut), "")` does a better job at showing the intention of what is to be tested.

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,120 @@
...
+				Gas:    3,
+				Code:   []byte{byte(STOP), 0},
+			}
+			code := test.code
+			buffer := bytes.NewBuffer([]byte{})
+			logger := newLogger(buffer)
+			config := interpreterConfig{
+				runner: logger,
+			}
+			_, err := run(config, params, code)
```
HerbertJordan, line 65: should probably call
```suggestion
			_, err := logger.run(params, code)
```
directly

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,120 @@
...
+		Gas:    10,
+		Code:   []byte{byte(STOP), 0},
+	}
+	code := []Instruction{{STOP, 0}}
+
+	// redirect stdout
+	oldOut := os.Stdout
+	rOut, wOut, _ := os.Pipe()
+	os.Stdout = wOut
+
```
HerbertJordan, line 92: nit: it is more error-tolerant if you recover the old `os.Stdout` in a defered call:
```suggestion
     defer func() { os.Stdout = oldOut } ()
```

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,120 @@
...
+	oldErr := os.Stderr
+	rErr, wErr, _ := os.Pipe()
+	os.Stderr = wErr
```
HerbertJordan, line 95: same here

go/interpreter/lfvm/instruction_logger.go
HerbertJordan: @LuisPH3: I agree, those two categories of errors should be separated.

I see two options:
- return three result values: the status, an "error" status which is an enum but not really an error, and an error;
- return two results: the status which is extended with more error states, and an error that is always nil in the vanilla runner but used for this situation here

I think I would prefer the second one, reserving "errors" for real errors, not valid execution results.

LuisPH3: yes, this may require a little work, but it would be more solid. Let me coordinate this work with Facu

go/interpreter/lfvm/instruction_logger_test.go
facuMH: I will take the renaming. Vanilla only tests for stdout.

go/interpreter/lfvm/instruction_logger.go
facuMH: if you agree, I would leave it as is in this PR, and handle `config.run` return in a different PR.

LuisPH3: :+1:  I will start the fix shortly today. 
It is not really in the scope of the logger, therefore I am fine with this for now.

LuisPH3 (approved):

go/interpreter/lfvm/instruction_logger.go
LuisPH3: TBD: #758

LuisPH3 (approved):

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,127 @@
...
+			logger := newLogger(buffer)
+			config := interpreterConfig{
+				runner: logger,
+			}
+			_, err := run(config, params, code)
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+
+			if strings.Compare(string(buffer.String()), test.want) != 0 {
```
simonlechner, line 70: Is the extra string cast needed, since the String method should already return a string.

simonlechner (dismissed):

go/interpreter/lfvm/instruction_logger.go
```diff
@@ -0,0 +1,55 @@
...
+package lfvm
+
+import (
+	"fmt"
+	"io"
+)
+
+// loggingRunner is a runner that logs the execution of the contract code to
+// an io.Writer. If no writter is provided with NewLoggingRunner, the log will
+// be written to os.Stderr.
```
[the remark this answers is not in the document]
facuMH: right this is a comment form a previous implementation, thanks !! will correct it

LuisPH3 (approved):

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			// Get tosca.Parameters
+			params := tosca.Parameters{
+				Input:  []byte{},
+				Static: true,
```
HerbertJordan, line 55: `Input`, `Static`, and `Code` are likely irrelevant for those tests.

HerbertJordan (changes requested):

go/interpreter/lfvm/instruction_logger_test.go
facuMH: code needs to not be empty for early return checks, but input and static are irrelevant.

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,127 @@
...
+		t.Errorf("unexpected error: %v", err)
+	}
+
+	if len(outOut) != 0 {
+		t.Errorf("unexpected stdout: want \"\", got \"%v\"", outOut)
+	}
+	if len(outErr) != 0 {
+		t.Errorf("unexpected stderr: want \"\", got \"%v\"", outErr)
+	}
+}
```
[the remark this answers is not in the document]
facuMH: this test will be added in #823 , because it needs to differentiate between the logical and execution errors to not propagate out of gas same way as out of memory (for example)

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -185,52 +185,54 @@ func TestInterpreter_step_DetectsUpperStackLimitViolation(t *testing.T) {
...
-				mock.EXPECT().SelfDestruct(gomock.Any(), gomock.Any()).AnyTimes()
-				mock.EXPECT().GetTransientStorage(gomock.Any(), gomock.Any()).AnyTimes()
-				mock.EXPECT().SetTransientStorage(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
-
-				ctx := context{
-					params: tosca.Parameters{
-						BlockParameters: tosca.BlockParameters{
-							Revision: revision,
+	for _, runner := range []runner{vanillaRunner{}, loggingRunner{}, &statisticRunner{}} {
+		for _, op := range allOpCodesWhere(isExecutable) {
```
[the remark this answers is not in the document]
facuMH: aahhh, ok, then I'll leave it as it was before.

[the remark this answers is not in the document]
LuisPH3: I did not notice this either. True. Step is enough.

go/interpreter/lfvm/instruction_logger_test.go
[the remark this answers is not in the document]
LuisPH3: :+1:

LuisPH3 (approved):

LuisPH3 (dismissed):
Missing test case and error management are being fixed in follow up PR here: #823

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			// Get tosca.Parameters
+			params := tosca.Parameters{
+				Gas:  3,
+				Code: []byte{byte(STOP), 0},
```
HerbertJordan, line 55: This code here should be irrelevant and be removed because it is inconsistent with the actual code executed.

It seems like it is only needed because of this check here:
https://github.com/Fantom-foundation/Tosca/blob/3ba757ff83347886881383df638ca2143aed6509/go/interpreter/lfvm/interpreter.go#L89

This check should be `if len(code) == 0`, not `if len(params.Code) == 0`. Feel free to update this as part of this PR.

go/interpreter/lfvm/instruction_logger_test.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+		})
+	}
+}
+
+func TestInterpreter_Logger_RunsWithoutOutput(t *testing.T) {
+
+	// Get tosca.Parameters
+	params := tosca.Parameters{
+		Code: []byte{byte(STOP), 0},
```
HerbertJordan, line 79: also here

HerbertJordan (approved):
Looks good, thanks.

LuisPH3 (approved):

PR #819, opened 2024-09-27
LFVM: Test execution of instructions before they are introduced

LuisPH3:
part of #751

Although this test affects the domain of the instructions, it is straight forward to implement it in the interpreter dispatcher, to avoid having to call the appropriate instruction callbacks.

### DIFF
--- a/go/interpreter/lfvm/interpreter_test.go
+++ b/go/interpreter/lfvm/interpreter_test.go
@@ -557,12 +557,12 @@ func TestSteps_FailsWithLessGasThanStaticCost(t *testing.T) {
 
 func TestInterpreter_InstructionsFailWhenExecutedInRevisionsEarlierThanIntroducedIn(t *testing.T) {
 	for _, op := range allOpCodes() {
-		minRev := _introducedIn.get(op)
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/interpreter_test.go
```diff
@@ -555,6 +555,25 @@ func TestSteps_FailsWithLessGasThanStaticCost(t *testing.T) {
 	}
 }
 
+func TestInterpreter_InstructionsFailWhenExecutedInRevisionsEarlierThanIntroducedIn(t *testing.T) {
+	for _, op := range allOpCodes() {
+		minRev := _introducedIn.get(op)
```
facuMH, line 560: nit: `minRevision` is used as the max boundary for the `for` loop in the line below. Is it min revision or max for the loop? maybe `introducedRevision` or some other name would be a little nicer.

facuMH (approved):
just a couple of naming suggestions otherwise looks good

facuMH (approved):

HerbertJordan (approved):

PR #820, opened 2024-09-28
evmrs: make bench script feature based (not commit based)

LorenzSchueler:
This PR changes `bench.sh` such that it benchmarks all interpreters in the same branch/commit but uses different feature flags for evmrs.


rust/scripts/test.sh
LuisPH3, line 1: empty? delete?

rust/scripts/bench.sh
```diff
@@ -1,44 +1,48 @@
...
+    cd rust
+    cargo build --lib --release --features "$FEATURES"
+    cd ..
+
+    BENCH_NAME=$(basename ${GIT_REF})-${INTERPRETER}-${FEATURES}-${RUNS}
+    echo running ${BENCH_NAME}
+    go test ./go/integration_test/interpreter \
+        --run none --bench ^Benchmark[a-zA-Z]+/./${INTERPRETER}$ \
+        --timeout $TIMEOUT --count $RUNS | tee ../${BENCH_DIR}/${BENCH_NAME}
+    sed -i "s/${INTERPRETER}-//g" ../${BENCH_DIR}/${BENCH_NAME}
```
LuisPH3, line 47: what are you doing here?  need to patch the path?

LuisPH3 (approved):

rust/scripts/test.sh
LorenzSchueler: It's not empty. I changed the file permissions so there are no changed lines...

rust/scripts/bench.sh
LorenzSchueler: you mean the `sed ...`? The benchmark names contain the name of the interpreters so in order for benchstat to compare them, they must be renamed so that they all have the same names.

PR #821, opened 2024-09-30
evmrs: add benchmarks

LorenzSchueler:
This PR adds the `benchmarks` which contains benchmarks (currently only `ffi_overhead`) for calling from Rust via FFI into evmrs.

The benchmarks can be used in two ways:
- with criterion to compare different features:
  `cargo bench --package benchmarks --profile profiling [--features <feature1>,<feature2>,...]`
- as a standalone executable for profiling
  `cargo build --package benchmarks --profile profiling [--features <feature1>,<feature2>,...]`
  then use `./target/profiling/benchmarks` with the profiler of your choice

### DIFF
--- a/rust/benchmarks/src/main.rs
+++ b/rust/benchmarks/src/main.rs
@@ -1,0 +1,9 @@
+use benchmarks::RunArgs;
+
+fn main() {
+    let args = RunArgs::ffi_overhead();
+    const ITERATIONS: usize = 200_000_000;
+    for _ in 0..ITERATIONS {
+        benchmarks::run(&args);
+    }
+}
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -12,10 +12,32 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstyle"
 version = "1.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1bec1de6f59aedf83baf9ff929c98f2ad654b97c9510f4e70cf6f661d49fd5b1"
+
+[[package]]
+name = "autocfg"
+version = "1.3.0"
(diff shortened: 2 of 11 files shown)


rust/benchmarks/benches/criterion.rs
LuisPH3, line 1: nit: name this after ffi

rust/benchmarks/src/main.rs
LuisPH3, line 1: what is this?  there are criterion benchmarks. duplicated?

rust/benchmarks/benches/criterion.rs
```diff
@@ -0,0 +1,10 @@
+use benchmarks::RunArgs;
+use criterion::{criterion_group, criterion_main, Criterion};
+
```
LuisPH3, line 3: We have the tradition of writing the command line that executes this program as a comment in the code

LuisPH3 (commented):
Unfortunately, it is still not clear why we need so many benchmark functions. 
Can you write an example of the use of the main function in benchmarks/src/main.rs 
I find difficult to justify different entry points to the program.

rust/benchmarks/src/main.rs
LorenzSchueler: Yes, we call the same benchmarks but this time without criterion to that when we profile this there is no additional overhead that we don't want to measure.

LorenzSchueler:
We need two entry points because with criterion be get statistics about the benchmarks (in particular accurate timing data), but criterion also has some overhead. `main.rs` executes the same benchmark function(s) but without criterion is therefore better suited for profiling e.g. with perf.

rust/benchmarks/benches/criterion.rs
LuisPH3: It feels weird naming your code after the library you use to implement it.
Can we have more than one file in this module?

LuisPH3 (dismissed):

LorenzSchueler:
You could have multiple files but as long as they all use criterion that does not make too much sense because you can already filter the tests within one test binary. The only thing you would gain is that they are compiled as two separate executables. But they would both depend on `benchmarks/src/lib.rs` anyway.

LorenzSchueler:
I renamed the file to `interpreter.rs` because all tests in here will test the whole interpreter and not just parts of it.

LuisPH3 (approved):
:+1:

PR #822, opened 2024-09-30
evmrs: test feature powerset in ci

LorenzSchueler:
This PR runs all rust related ci steps (except for cargo machete which does not support this) with the feature powerset. This in preparation for optimizations which will be put behind features to make sure each combination works.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -46,9 +46,11 @@ jobs:
       uses: dtolnay/rust-toolchain@stable
       with:
         components: clippy
+    - name: install cargo-hack
+      run: cargo install cargo-hack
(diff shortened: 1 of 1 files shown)


LuisPH3 (approved):
note for future PR: 

The library XXX was introduced to solve YYY problem... 
The flag --ZZZ does that one thing... 

 I usually like to use off-the-shelf solutions, but I have to catch up with you here and It will be usefull if you point me to the right places.

LorenzSchueler:
Sure. I would argue that [cargo-hack](https://crates.io/crates/cargo-hack) is the of the shelf solution for this use cases. It basically runs other cargo subcommands with `--each-feature` or each feature set in the `--feature-powerset`.

PR #823, opened 2024-09-30
LFVM Differentiate logic errors from execution errors

LuisPH3:
This is a minimal change that addresses the problem detected in #818. 
The issue being the impossibility to differentiate logical errors (execution violations, program errors) that turn into a failed but still valid execution from runtime errors that are outside from the domain of valid executions (namely the possibility of not being able to write into the file when logging an execution) 

This PR is intentional minimal because there are many helper tools inside the VM using errors to flag execution violations and the error type is the most convenient way to do such. 
alternatives considered were:
- use an enum type for VM logical errors: this would have needed a successful default value and replace all the comparisons `!= nul` with `!= Ok`, Changes would be large, code will be verbose, and future changes could modify states. 
- move error codes into status like evmc does:  this would require writing status and forwarding current status in many places or to do a context status variable again. I consider the latter very error-prone and would like to avoid it.
- Use an error like type which is not error:  nullable to repeat patterns, but incompatible with standard error. This seems quite ad-hoc to me, after all the error type does exactly that. 

The current change defines a boundary function between the VM where only logic errors can happen, and an external interface where status (maybe failure)  or a runtime error can be returned. As long as the VM cannot return a runtime error (current case)  this change keeps the error propagation inside the gas/memory/call routines idiomatic and easy to understand.

### DIFF
--- a/go/interpreter/lfvm/ct.go
+++ b/go/interpreter/lfvm/ct.go
@@ -79,12 +79,8 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 
 	// Run interpreter.
 	status := statusRunning
-	var executionErr error
 	for i := 0; status == statusRunning && i < numSteps; i++ {
-		status, executionErr = step(ctxt)
-		if executionErr != nil {
-			break
-		}
+		status = execute(ctxt, true)
 	}
 
 	// Update the resulting state.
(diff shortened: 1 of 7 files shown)


HerbertJordan (commented):
Nice solution! 

For a full review, I would like to see the PR in sync with the base. Right now, it looks like this PR includes also all the changes of #818

facuMH (approved):

simonlechner (approved):
Nice solution.

facuMH (approved):

HerbertJordan (approved):

PR #824, opened 2024-09-30
LFVM: add test for operations with internal control flow

facuMH:
part of #751 

This test adds coverage for simple instructions with internal control flow. 

Add test for:
- isZero
- less/greater than
- signed less/greater than
- shift right/left
- signed arithmetic right shift

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1624,7 +1624,7 @@ func TestGeneralCall_ResultIsWrittenToStack(t *testing.T) {
 	}
 }
 
-func TestInstructions_ComparingOperations(t *testing.T) {
+func TestInstructions_ComparisonAndShiftOperations(t *testing.T) {
 
 	zero := *uint256.NewInt(0)
 	one := *uint256.NewInt(1)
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -1384,3 +1384,138 @@ func TestInstructions_ConditionalJumpOpsIgnoreDestinationWhenJumpNotTaken(t *tes
...
+
+	zero := *uint256.NewInt(0)
+	one := *uint256.NewInt(1)
+	two := *uint256.NewInt(2)
+	signedMinusOne := *uint256.NewInt(0).Sub(&zero, &one)
+	signedMinusTwo := *uint256.NewInt(0).Sub(&zero, &two)
+	u256 := *uint256.NewInt(256)
+	u257 := *uint256.NewInt(257)
+
+	fillStack := func(values ...uint256.Int) *stack {
```
LuisPH3, line 1398: I like this name

LuisPH3 (approved):
The names of the unit tests the SAR test is confusing because evmcodes does not describe the operation having anything to do with signed/unsigned values. Nevertheless, the code is more illustrative about the behavior, and now I think it makes more sense.

simonlechner (dismissed):
At first sight I thought this test was too long but it is well structured and easy to read. Thank you for the addition.

HerbertJordan (approved):

LuisPH3 (approved):

PR #827, opened 2024-10-01
evmrs: create vm instance only once in driver

LorenzSchueler:
This PR adds `Instance` and `SteppableInstance` in the driver lib and makes all `run*` functions methods of those structs. They both call the respective create function in their constructor (`Default` impl) and the destroy function in their `Drop` impl.

In the benchmark lib `RunArgs` now have a field of type `Instance`. Therefore it is no longer necessary to call the create function in each benchmark iteration but instead only once, which more closely resembles how `evmrs` would be used.

Additionally `evmc_create_steppable_evmrs` and `evmc_create_evmrs` not return mutable pointers which should have been the case all along.

### DIFF
--- a/rust/benchmarks/benches/interpreter.rs
+++ b/rust/benchmarks/benches/interpreter.rs
@@ -6,8 +6,8 @@ use benchmarks::RunArgs;
 use criterion::{criterion_group, criterion_main, Criterion};
 
 fn criterion_benchmark(c: &mut Criterion) {
-    let args = RunArgs::ffi_overhead();
-    c.bench_function("ffi_overhead", |b| b.iter(|| benchmarks::run(&args)));
+    let mut args = RunArgs::ffi_overhead();
+    c.bench_function("ffi_overhead", |b| b.iter(|| benchmarks::run(&mut args)));
 }
 
 criterion_group!(
--- a/rust/benchmarks/src/lib.rs
+++ b/rust/benchmarks/src/lib.rs
@@ -2,7 +2,7 @@
 
 use core::slice;
 
-use driver::{self, get_tx_context_zeroed, host_interface::null_ptr_host_interface};
+use driver::{self, get_tx_context_zeroed, host_interface::null_ptr_host_interface, Instance};
 use evmc_vm::{
     ffi::{evmc_host_interface, evmc_message},
     Revision, StatusCode,
@@ -10,10 +10,11 @@ use evmc_vm::{
 use evmrs::{MockExecutionMessage, Opcode};
 
 pub struct RunArgs {
-    pub host: evmc_host_interface,
-    pub revision: Revision,
-    pub message: evmc_message,
-    pub code: &'static [u8],
+    instance: Instance,
+    host: evmc_host_interface,
+    revision: Revision,
+    message: evmc_message,
+    code: &'static [u8],
 }
 
 impl RunArgs {
@@ -27,6 +28,7 @@ impl RunArgs {
     /// - `code` contains `Opcode::MStore` so that `memory` is non-empty
     /// - `code` returns a single word so that `output` is non-empty
     pub fn ffi_overhead() -> Self {
+        let instance = Instance::default();
         let mut host = null_ptr_host_interface();
         host.get_tx_context = Some(get_tx_context_zeroed);
         let message = MockExecutionMessage {
@@ -35,6 +37,7 @@ impl RunArgs {
         };
 
         Self {
+            instance,
             host,
             revision: Revision::EVMC_CANCUN,
             message: message.to_evmc_message(),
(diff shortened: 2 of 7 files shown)


rust/benchmarks/benches/interpreter.rs
```diff
@@ -6,8 +6,8 @@ use benchmarks::RunArgs;
 use criterion::{criterion_group, criterion_main, Criterion};
 
 fn criterion_benchmark(c: &mut Criterion) {
-    let args = RunArgs::ffi_overhead();
-    c.bench_function("ffi_overhead", |b| b.iter(|| benchmarks::run(&args)));
```
LuisPH3, line 10: Because args is now mutable, best benchmarking would be to have a fresh instance per iteration (truly stateless) 

This is a common pattern when benchmarking, 

```
static context init
Benchmark loop {
      Off timer section: transient state construction
      On Timer secition: the code benchmarked
}


Is this posible using criterion?

File: rust/benchmarks/src/lib.rs [hunk 2]
```diff
@@ -54,12 +57,14 @@ impl RunArgs {
     }
 }
 
-pub fn run(args: &RunArgs) {
+pub fn run(args: &mut RunArgs) {
```
LuisPH3, line 60: Can we experiment with this? 
Maybe another grouping removes the mutability of the arguments to scope it only to the context. 
maybe 2 arguments, or instance being self?  wdyt?

rust/src/ffi/steppable_evmc_vm.rs
```diff
@@ -15,9 +15,9 @@ use crate::{
 };
 
 #[no_mangle]
-extern "C" fn evmc_create_steppable_evmrs() -> *const evmc_vm_steppable {
+extern "C" fn evmc_create_steppable_evmrs() -> *mut evmc_vm_steppable {
     let new_instance = evmc_vm_steppable {
-        vm: evmc_vm::evmc_create_evmrs() as *mut evmc_vm_t,
+        vm: evmc_vm::evmc_create_evmrs(),
```
LuisPH3, line 20: This is definitely one code smell less. Good to see when refactors make code more reasonable.

rust/driver/src/lib.rs
```diff
@@ -42,255 +46,319 @@ pub unsafe extern "C" fn get_tx_context_zeroed(_context: *mut ffi::c_void) -> ev
...
-            if code.is_empty() {
-                ptr::null()
-            } else {
-                code.as_ptr()
-            },
-            code.len(),
-        )
+impl Default for Instance {
+    fn default() -> Self {
+        let instance = unsafe { evmc_create_evmrs() };
```
LuisPH3, line 66: safety?

rust/driver/src/lib.rs
```diff
@@ -42,255 +46,319 @@ pub unsafe extern "C" fn get_tx_context_zeroed(_context: *mut ffi::c_void) -> ev
...
+        if instance.is_null() {
+            panic!("vm instance is null")
```
LuisPH3, line 68: Messages read better when they are in the form:
`failed to do X` (+  optionally `, reason is Y`)
here message would be better: `failed to construct default VM instance`

rust/driver/src/lib.rs
```diff
@@ -42,255 +46,319 @@ pub unsafe extern "C" fn get_tx_context_zeroed(_context: *mut ffi::c_void) -> ev
...
+        stack_len: usize,
+        memory: *mut u8,
+        memory_len: usize,
+        last_call_result_data: *mut u8,
+        last_call_result_data_len: usize,
+        steps: i32,
+    ) -> evmc_step_result {
+        let step_n = self.step_n.unwrap();
+
+        unsafe {
```
LuisPH3, line 230: if the function is unsafe, do we  need to add unsafe inside? 
can we remove one of the two?

rust/driver/src/lib.rs
```diff
@@ -42,255 +46,319 @@ pub unsafe extern "C" fn get_tx_context_zeroed(_context: *mut ffi::c_void) -> ev
...
-            memory.len(),
-            if last_call_result_data.is_empty() {
-                ptr::null_mut()
-            } else {
-                last_call_result_data.as_mut_ptr()
-            },
-            last_call_result_data.len(),
-            steps,
-        )
+impl SteppableInstance {
```
LuisPH3, line 203: The three implementations here seem redundant/too complicated to me, I would try to make an effort to have a single entry point into the VM. The minimum we can do now is document precisely why do they exist.

rust/driver/src/lib.rs
LorenzSchueler: No we don't. I'll remove it.

rust/driver/src/lib.rs
LorenzSchueler: will add a comment

rust/benchmarks/benches/interpreter.rs
LuisPH3: yes, the example presented is a common pattern which achieve two goals: it keeps stateless runs (no carry over data)  but measures time only for the critical path. 

see https://github.com/google/benchmark/blob/main/docs/user_guide.md pause function.

LorenzSchueler: I see. But it seems like criterion has nothing like that.

LuisPH3 (approved):

PR #828, opened 2024-10-01
evmrs: fix&improve ci

LorenzSchueler:
- fix rust-cache action: because there is no `Cargo.toml` in the root it always failed. With an explicit path to the workspace it should now hopefully work.
- add rust-cache action to all jobs
- run clippy also work benchmarks


LuisPH3 (approved):
This pr adds caching to CI jobs to speedup PR builds

PR #826, opened 2024-10-01
evmrs: add BUILD.md

LorenzSchueler:

PR #832, opened 2024-10-01
evmrs: add feature "dump-cov" and bash script to collect CT coverage data

LorenzSchueler:


LorenzSchueler:
I'll wait until #826 is merged because I want to include documentation in BUILD.md for this but maybe @LuisPH3 you can have a look at in in the meantime anyway.

LuisPH3 (changes requested):
When we implemented the C++ coverage, we agreed on adding a test to verify that the infrastructure is working.
The reason is that these mechanisms are rather flimsy and could silently stop working, specially when not used for a while. 

Please take a look on go/lib/cpp/coverage_test.go, and the last stage on the jenkins file. And try to repeat this patern. 
- build again with coverage enabled.
- run the go-test that checks that coverage data has been generated.

LorenzSchueler:
The fix in the last commit is dirty but it was the only way I could find to get it to work.

LuisPH3 (approved):

PR #834, opened 2024-10-02
LFVM: Add test for the empty code bypass.

LuisPH3:
Part of #684 

This PR adds a test for the special case where the contract is an empty code and no interpreter is used. 
Although this seems superfluous code, it is a small and easy optimization that can have an impact in the real block history. 

The test creates a mock to verify that no VM implementation is needed to execute the context in this scenario.


facuMH (dismissed):

go/interpreter/lfvm/interpreter_test.go
```diff
@@ -336,6 +336,23 @@ func TestInterpreter_Vanilla_RunsWithoutOutput(t *testing.T) {
 	}
 }
 
+func TestInterpreter_EmptyCodeBypassesInstanceAndSucceeds(t *testing.T) {
```
HerbertJordan, line 339:
```suggestion
func TestInterpreter_EmptyCodeBypassesRunnerAndSucceeds(t *testing.T) {
```

go/interpreter/lfvm/interpreter_test.go
HerbertJordan, line 1: I think this line here
https://github.com/Fantom-foundation/Tosca/blob/3ba757ff83347886881383df638ca2143aed6509/go/interpreter/lfvm/interpreter.go#L89
should actually check `if len(code) == 0`.

Wdyt?

HerbertJordan (approved):
Thanks!

go/interpreter/lfvm/interpreter_test.go
LuisPH3: yes.

simonlechner (approved):

HerbertJordan (approved):

PR #836, opened 2024-10-03
evmrs: pass feature sets to bench.sh and document usage in BUILD.md

LorenzSchueler:
This PR changes bench.sh such that it accepts a list of feature sets for evmrs to run the Go VM benchmarks and documents how to use this script in BUILD.md

### DIFF
--- a/rust/BUILD.md
+++ b/rust/BUILD.md
@@ -76,9 +76,7 @@ cargo build --features mimalloc,stack-array
 1. Identify a possible optimization opportunity by
     - running the Go VM benchmarks and comparing them, in cases where `evmrs` is slower than the other interpreters
         ```sh
-        # assuming you are in Tosca
-        cd ..
-        ./Tosca/rust/scripts/bench.sh performance
+        ./scripts/bench.sh performance
         ```
     - running a profiler of you choice and identifying a bottleneck
 1. Add a feature in [Cargo.toml](Cargo.toml)
@@ -90,9 +88,7 @@ cargo build --features mimalloc,stack-array
    This will run the Rust benchmarks and generate flamegraphs for all currently implemented features and for all currently implemented features and the new feature
 1. Run Go VM Benchmarks 
     ```sh
-    # assuming you are in Tosca
-    cd ..
-    ./Tosca/rust/scripts/bench.sh performance performance,my-new-feature
+    ./scripts/bench.sh performance performance,my-new-feature
     ```
 1. If all benchmarks indicate that the performance with the optimization is better that before, add the feature name to the features enabled by the `performance` feature in [Cargo.toml](Cargo.toml).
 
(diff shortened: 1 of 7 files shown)


rust/BUILD.md
```diff
@@ -75,6 +75,11 @@ cargo build --features mimalloc,stack-array
 
 1. Identify a possible optimization opportunity by
     - running the Go VM benchmarks and comparing in which cases `evmrs` is slower than the other interpreters
```
LuisPH3, line 77:
```suggestion
    - running the Go VM benchmarks and comparing them, in cases where `evmrs` is slower than the other interpreters
```

rust/BUILD.md
```diff
@@ -75,6 +75,11 @@ cargo build --features mimalloc,stack-array
...
+        ```sh
+        # assuming you are in Tosca
```
LuisPH3, line 79: Simply this, can we run it from the Root directory?

LuisPH3 (changes requested):
If this were my choice: generated files of any kind would be in the build folder. 
Nevertheless, current workflow generates temp files in the project folder and relies on the developer not to commit them. This is a violation of out-of-source-build, but it is the current workflow.
Please do not introduce another workflow where such files end up in a new folder. This is rather unintuitive. We will clean this up, but all at the same time

rust/BUILD.md
LorenzSchueler: I've now changed all scripts s.t. they are expected to be run from Tosca/rust.

LuisPH3: I see, I have to say that I am quite worried about the escalation of scripts and different procedures for the rust project. I understand the use, but I believe most of it is scaffolding needed for early development and not functional parts of the project.
Maybe we should sit down some day and explore what options do we have with the goal of:
- use all the tools we can to assist the task
- prevent the escalation of scaffolding scripts and obscure tools to facilitate reviews and define the boundaries of what is the final product. 

Usually most of the scripting you have been adding here I add it to vscode tasks, I keep private my workflows because they involve temporary files, user folders, and other procedures which other developpers may not use.

LuisPH3 (approved):
We will have to clean this up some day y the future. 
The reason is that I believe that code is code, whenever is rust, go or something scripted. For this reason some guarantees of quality need to be implemented for scripting as well.

LuisPH3:
> All scripts now write their output to Tosca/rust/output. I did not want them to write to rust/target because `cargo clean` should not clean up those outputs. The only exception is coverage.sh which writes into rust/target to make it consistent with cargo llvm-cov.

Why should clean not delete every byproduct of development? I don't know about llvm-cov, but the moment the object files cease to exist, coverage should not be valid anymore.
Byproducts of code compilation/testing/profiling could be really large, and having them all grouped and deleted at the same time makes everyone's life much easier.

LorenzSchueler:
As is mentioned, coverage is the exception which does write into target and will get cleaned up. For the other results I want to keep the artifacts (which are all small) because maybe i need a clean build but still want to compare the benchmark results with what i had before.

PR #837, opened 2024-10-03
LFVM: genericCreate tests 

facuMH:
part of #751 

This PR adds tests for `genericCreate`:
- In all error cases, out of gas due to different checks depending on revision and create kind.
- ensuring result of creation is properly saved in stack

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -844,6 +844,8 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 	gas := c.gas
 	gas -= gas / 64
 	if err := c.useGas(gas); err != nil {
+		// this usage can never fail because the endowment is at most
+		// 63/64 of the current gas level.
 		return err
 	}
 
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -821,11 +821,6 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 		}
 	}
 
-	input, err := c.memory.getSlice(offset.Uint64(), sizeU64, c)
-	if err != nil {
-		return err
-	}
-
 	if !value.IsZero() {
 		balance := c.context.GetBalance(c.params.Recipient)
 		balanceU256 := new(uint256.Int).SetBytes(balance[:])
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -837,7 +837,7 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 		}
 	}
 
-	// Apply EIP150
+	// compute and apply eip150 https://eips.ethereum.org/EIPS/eip-150
 	nestedCallGas := c.gas
 	nestedCallGas -= nestedCallGas / 64
 
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -821,6 +821,11 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 		}
 	}
 
+	input, err := c.memory.getSlice(offset.Uint64(), sizeU64, c)
+	if err != nil {
+		return err
+	}
+
 	if !value.IsZero() {
 		balance := c.context.GetBalance(c.params.Recipient)
 		balanceU256 := new(uint256.Int).SetBytes(balance[:])
@@ -833,19 +838,14 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 	}
 
 	// Apply EIP150
-	gas := c.gas
-	gas -= gas / 64
-	if err := c.useGas(gas); err != nil {
-		// this usage can never fail because the endowment is at most
-		// 63/64 of the current gas level.
-		return err
-	}
+	nestedCallGas := c.gas
+	nestedCallGas -= nestedCallGas / 64
 
 	res, err := c.context.Call(kind, tosca.CallParameters{
 		Sender: c.params.Recipient,
 		Value:  tosca.Value(value.Bytes32()),
 		Input:  input,
-		Gas:    gas,
+		Gas:    nestedCallGas,
 		Salt:   salt,
 	})
 
@@ -862,6 +862,8 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 	} else {
 		c.returnData = nil
 	}
+
+	c.gas -= nestedCallGas
 	c.gas += res.GasLeft
 	c.refund += res.GasRefund
 	return nil
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -781,14 +781,6 @@ func opExtcodehash(c *context) error {
 	return nil
 }
 
-func opCreate(c *context) error {
-	return genericCreate(c, tosca.Create)
-}
-
-func opCreate2(c *context) error {
-	return genericCreate(c, tosca.Create2)
-}
-
 func genericCreate(c *context, kind tosca.CallKind) error {
 
 	// Create is a write instruction, it shall not be executed in static mode.
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -664,7 +664,7 @@ func TestCreateShanghaiInitCodeSize(t *testing.T) {
 		"shanghai-maxuint64-running": {
 			revision:       tosca.R12_Shanghai,
 			init_code_size: math.MaxUint64,
-			expecedErr:     errInitCodeTooLarge,
+			expecedErr:     errOverflow,
 		},
 	}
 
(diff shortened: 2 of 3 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,87 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
 	}
 }
 
+func TestGenericCreate_ReportsErors(t *testing.T) {
```
HerbertJordan, line 1194:
```suggestion
func TestGenericCreate_ReportsErrors(t *testing.T) {
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,87 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+			revision:      tosca.R12_Shanghai,
+			expectedError: errOutOfGas,
+		},
+		"not enough gas for create 2 init code hashing": {
+			offset:        *one,
+			size:          *one,
+			kind:          tosca.Create2,
+			revision:      tosca.R11_Paris,
+			expectedError: errOutOfGas,
+		},
```
HerbertJordan, line 1227: These tests look like they are testing specific properties that got introduced by Shanghai and Paris. It should thus be tested that before those releases some worked and after those not -- or the other way around, as needed.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,87 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+}
+
+func TestGenericCreate_ResultIsWrittenToStack(t *testing.T) {
+	CreatedAddress := tosca.Address([20]byte{19: 0x1})
+	for _, success := range []bool{true, false} {
+		runContext := tosca.NewMockRunContext(gomock.NewController(t))
+		runContext.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{Success: success, CreatedAddress: CreatedAddress}, nil)
+		ctxt := getEmptyContext()
+		ctxt.context = runContext
+		ctxt.stack.stackPointer = 3
```
HerbertJordan, line 1263: Careful: there is no guarantee that the stack will only contain zero-elements. This lead to flaky tests in the past. I think here you depend on all 3 elements to be zero. Thus, you should push 3 zeros on the stack instead.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,87 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+		_ = genericCreate(&ctxt, tosca.Create)
```
HerbertJordan, line 1264: nit: check this error

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,87 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+		want := uint256.NewInt(0)
+		if success {
+			want = uint256.NewInt(1)
+		}
+		if got := ctxt.stack.data[0]; !want.Eq(&got) {
```
HerbertJordan, line 1269: nit: use `ctxt.stack.peek()` or another public function instead of accessing internals.

HerbertJordan (changes requested):

go/interpreter/lfvm/instructions_test.go
facuMH: max init code size has its own test checking that it does not fail before shanghai. I could add it as another case here with expected error nil and unify both tests. As for the create 2 case, it does not depend on revision, this was just a workaround to not have to add a the gas as a test case parameter as well. I'll think it over

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,106 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+			size:          *u64Overflow,
+			expectedError: errOverflow,
+		},
+		"not enough gas for code size": {
+			offset:        *one,
+			size:          *uint256.NewInt(31),
+			revision:      tosca.R12_Shanghai,
+			expectedError: errOutOfGas,
+		},
+		"gas not checked for max code isze before shanghai": {
```
simonlechner, line 1221:
```suggestion
		"gas not checked for max code size before shanghai": {
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,106 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+			offset:        *one,
+			size:          *uint256.NewInt(31),
+			revision:      tosca.R11_Paris,
+			expectedError: nil,
+		},
+		"not enough gas for create 2 init code hashing": {
```
LuisPH3, line 1227:
```suggestion
		"not enough gas for create2 init code hashing": {
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,106 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+			kind:          tosca.Create,
+			expectedError: nil,
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			mockRunContext := tosca.NewMockRunContext(gomock.NewController(t))
+			mockRunContext.EXPECT().Call(gomock.Any(), gomock.Any()).Return(tosca.CallResult{}, nil).AnyTimes()
+			ctxt := context{
```
LuisPH3, line 1245: please use `getEmptyContext` to reduce verbosity

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1191,6 +1191,106 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+		"offset overflow": {
+			offset:        *u64Overflow,
+			size:          *one,
+			expectedError: errOverflow,
+		},
+		"size overflow": {
+			offset:        *one,
+			size:          *u64Overflow,
+			expectedError: errOverflow,
+		},
```
LuisPH3, line 1214: `size + offset` can overflow as well

simonlechner (approved):

LuisPH3 (changes requested):
There are a number of refactors that could be done to genericCreate:
- Remove opCreate and opCreate2 to reduce symbols in the file. 
- Reorder operations to guaratee that state is not modified before gas is consumed:
    - Move getSlice as late as possible, to consume other charges first. 
    - Move balance charge as late as possible, same reason.
    - Leave call and result process at the very end.

go/interpreter/lfvm/instructions.go
```diff
@@ -919,6 +919,8 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 	gas := c.gas
 	gas -= gas / 64
 	if err := c.useGas(gas); err != nil {
+		// this usage can never fail because the endowment is at most
+		// 63/64 of the current gas level.
```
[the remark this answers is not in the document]
facuMH: actually I prefer having the coverage exception (and unreachable code) before not checking for an error. In case anything changes in the future, if the error is ignored, we would never know if something breaks. 

@HerbertJordan what do you think ?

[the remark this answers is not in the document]
LuisPH3: If you two have already discussed this, please ignore my comment and carry on.

LuisPH3 (approved):

go/interpreter/lfvm/instructions.go
```diff
@@ -917,10 +904,17 @@ func genericCreate(c *context, kind tosca.CallKind) error {
 		}
 	}
 
+	input, err := c.memory.getSlice(offset.Uint64(), sizeU64, c)
```
HerbertJordan, line 907: Moving the `getSlice` after the balance check of lines 896-905 might introduce the following question: if create or create2 is called with insufficient balance, should the memory grow to fetch the initialization data and then skip the creation, or should it grow either way.

Before this change, memory grow either way. With this change, if there is not enough balance, memory does not grow.

go/interpreter/lfvm/instructions.go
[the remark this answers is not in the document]
HerbertJordan: Idea: merge this part with the line 942, where the left-over gas of the call is returned (and where encapsulation is broken right now).

Instead subtracting all the gas given to the recursive call initially, and then re-adding whatever is left, just charge the consumed gas once, after the call.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1192,6 +1192,104 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+		offset, size  uint256.Int
+		kind          tosca.CallKind
+		revision      tosca.Revision
+		expectedError error
+	}{
+		"offset overflow": {
+			offset:        *u64Overflow,
+			size:          *one,
+			expectedError: errOverflow,
+		},
```
HerbertJordan, line 1210: Something that should be tested: if the offset is larger the MaxUint64 but the size is 0, everything should be fine -- no error produced.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1192,6 +1192,104 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+		ctxt.stack.push(uint256.NewInt(0))
+		ctxt.stack.push(uint256.NewInt(0))
+		ctxt.stack.push(uint256.NewInt(0))
+		err := genericCreate(&ctxt, tosca.Create)
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		want := uint256.NewInt(0)
+		if success {
+			want = uint256.NewInt(1)
```
HerbertJordan, line 1285: The resulting `want` should be the created address, not 1. In this case, this happens to be the same, but the test code should make this link apparent.

HerbertJordan (changes requested):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1192,6 +1192,104 @@ func TestGenericCreate_MaxInitCodeSizeIsNotCheckedBeforeShanghai(t *testing.T) {
...
+		"size overflow": {
+			offset:        *uint256.NewInt(0),
+			size:          *u64Overflow,
+			expectedError: errOverflow,
+		},
+		"size+offset overflow": {
+			offset:        *uint256.NewInt(math.MaxUint64 - 4),
+			size:          *uint256.NewInt(5),
+			expectedError: errOverflow,
+		},
```
HerbertJordan, line 1220: Some, if not all of these overflow checks may get obsolete once #838 is merged. This PR should be sequenced after #838.

go/interpreter/lfvm/instructions.go
facuMH: yeap, as you mentioned, CT (and hence geth) do expect the memory to grow even if there is not enough balance. So I will move get slice above the balance check.

LuisPH3 (approved):
thanks

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1153,49 +1153,86 @@ func TestComputeCodeSizeCost(t *testing.T) {
...
+			offset:        *one,
+			size:          *uint256.NewInt(31),
+			revision:      tosca.R12_Shanghai,
+			expectedError: errOutOfGas,
+		},
+		"gas not checked for max code size before shanghai": {
+			offset:        *one,
+			size:          *uint256.NewInt(31),
+			revision:      tosca.R11_Paris,
+			expectedError: nil,
```
HerbertJordan, line 1174: nit: maybe `kind` should be specified in both cases since the default value for `tosca.CallKind` is actually `Call`, not `Create`.

HerbertJordan (dismissed):

simonlechner (approved):

PR #838, opened 2024-10-03
LFVM: Refactor overflow checks, add tests

LuisPH3:
part of #751 

This PR removes `checkSizeOffsetUint64Overflow` and rounding done to delay overflow checks
This is done by moving range check for offset and length Op parameters in the memory expansion or getData calls instead of it doing at the beginning of the instructions.
This change simplifies many instructions control flow, and facilitates testing

LFVM has 3 kinds of memory management in the VM, in the sense of manipulating slices that could suffer an out of range access: 
- Memory: where every access Read/Write is preceded by a memory expansion, with a gas check.
- input/code accesses: where memory is read-only and out-of-range scenarios are handled by allocating the requested data and right padding with zeroes.
- result Data: although close related to the latter input buffers, It returns errors on any kind of out-of-range errors because it is accessed to write into it. 

Regarding error reporting, these functions behave as follows: 

- Memory shall return Out of gas for any large access. One could argue that overflows are just a limitation of the architecture, and given a 256 bits addressable space, memory accesses would only produce out of gas errors.
- input/code data accesses are zeroed in any out-of-bounds access, except for having a size value that can not be allocated because it exceeds the maximum addressable space. This is the only error possible. 
- read Data will produce errors whenever, the offset, size, or the sum of both are larger than max uint64. Additionally, will produce an error if the result Data buffer was not allocated large enough for the values computed during execution.

other changes:
- jump overflow errors use now invalid jump to denote overflows in the program counter.


HerbertJordan (commented):
Looks good, and I think it is a clear improvement. Did you encounter any major trade-offs you needed to make during the refactoring?

facuMH (dismissed):

LuisPH3:
> Looks good, and I think it is a clear improvement. Did you encounter any major trade-offs you needed to make during the refactoring?

TL;DR: `getData` avoids checks because it correlates to memory expansions, which are gas-guarded.

There are 2 small details that may require discussion:
- `getData` relies on being called after memory expansion checks: Overflow of the `size` parameter is a non-recoverable error. Reason being that the architecture cannot address 'MaxUint64' + 1 buffer. Adding this error report to `getData` generated a number of non-reachable regions in the op functions. For this reason, the length argument is uin64 in the `getData` function and not uint256 as int `mem.getSlice` which performs the check on size as well. 
- There is an unguarded check that I would have liked to do, but it is not executable: `getData` when `offset` and `size` are smaller than `MaxUint64` but `(offset + size) > MaxUint64`. This scenario should return a right-padded buffer containing the overlap of the original buffer and the offset:offset+size. Nevertheless, the original buffer would require a large amount of memory, which is not allocatable. CT cannot trigger this error neither because the required memory consumption is gas guarded.

HerbertJordan (approved):
Great improvement, thanks!

go/interpreter/lfvm/instructions.go
```diff
@@ -188,44 +175,26 @@ func opMcopy(c *context) error {
...
 
-	destOffset, destOverflow := destAddr.Uint64WithOverflow()
-	srcOffset, srcOverflow := srcAddr.Uint64WithOverflow()
-	if destOverflow || srcOverflow || !sizeU256.IsUint64() {
-		return errOverflow
-	}
-
-	size := sizeU256.Uint64()
-	price := tosca.Gas(3 * tosca.SizeInWords(size))
+	price := tosca.Gas(3 * tosca.SizeInWords(size.Uint64()))
```
facuMH, line 180: this could lead to deducting gas when size is greater than u256, for example if size is [1,1], and gas is >3, this `useGase` call would not fail when it should. I hope the example is clear enough.

go/interpreter/lfvm/instructions.go
```diff
@@ -330,27 +299,8 @@ func opCallDatasize(c *context) {
...
-			return
-		}
-		if pos < len(input) {
-			value[i] = input[pos]
-		} else {
-			value[i] = 0
-		}
-	}
-	top.SetBytes(value[:])
+	value := getData(c.params.Input, top, 32)
```
facuMH, line 302: this could lead to an out of memory panic, this is why `getData` was guarded by `getSlice` before.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1430,82 +1437,81 @@ func TestInstructions_ConditionalJumpOpsIgnoreDestinationWhenJumpNotTaken(t *tes
...
+			size:           2,
+			expectedResult: []byte{0x0, 0x0},
 		},
 	}
 
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
-			if got := checkSizeOffsetUint64Overflow(test.offset, test.size); got != test.expected {
-				t.Errorf("unexpected result, wanted %v, got %v", test.expected, got)
+			_ = test
```
facuMH, line 1511: why is this line needed ?

go/interpreter/lfvm/super_instructions.go
```diff
@@ -85,10 +85,8 @@ func opSwap2_Swap1(c *context) {
 func opDup2_Mstore(c *context) error {
 	var value = c.stack.pop()
 	var addr = c.stack.peek()
```
facuMH, line 87: since you are modifying the functions, is this address or offset ?

facuMH (commented):
I have a couple questions/comments before fully approving

go/interpreter/lfvm/instructions_test.go
LuisPH3: it is not.

go/interpreter/lfvm/memory.go
```diff
@@ -123,25 +123,32 @@ func (m *Memory) length() uint64 {
 // The returned slice is backed by the memory's internal data. Updates to the
 // slice will thus effect the memory states. This connection is invalidated by any
 // subsequent memory operation that may change the size of the memory.
-func (m *Memory) getSlice(offset, size uint64, c *context) ([]byte, error) {
-	err := m.expandMemory(offset, size, c)
+func (m *Memory) getSlice(offset, size *uint256.Int, c *context) ([]byte, error) {
+	if size.IsZero() {
```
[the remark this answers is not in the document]
LuisPH3: It is important to check size first, because in the event of being zero, overflow in offset is allowed.

go/interpreter/lfvm/memory.go
```diff
@@ -123,25 +123,32 @@ func (m *Memory) length() uint64 {
...
-	// since memory does not expand on size 0 independently of the offset,
-	// we need to prevent out of bounds access
-	if size == 0 {
-		return nil, nil
-	}
-	return m.store[offset : offset+size], nil
+	return m.store[offset64 : offset64+size64], nil
 }
 
+var _wordSize = uint256.NewInt(32)
```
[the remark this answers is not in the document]
LuisPH3: The global value prevents a heap allocation on each call. 
It is a premature optimization and Ill delete it.

go/interpreter/lfvm/memory_test.go
```diff
@@ -205,38 +205,40 @@ func TestMemory_expandMemory_expandsMemoryOnlyWhenNeeded(t *testing.T) {
...
-		offset   uint64
-		size     uint64
+		offset   *uint256.Int
+		size     *uint256.Int
 		expected []byte
 	}{
 		"size zero returns empty slice": {
-			offset:   64,
-			size:     0,
+			offset:   uint256.NewInt(64),
```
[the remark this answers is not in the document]
LuisPH3: yes, we may want to go over the semantics of all of these uses.
I don't like neither the uses we do (by value)  vs the uses from geth (by pointer). I would prefer to have only one semantic.

go/interpreter/lfvm/instructions.go
LuisPH3: yes, this code crops overflows and charges only for the first 64 bit component. 
This is an example of error management shadowing other error returns. 
The reason why this is allowed is that `getSlice` is called and will perform the check nevertheless. Maybe it would be clear if getSlice is called before the word count cost.

go/interpreter/lfvm/instructions.go
LuisPH3: This call cannot lead to any panic. Other than the allocator failing to allocate 32 bytes. 
the explanation is in the pr description.

go/interpreter/lfvm/super_instructions.go
LuisPH3: offset

go/interpreter/lfvm/memory_test.go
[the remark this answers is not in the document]
facuMH: I am not sure that is possible since we cannot change how geth does some things, and sometimes we can find more optimal ways of dealing with certain things.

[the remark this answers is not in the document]
LuisPH3: We can stop using them by value. After all our stack API uses them by pointer as well.

facuMH (approved):
thanks for the great changes !!

LuisPH3:
- [ ] ct

facuMH (approved):

PR #842, opened 2024-10-07
Refactor geth adapter

simonlechner:
The geth adapter required some refactoring to enable unit testing.
Further, the revision conversion has been extended up to Cancun and the TODO regarding blobs has been fixed.

### DIFF
--- a/go/geth_adapter/adapter.go
+++ b/go/geth_adapter/adapter.go
@@ -48,37 +48,22 @@ type gethInterpreterAdapter struct {
 }
 
 func (a *gethInterpreterAdapter) Run(contract *geth.Contract, input []byte, readOnly bool) (ret []byte, err error) {
-
 	if adapterDebug {
-		fmt.Printf("Begin of interpreter:\n")
-		fmt.Printf("\tInput:  %v\n", input)
-		fmt.Printf("\tStatic: %v\n", readOnly)
+		debugRunStart(input, readOnly)
 	}
 
+	// Tosca EVM implementations update the refund in the StateDB only at the
+	// end of a contract execution. As a result, it may happen that the refund
+	// becomes temporary negative, since a nested contract may trigger a
+	// refund reduction of some refund earned by an enclosing, yet not finished
+	// contract. However, geth can not handle negative refunds. Thus, we are
+	// shifting the refund base line for a Tosca execution artificially by 2^60
+	// to avoid temporary negative refunds, and eliminate this refund at the
+	// end of the contract execution again.
 	if a.evm.GetDepth() == 0 {
-		// Tosca EVM implementations update the refund in the StateDB only at the
-		// end of a contract execution. As a result, it may happen that the refund
-		// becomes temporary negative, since a nested contract may trigger a
-		// refund reduction of some refund earned by an enclosing, yet not finished
-		// contract. However, geth can not handle negative refunds. Thus, we are
-		// shifting the refund base line for a Tosca execution artificially by 2^60
-		// to avoid temporary negative refunds, and eliminate this refund at the
-		// end of the contract execution again.
-		const refundShift = uint64(1 << 60)
+		const refundShift = 1 << 60
 		a.evm.StateDB.AddRefund(refundShift)
-		defer func() {
-			if err == nil || err == geth.ErrExecutionReverted {
-				// In revert cases the accumulated refund to this point may be negative,
-				// which would cause the subtraction of the original refundShift to
-				// underflow the refund in the StateDB. Thus, the back-shift is capped
-				// by the available refund.
-				shift := refundShift
-				if cur := a.evm.StateDB.GetRefund(); cur < shift {
-					shift = cur
-				}
-				a.evm.StateDB.SubRefund(shift)
-			}
-		}()
+		defer func() { shiftRefundBack(a.evm.StateDB, err, refundShift) }()
 	}
 
 	// The geth EVM infrastructure does not offer means for forwarding read-only
@@ -88,12 +73,7 @@ func (a *gethInterpreterAdapter) Run(contract *geth.Contract, input []byte, read
 	// interpreter). To circumvent this, this adapter encodes the read-only mode
 	// into the highest bit of the gas value (see Call function below). This section
 	// is eliminating this encoded information again.
-	if a.evm.GetDepth() > 0 {
-		readOnly = readOnly || contract.Gas >= (1<<63)
-		if contract.Gas >= (1 << 63) {
-			contract.Gas -= (1 << 63)
-		}
-	}
+	readOnly, contract.Gas = removeReadOnlyInGas(a.evm.GetDepth(), readOnly, contract.Gas)
 
 	// Track the recursive call depth of this Call within a transaction.
 	// A maximum limit of params.CallCreateDepth must be enforced.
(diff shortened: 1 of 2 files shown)


go/geth_adapter/adapter_test.go
```diff
@@ -466,3 +471,212 @@ func TestRunContextAdapter_bigIntToWord(t *testing.T) {
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			gotResult, gotErr := gethToVMErrors(test.input, gas)
+			if !errors.Is(gotErr, test.wantError) {
+				t.Errorf("Unexpected error: expected %v, got %v", test.wantError, gotErr)
+			}
+			if gotResult.Success != test.wantResult.Success {
```
HerbertJordan, line 602: nit: could probably be tested using [`reflect.DeepEqual`](https://pkg.go.dev/reflect#DeepEqual). This would have the additional benefit that if the result struct gets altered in the future for whatever reason, the equality check would cover this.

go/geth_adapter/adapter.go
```diff
@@ -230,13 +194,215 @@ func (a *gethInterpreterAdapter) Run(contract *geth.Contract, input []byte, read
...
+		// by the available refund.
+		shift := refundShift
+		if cur := stateDB.GetRefund(); cur < shift {
+			shift = cur
+		}
+		stateDB.SubRefund(shift)
+	}
+}
+
+func removeReadOnlyInGas(depth int, readOnly bool, gas uint64) (bool, uint64) {
```
HerbertJordan, line 225: nit: it would be nice to group the `setReadOnlyInGas` and `removeReadOnlyInGas` in the file such that they are next to each other;

Also, maybe consider renaming it to `encodeReadOnlyInGas` and `decodeReadOnlyFromGas` ...

go/geth_adapter/adapter.go
```diff
@@ -230,13 +194,215 @@ func (a *gethInterpreterAdapter) Run(contract *geth.Contract, input []byte, read
...
+
+func debugRunEnd(res tosca.Result) {
+	fmt.Printf("End of interpreter:\n")
+	fmt.Printf("\tSuccess:  %v\n", res.Success)
+	fmt.Printf("\tOutput:   %v\n", res.Output)
+	fmt.Printf("\tGas Left: %v\n", res.GasLeft)
+	fmt.Printf("\tRefund:   %v\n", res.GasRefund)
+}
+
+func shiftRefundBack(stateDB geth.StateDB, err error, refundShift uint64) {
```
HerbertJordan, line 211: From the name of this function, one would not expect that it also pays back the refund. One would expect that this function returns the corrected refund.

How about renaming it to `undoRefundShift`?

HerbertJordan (changes requested):
Thanks for this refactoring. Looks good!

The requested changes cover mostly some requests to rename stuff.

go/geth_adapter/adapter_test.go
```diff
@@ -466,3 +471,212 @@ func TestRunContextAdapter_bigIntToWord(t *testing.T) {
...
+
+	tests := map[string]struct {
+		mergeBlock *big.Int
+		block      *big.Int
+		time       uint64
+		want       tosca.Revision
+	}{
+		"Istanbul": {
+			block: istanbulBlock,
+			time:  uint64(0),
```
LuisPH3, line 504: params seem to treat nil time as no fork. I think it would be interesting to check that nil is a valid value

go/geth_adapter/adapter_test.go
```diff
@@ -466,3 +472,201 @@ func TestRunContextAdapter_bigIntToWord(t *testing.T) {
...
+				t.Errorf("Gas was not set or unset correctly, expected %v, got %v", gas, unsetGas)
+			}
+			if gotReadOnly != readOnly {
+				t.Errorf("ReadOnly was not set or unset correctly, expected %v, got %v", readOnly, gotReadOnly)
+			}
+		})
+	}
+}
+
+func TestGethInterpreterAdapter_RefundShiftIsRevertedCorrectly(t *testing.T) {
```
LuisPH3, line 631: nit: `TestGethInterpreterAdapter_RefundShiftIsReverted` 
I am having a crusade against unnecessary adjectives, our code should be correct ;)

go/geth_adapter/adapter.go
```diff
@@ -230,13 +194,215 @@ func (a *gethInterpreterAdapter) Run(contract *geth.Contract, input []byte, read
...
+		vmSalt := &uint256.Int{}
+		vmSalt.SetBytes(parameter.Salt[:])
+		output, newAddr, returnGas, err = a.evm.Create2(a.contract, parameter.Input, gas, parameter.Value.ToUint256(), vmSalt)
+		createdAddress = tosca.Address(newAddr)
+	default:
+		panic(fmt.Sprintf("unsupported call kind: %v", kind))
+	}
+
+	if adapterDebug {
+		fmt.Printf("Result:\n\t%v\n\t%v\n\t%v\n\t%v\n", output, createdAddress, returnGas, err)
```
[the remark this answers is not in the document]
simonlechner: All these fields are printed in the debug call at the end, so I removed this print completely.

go/geth_adapter/adapter_test.go
simonlechner: time is a uint64 in the `ChainConfig` therefore it can not be nil.

LuisPH3:
```go

	chainConfig := &params.ChainConfig{
		ChainID:            big.NewInt(42),
		IstanbulBlock:      istanbulBlock,
		LondonBlock:        londonBlock,
		BerlinBlock:        berlinBlock,
		MergeNetsplitBlock: parisBlock,
		ShanghaiTime:       &shanghaiTime,
		CancunTime:         &cancunTime,
	}
```

I see a pointer

simonlechner: But this is the config of the EVM not the input of the function.
This time is nil in other tests such as `TestRunContextAdapter_Run`.

go/geth_adapter/adapter.go
```diff
@@ -237,6 +248,157 @@ type runContextAdapter struct {
...
+}
+
+func gethToVMErrors(err error, gas tosca.Gas) (tosca.CallResult, error) {
+	switch err {
+	case geth.ErrInsufficientBalance:
+		// In this case, the caller get its gas back.
+		// TODO: this seems to be a geth implementation quirk that got
+		// transferred into the LFVM implementation; this should be fixed.
+		return tosca.CallResult{
+			GasLeft: gas,
```
facuMH, line 342: if I understand this function correctly, it only returns the gas when `geth.ErrInsufficientBalance`, and all other cases it just returns `success: false}, nil`, no ? would it no bee simpler to check for that error to set gas only? I'm thinking ways to reduce the amount of returns (with the same values)

facuMH (dismissed):

LuisPH3 (approved):

HerbertJordan (approved):

go/geth_adapter/adapter.go
simonlechner: This code checks all known error to handle unknown errors differently. Only in the case of an unknown error we return it. I tried a few things during refactoring and I think this is the most readable way.

facuMH (approved):

PR #846, opened 2024-10-07
LFVM: refactor memory expansion to improve testability.

LuisPH3:
Part of #751 

Later refactors in instruction handling made apparent that `expandWithoutChargingGas` could never return an error. 
This PR refactors the process of computing costs and expanding the memory to remove dead code (error handling branches of redundant error checks).

The computation of cost expansion and valid word size are fused into a single function because the two computations are closely related to each other and produce errors with the same inputs. This lead to one of the two functions shadowing the error handling of the other and rendering dead code.

### DIFF
--- a/go/interpreter/lfvm/ct.go
+++ b/go/interpreter/lfvm/ct.go
@@ -203,7 +203,9 @@ func convertCtMemoryToLfvmMemory(memory *st.Memory) *Memory {
 
 	result := NewMemory()
 	size := uint64(len(data))
-	_ = result.expandMemoryWithoutCharging(size)
+	value, validSize, _ := result.getExpansionCostsAndSize(size)
+	result.store = make([]byte, validSize)
+	result.currentMemoryCost = value
 	copy(result.store, data)
 	return result
 }
--- a/go/interpreter/lfvm/ct.go
+++ b/go/interpreter/lfvm/ct.go
@@ -12,6 +12,7 @@ package lfvm
 
 import (
 	"fmt"
+	"math"
 
 	"github.com/Fantom-foundation/Tosca/go/ct"
 	"github.com/Fantom-foundation/Tosca/go/ct/common"
@@ -19,6 +20,7 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/ct/utils"
 	"github.com/Fantom-foundation/Tosca/go/tosca"
 	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/holiman/uint256"
 )
 
 func NewConformanceTestingTarget() ct.Evm {
@@ -58,7 +60,10 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 
 	pcMap := a.getPcMap(state.Code)
 
-	memory := convertCtMemoryToLfvmMemory(state.Memory)
+	memory, err := convertCtMemoryToLfvmMemory(state.Memory)
+	if err != nil {
+		return nil, err
+	}
 
 	// Set up execution context.
 	var ctxt = &context{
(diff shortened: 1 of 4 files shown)


go/interpreter/lfvm/ct.go
```diff
@@ -209,7 +209,9 @@ func convertCtMemoryToLfvmMemory(memory *st.Memory) *Memory {
 
 	result := NewMemory()
 	size := uint64(len(data))
-	_ = result.expandMemoryWithoutCharging(size)
+	value, validSize, _ := result.getExpansionCostsAndSize(size)
+	result.store = make([]byte, validSize)
+	result.currentMemoryCost = value
```
facuMH, line 214: I like this, but I kind of want to add a comment explicitly saying this is done like this because it's an internal conversion from LFVM to CT, and it should not be done anywhere else. Or some kind of warning. Then again, this holds for the whole file. So maybe not necessary

facuMH (approved):

facuMH (approved):

go/interpreter/lfvm/ct.go
LuisPH3: the code has been refactored to call mem.set directly.

go/interpreter/lfvm/memory_test.go
```diff
@@ -28,87 +28,78 @@ func TestMemory_NewMemoryIsEmpty(t *testing.T) {
 	}
 }
 
-func TestGetExpansionCosts(t *testing.T) {
-
-	tests := []struct {
+func TestGetExpansionCostsAndSize(t *testing.T) {
```
HerbertJordan, line 31: All tests seem to assume an initial memory size of 0. There should be a test where the initial size of the memory is something larger than zero, verifying that the expansion costs only cover the difference.

This could be as simple as having two nested loops, one over the initial size, another over the target size and the use the function to make sure that allocating the full size at once has the same costs as to sum up the costs for allocating the memory in two steps. Something like:
```

func TestGetExpansionCostsAndSize_ConsistentIncrementalCharging(t *testing.T) {
	zero := uint256.NewInt(0)
	for a := uint64(0); a < 128; a += 32 {
		for b := uint64(0); b < 128; b += 32 {

			m := NewMemory()

			// Compute costs from 0 to target size.
			costA, sizeA, errA := m.getExpansionCostsAndSize(a)
			costB, sizeB, errB := m.getExpansionCostsAndSize(b)
			if errA != nil || errB != nil {
				t.Fatalf("unexpected error: %v, %v", errA, errB)
			}

			// Compute costs for increasing from size a to size b.
			ctxt := &context{gas: math.MaxInt}
			if err := m.set(zero, make([]byte, a), ctxt); err != nil {
				t.Fatalf("unexpected error: %v", err)
			}
			delta, sizeAB, err := m.getExpansionCostsAndSize(b)
			if err != nil {
				t.Fatalf("unexpected error: %v", err)
			}

			max := sizeA
			if sizeB > sizeA {
				max = sizeB
			}
			if sizeAB != max {
				t.Fatalf("size should be monotonically increasing, got: %d, %d, %d", sizeA, sizeB, sizeAB)
			}

			wantDelta := tosca.Gas(0)
			if a <= b {
				wantDelta = costB - costA
			}

			if wantDelta != delta {
				t.Fatalf("unexpected delta for expansions to %d and %d: want: %d, got: %d", a, b, wantDelta, delta)
			}
		}
	}
}
```
```

HerbertJordan (commented):
There are a few issues that should be handled in a follow-up.

go/interpreter/lfvm/memory_test.go
LuisPH3: I understand the test target, but I do not like that the test relies on utilities which are consumers of the function being tested (set). I will come out with something that test the property.

PR #847, opened 2024-10-07
LFVM: calldataload tests

facuMH:
part of #751 

This PR adds a test for `opCallDataload`. 

One 3 test cases where the input parameter, the offset read from the top of the stack, is bigger than u64, or produces an u64 overflow or is bigger than code length. Another test case checks for the regular behavior of the function.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1827,6 +1827,49 @@ func TestOpBlockhash(t *testing.T) {
 	}
 }
 
+func TestCallDataload_WritesZeroOnStackWhen(t *testing.T) {
+
+	tests := map[string]struct {
+		offset *uint256.Int
+	}{
+		"offset larger than uint64": {
+			offset: new(uint256.Int).Add(uint256.NewInt(2), uint256.NewInt(math.MaxUint64)),
+		},
+		"offset+31 overflows uint64": {
+			offset: uint256.NewInt(math.MaxUint64 - 31),
+		},
+		"offset bigger than input code size": {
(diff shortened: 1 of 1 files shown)


LuisPH3 (dismissed):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1827,6 +1827,48 @@ func TestOpBlockhash(t *testing.T) {
...
+			offset: new(uint256.Int).Add(uint256.NewInt(2), uint256.NewInt(math.MaxUint64)),
+			want:   zero,
+		},
+		"writes zero on the stack when offset+31 overflows uint64": {
+			offset: uint256.NewInt(math.MaxUint64 - 31),
+			want:   zero,
+		},
+		"pads with zeros when offset bigger than input code size": {
+			offset: uint256.NewInt(31),
+			want:   uint256.NewInt(0).SetBytes32([]byte{firstByte, 31: 0x0}),
```
HerbertJordan, line 1851: nit: since `firstByte` is not used anywhere else, and the name is ambiguous (from the point of view of the input data it is actually the last byte), I'd just inline it here and say
```suggestion
			want:   uint256.NewInt(0).SetBytes32([]byte{someData[31], 31: 0x0}),
```

HerbertJordan (approved):
Looks good, thanks!

LuisPH3 (commented):
This looks good, I am going to rebase this with the other PRs and see if we may need to tune something.

LuisPH3:
image

LuisPH3:
The ad-hoc logic for calldataload is now refactored inside of getData, this tests are no longer needed

PR #849, opened 2024-10-08
LFVM: extcodecopy offset overflow case

facuMH:
This PR is part of #751.
This PR adds a test for `opExtCodeCopy` for the specific case where the parameter for `codeOffset` is greater than `maxUint64`, in that case, code offset should be replaced by `maxUint64`, provoking the following `getData` to simply return a slice of size `length` filled with zeros.


LuisPH3:
This PR has been deprecated by #838,
Overflow checks are carried out by getData or expandMemory.

LuisPH3:
from #838 
image

PR #853, opened 2024-10-09
LFVM: Add test for mcopy

LuisPH3:
part of #751 

Old test is replaced with new ones:

- min version is tested for all ops in "TestInterpreter_InstructionsFailWhenExecutedInRevisionsEarlierThanIntroducedIn" 
- memory expansion is tested in memory tests.
- add overlapping range test
- test for errors happening both on read and write. 
- test gas cost for word count.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -13,6 +13,7 @@ package lfvm
 import (
 	"bytes"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"math"
 	"slices"
@@ -2082,8 +2083,12 @@ func TestInstructions_MCopy_ReturnsErrorOnFailure(t *testing.T) {
 				*uint256.NewInt(test.srcOffset),
 				*uint256.NewInt(test.size))
 
-			ctxt.memory.expandMemory(test.destOffset, test.size, &context{gas: 1 << 32})
-			ctxt.memory.expandMemory(test.srcOffset, test.size, &context{gas: 1 << 32})
+			if err := errors.Join(
+				ctxt.memory.expandMemory(test.destOffset, test.size, &context{gas: 1 << 32}),
+				ctxt.memory.expandMemory(test.srcOffset, test.size, &context{gas: 1 << 32}),
+			); err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
 			ctxt.gas = tosca.Gas(3*tosca.SizeInWords(test.size) - test.gasRemoved)
 
 			err := opMcopy(&ctxt)
@@ -2103,14 +2108,17 @@ func TestInstructions_MCopy_CopiesOverlappingRanges(t *testing.T) {
 		*uint256.NewInt(0),
 		*uint256.NewInt(10))
 
-	ctxt.memory.set(
+	err := ctxt.memory.set(
 		uint256.NewInt(0),
 		[]byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9},
 		&context{gas: 1 << 32},
 	)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
 	ctxt.gas = tosca.Gas(3 * tosca.SizeInWords(10))
 
-	err := opMcopy(&ctxt)
+	err = opMcopy(&ctxt)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -13,7 +13,6 @@ package lfvm
 import (
 	"bytes"
 	"crypto/rand"
-	"errors"
 	"fmt"
 	"math"
 	"slices"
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -1879,6 +1755,85 @@ func TestInstructions_ReturnDataCopy_ReturnsErrorOnParameterOverflow(t *testing.
...
+			srcOffset: 128, destOffset: 1024, size: 0,
+		},
+		"returns error when failed write memory expansion": {
+			srcOffset: math.MaxUint64, destOffset: 0, size: 1,
+			expectedError: errOverflow,
+		},
+		"returns error when failed read memory expansion": {
+			srcOffset: 0, destOffset: math.MaxUint64, size: 1,
+			expectedError: errOverflow,
+		},
```
HerbertJordan, line 1775: nit: not sure, but I think the labels are interchanged; the first one should be the read case, the second the write case.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1879,6 +1755,85 @@ func TestInstructions_ReturnDataCopy_ReturnsErrorOnParameterOverflow(t *testing.
...
+
+			err := opMcopy(&ctxt)
+			if err != test.expectedError {
+				t.Fatalf("unexpected error, wanted %v, got %v", test.expectedError, err)
+			}
+		})
+	}
+}
+
+func TestInstructions_MCopy_CopiesOverlappingRanges(t *testing.T) {
```
HerbertJordan, line 1807: nice one!

HerbertJordan (approved):

go/interpreter/lfvm/instructions_test.go
LuisPH3: right

go/interpreter/lfvm/instructions_test.go
```diff
@@ -2036,6 +1912,86 @@ func TestOpExtCodeHash_WritesHashOnStackIfAccountExists(t *testing.T) {
...
 
+func TestInstructions_MCopy_ReturnsErrorOnFailure(t *testing.T) {
+
+	tests := map[string]struct {
+		destOffset, offset, size uint64
+		expectedError            error
+		gasRemoved               uint64
+	}{
+		"does nothing if size is zero": {
+			destOffset: 1024, offset: 125, size: 0,
```
facuMH, line 1923: this test case is only checking that no error is produced, not that "does nothing". Perhaps this case should be its own test.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -2036,6 +1912,86 @@ func TestOpExtCodeHash_WritesHashOnStackIfAccountExists(t *testing.T) {
...
+	}
+	ctxt.gas = tosca.Gas(3 * tosca.SizeInWords(10))
+
+	err = opMcopy(&ctxt)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	expected := []byte{0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9}
+	if want, got := expected, ctxt.memory.store[0:15]; !bytes.Equal(want, got) {
```
facuMH, line 1990: why not check the whole storage slice ?

go/interpreter/lfvm/instructions_test.go
```diff
@@ -2036,6 +1912,86 @@ func TestOpExtCodeHash_WritesHashOnStackIfAccountExists(t *testing.T) {
...
+				*uint256.NewInt(test.offset),
+				*uint256.NewInt(test.size),
+			)
+
+			// ignore memory setup errors, to focus on the mcopy operation
+			// expansion is done to accumulate memory cost and focus on the
+			// word count gas cost.
+			_ = ctxt.memory.expandMemory(test.destOffset, test.size, &context{gas: 1 << 32})
+			_ = ctxt.memory.expandMemory(test.offset, test.size, &context{gas: 1 << 32})
+			ctxt.gas = tosca.Gas(3*tosca.SizeInWords(test.size) - test.gasRemoved)
```
facuMH, line 1955: this is unnecessary. if you set `ctxt.gas` to 6, (for the case of write memory overflow), and the size of the "out of gas" case to 64, the tests still works as intended, but with less code and you can remove `gasRemoved` from tests struct.

facuMH (changes requested):
There are a few things I think can be to done a little better

go/interpreter/lfvm/instructions_test.go
LuisPH3: the memory copied is the one in the range defined. This is needed to keep the test generic, if we add a new test case this code must still be capable of executing both cases.

go/interpreter/lfvm/instructions_test.go
LuisPH3: Same issue as the memory, 
The test code is generic and works with any input test case, if we hardcode gas here two things may happen:
we cannot execute new test cases
we and up triggering out of gas for a different reason than the one the test intends to address. 

This test is modelled to exploit precisely the scenario where gas is less than the size, provided by the parameters of the test.

LuisPH3:
> @facuMH I dont think hard-coding values for a parametric test brings any value. If we hardcode gas or memory contets, there is no point on having test cases in the table.

As discussed offline, 
- test name does not fit pattern. "does-nothing" test gets its own function.
- compare complete memory to test out-of-range access violations.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -2036,6 +1912,115 @@ func TestOpExtCodeHash_WritesHashOnStackIfAccountExists(t *testing.T) {
...
+
+	data := [1024]byte{}
+	_, _ = rand.Read(data[:])
+
+	ctxt := getEmptyContext()
+	ctxt.params.Revision = tosca.R13_Cancun
+	ctxt.stack = fillStack(
+		*uint256.NewInt(2500),
+		*uint256.NewInt(137),
+		*uint256.NewInt(0))
```
HerbertJordan, line 1925: nit:
```suggestion
	ctxt.stack = fillStack(
		*uint256.NewInt(2500),  // destination offset
		*uint256.NewInt(137),    // source offset
		*uint256.NewInt(0),        // size
	)
```

HerbertJordan (approved):

go/interpreter/lfvm/instructions_test.go
LuisPH3: as discussed, new test is written

facuMH (approved):

PR #856, opened 2024-10-09
LFVM: Add test for log operation

LuisPH3:
part of #751 

Add tests for log operation. The test regarding minimum version is done in together with all operations introduced in different revisions.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -15,6 +15,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"math"
+	"slices"
 	"testing"
 
 	"github.com/Fantom-foundation/Tosca/go/tosca"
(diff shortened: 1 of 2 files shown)


facuMH (approved):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1997,6 +1939,69 @@ func TestInstructions_Sha3_WritesCorrectHashInStack(t *testing.T) {
...
+			t.Run(fmt.Sprintf("%v/LOG%d", name, n), func(t *testing.T) {
+
+				runContext := tosca.NewMockRunContext(gomock.NewController(t))
+				ctxt := getEmptyContext()
+				for i := 0; i < n; i++ {
+					ctxt.stack.push(uint256.NewInt(0))
+				}
+				ctxt.stack.push(test.size)
+				ctxt.stack.push(test.offset)
+				ctxt.memory.store = make([]byte, 10)
```
HerbertJordan, line 1973: nit: this actually gets the memory in an invalid state; a memory with data of size 10 should never exist; maybe use `NewMemory` instead?

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1997,6 +1939,69 @@ func TestInstructions_Sha3_WritesCorrectHashInStack(t *testing.T) {
...
+				// this expansion is done to remove expansion costs (if any) and
+				// test word count cost only.
+				_ = ctxt.memory.expandMemory(test.offset.Uint64(), test.size.Uint64(), &context{gas: 1 << 32})
+
+				if test.expectedError == nil {
+					runContext.EXPECT().EmitLog(gomock.Any()).Do(func(log tosca.Log) {
+						if want, got := ctxt.params.Recipient, log.Address; want != got {
+							t.Errorf("unexpected log address, wanted %v, got %v", want, got)
+						}
+						if want, got := n, len(log.Topics); want != got {
```
HerbertJordan, line 1987: If it is not too much effort: it would be great to check that the right topics are forwarded in the correct order.

I guess the simplest would be to always push the topics 5, 4, 3, 2, 1 on the stack and here make sure that the topics 1-n are present;

HerbertJordan (dismissed):

go/interpreter/lfvm/instructions_test.go
LuisPH3: Added check for topic order and data copy.

go/interpreter/lfvm/instructions_test.go
LuisPH3: Using `memory.set` now, 
`set` error is ignored, because if caught, it would shadow the logn error cases.

PR #860, opened 2024-10-09
LFVM: Refactor test to validate transient storage run context calls. 

LuisPH3:
part of #751 

This PR modifies existing tests to validate arguments passed to the run context calls, in context of transactions.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -591,40 +591,48 @@ func TestCreateShanghaiDeploymentCost(t *testing.T) {
 }
 
 func TestTransientStorageOperations(t *testing.T) {
+	address := tosca.Address{}
+	_, _ = rand.Read(address[:])
+	key := tosca.Key{}
+	_, _ = rand.Read(key[:])
+	value := tosca.Word{}
+	_, _ = rand.Read(value[:])
+
 	tests := map[string]struct {
 		op       func(*context) error
 		setup    func(*tosca.MockRunContext)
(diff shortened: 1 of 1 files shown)


HerbertJordan (approved):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -772,40 +772,48 @@ func TestCreateShanghaiDeploymentCost(t *testing.T) {
 }
 
 func TestTransientStorageOperations(t *testing.T) {
+	address := tosca.Address{}
+	_, _ = rand.Read(address[:])
```
facuMH, line 776: this pattern is repeated in multiple tests, should we turn it into a utils fun? or use single explicit values such as `tosca.Address{1}; tosca.Key{2}`? Does it gain anything from being a fully random value rather than just a non zero initialized slice ?

facuMH (approved):

go/interpreter/lfvm/instructions_test.go
LuisPH3: Yes, you are right. I chose this pattern because it was the fastest way to initialize all elements I thought of. 
I find the result much easier to debug, since the values are not 19/31 zeroes. And much faster to compare when printed out.

facuMH: I don't see the problem with turning it into a one liner, I just don't like seeing the same repeated line multiple times per tests, in many test. I agree that this should be used with care and never where it affects the control flow of the test.

PR #861, opened 2024-10-09
LFVM: add test to validate storage runtime context calls arguments. 

LuisPH3:
part of #751 

This pr adds a test to check arguments passed to storage calls.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1247,7 +1247,6 @@ func TestInstructions_StorageOps_CallStorageContext(t *testing.T) {
 				ctxt.params.Recipient = address
 				ctxt.params.Revision = revision
 				ctxt.stack = fillStack(test.stack...)
-				ctxt.gas = tosca.Gas(math.MaxInt64)
 				runContext := tosca.NewMockRunContext(gomock.NewController(t))
 				if revision >= tosca.R09_Berlin {
 					runContext.EXPECT().AccessStorage(address, key).Return(tosca.WarmAccess)
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1212,6 +1212,69 @@ func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
 	}
 }
 
+func TestInstructions_StorageOps_CallStorageContext(t *testing.T) {
+	address := tosca.Address{}
+	_, _ = rand.Read(address[:])
+	key := tosca.Key{}
+	_, _ = rand.Read(key[:])
+	value := tosca.Word{}
+	_, _ = rand.Read(value[:])
+
+	tests := map[OpCode]struct {
+		implementation func(*context) error
+		stack          []uint256.Int
+	}{
+		SLOAD: {
+			implementation: opSload,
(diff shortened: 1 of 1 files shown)


HerbertJordan (dismissed):
nice!

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1380,6 +1380,69 @@ func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
...
+
+	for op, test := range tests {
+		t.Run(op.String(), func(t *testing.T) {
+			forEachRevision(t, op, func(t *testing.T, revision tosca.Revision) {
+
+				ctxt := getEmptyContext()
+				ctxt.params.Recipient = address
+				ctxt.params.Revision = revision
+				ctxt.stack = fillStack(test.stack...)
+				ctxt.gas = tosca.Gas(math.MaxInt64)
```
facuMH, line 1418: does it need to be max gas for some reason ? considering this is this tests is not meant to test gas boundaries, I would rather simply see a big number such as 9999, or the addition of dynamic costs plus static cost.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1380,6 +1380,69 @@ func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
...
+				runContext := tosca.NewMockRunContext(gomock.NewController(t))
+				if revision >= tosca.R09_Berlin {
+					runContext.EXPECT().AccessStorage(address, key).Return(tosca.WarmAccess)
+				}
+				if op == SLOAD {
+					runContext.EXPECT().GetStorage(address, key).Return(value)
+				}
+				if op == SSTORE {
+					runContext.EXPECT().SetStorage(address, key, value)
+				}
```
facuMH, line 1428: adding `value` to stack could also be done here. Or this could be par of the test case parameters. It just seems a little unconsistent to me.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1380,6 +1380,69 @@ func TestInstructions_EIP2929_SSTOREReportsOutOfGas(t *testing.T) {
 	}
 }
 
+func TestInstructions_StorageOps_CallStorageContext(t *testing.T) {
+	address := tosca.Address{}
+	_, _ = rand.Read(address[:])
```
facuMH, line 1385: as I commented in https://github.com/Fantom-foundation/Tosca/pull/860#discussion_r1798857499 
this is a repeated pattern that could be a util func

facuMH (dismissed):
looks good, just a couple of comments

go/interpreter/lfvm/instructions_test.go
LuisPH3: discussed in 860

go/interpreter/lfvm/instructions_test.go
LuisPH3: I prefer to keep stack values in the declaration.
I agree that it is different from what we have been doing before; I find the callback with the mock context pretty verbose: it needs 3 lines in the declaration to declare a lambda, and requires still an if in case that it is left nil. 
This code is not like the other tests, but I find it more straight forward.

facuMH (approved):

PR #862, opened 2024-10-09
LFVM: add test to bring converter coverage back to 100%

LuisPH3:
Part of #751 

Refactor introduced a new branch, this pr adds a test for the situation where converter does not find any SI fit for the input and it falls-back to use normal LFVM instructions.

### DIFF
--- a/.testcoverage.yml
+++ b/.testcoverage.yml
@@ -63,8 +63,12 @@ override:
     path: go/interpreter/lfvm/ct.go
   - threshold: 38
     path: go/interpreter/lfvm/lfvm.go
-  - threshold: 0
+  - threshold: 100
+    path: go/interpreter/lfvm/converter.go
+  - threshold: 100
     path: go/interpreter/lfvm/memory.go
+  - threshold: 100
+    path: go/interpreter/lfvm/interpreter.go
   - threshold: 0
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):

.testcoverage.yml
```diff
@@ -63,8 +63,12 @@ override:
     path: go/interpreter/lfvm/ct.go
   - threshold: 38
     path: go/interpreter/lfvm/lfvm.go
-  - threshold: 0
+  - threshold: 100
+    path: go/interpreter/lfvm/converter.go
+  - threshold: 100
     path: go/interpreter/lfvm/memory.go
+  - threshold: 100
+    path: go/interpreter/lfvm/interpreter.go
```
facuMH, line 71: package coverage should be 100, and here only specify the files that have a lower coverage. wdyt ?

LuisPH3: ok

LuisPH3: I tried that, it seems to be that overrides are exclusive, and I cannot set the lfvm to 100 yet. 
Hopefully we can change this soon to 100% for every file.

facuMH (approved):

PR #863, opened 2024-10-11
LFVM-SI: add test for  od dup2_lt

LuisPH3:
Part of #751

Although SI is not part of the release scope, adding a test for this file sets coverage to 100% for the super-instructions file.
This facilitates threshold for the complete LFVM

### DIFF
--- a/go/interpreter/lfvm/super_instructions_test.go
+++ b/go/interpreter/lfvm/super_instructions_test.go
@@ -1,0 +1,59 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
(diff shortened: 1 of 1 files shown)


go/interpreter/lfvm/super_instructions_test.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func TestSI_opDup2_Lt(t *testing.T) {
+
+	tests := map[string]struct {
+		a, b   uint256.Int
+		result uint256.Int
+	}{
+		"a<b": {
+			a:      *uint256.NewInt(1),
```
facuMH, line 26: nit: use a variable for `*uint256.NewInt(1)` and zero, making it less verbose.

facuMH (approved):

go/interpreter/lfvm/super_instructions_test.go
LuisPH3: I am going to merge this pr now because it is green. Using newInt is not that bad anyway.

PR #864, opened 2024-10-11
Add static overhead benchmark

LorenzSchueler:
This PR adds the `BenchmarkStaticOverhead` benchmark. This benchmark tries to represent the worst case for very short contracts. In particular, it tries to trigger all allocations that happen when using an evmc interpreter, which, depending on how it is implemented, may need to copy data into new allocations.

### DIFF
--- a/go/examples/static_overhead.go
+++ b/go/examples/static_overhead.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package examples
+
+import (
+	"github.com/ethereum/go-ethereum/core/vm"
+)
+
+// This example tries to represent the worst case for very short contracts.
+// In particular, it tries to trigger all allocations that happen when using an
+// evmc interpreter, which, depending on how it is implemented, may need to copy
+// data into new allocations.
+// - code being not empty may cause jump analysis to allocate memory
+// - opcode calldatacopy causes memory expansion
+// - opcode return causes output not to be empty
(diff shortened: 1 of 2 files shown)


go/examples/static_overhead.go
```diff
@@ -0,0 +1,44 @@
...
+// - opcode return causes output not to be empty
+func GetStaticOverheadExample() Example {
+	code := []byte{
+		byte(vm.PUSH1), 4, // push size 4
+		byte(vm.PUSH1), 32, // push offset 32
+		byte(vm.PUSH1), 28, // push destOffset 28
+		byte(vm.CALLDATACOPY), // copy 4 bytes at offset 32 from call data into memory at offset 28
+		byte(vm.PUSH1), 32,    // push len 32
+		byte(vm.PUSH1), 0, // push offset 0
+		byte(vm.RETURN), // return 32 bytes at offset 0
```
LuisPH3, line 32: Does it make sense to add input and return buffers to this example?  These two buffers need to be allocated before the contract execution. Calls to  `ReturnDataCopy` cannot produce other allocation than memory expansion, but it would be interesting to include in case someone implements a lazy allocation scheme.

go/integration_test/interpreter/vm_test.go
```diff
@@ -117,6 +117,12 @@ func BenchmarkEmpty(b *testing.B) {
 	}
 }
 
+func BenchmarkStaticOverhead(b *testing.B) {
+	b.Run("1", func(b *testing.B) {
```
LuisPH3, line 121: this line is superfluous.
this line is used in other benchmarks to change some parameter value and name the benchmark after the value. There are no parameters in this  version

LuisPH3 (changes requested):

go/examples/static_overhead.go
LuisPH3: as seen, RETURN writes into return data as well.

LuisPH3 (approved):

simonlechner (approved):
lgtm

PR #866, opened 2024-10-11
evmrs: create functions for each opcode impl

LorenzSchueler:
This PR extracts all opcode implementations into functions. This in in preparation for other function dispatch approaches, in particular to be able to use a jump table.

### DIFF
--- a/rust/src/interpreter.rs
+++ b/rust/src/interpreter.rs
@@ -14,6 +14,12 @@ use crate::{
     utils::{check_min_revision, check_not_read_only, word_size, Gas, SliceExt},
 };
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ShouldStop {
+    True,
+    False,
+}
+
 pub struct Interpreter<'a, E>
 where
     E: ExecutionContextTrait,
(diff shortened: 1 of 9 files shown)


rust/src/interpreter.rs
```diff
@@ -14,6 +14,12 @@ use crate::{
     utils::{check_min_revision, check_not_read_only, word_size, Gas, SliceExt},
 };
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ShouldStop {
```
LuisPH3, line 18: I am not convinced by this solution.
First, I think reengineering a built-in type like Bool looks a like a code smell to me. 
Have a look at this function:

```rust
    fn stop(&mut self) -> Result<ShouldStop, StatusCode> {
        self.step_status_code = StepStatusCode::EVMC_STEP_STOPPED;
        self.status_code = StatusCode::EVMC_SUCCESS;
        Ok(ShouldStop::True)
    }
```

Status code is returned and it is writen into the global state. This is redundant (even more if we consider that the function itself is private and the result of the call can only be consumed by self itself) . 

I would consider refactoring this change to prevent these pattern, maybe using something like: 
```rust
 self.status_code = match op{ 
 [...]
 }
 if self.status != Running{
     return Ok(...);
 }
 ```
Maybe there is even a less code option, that avoids having to return a state from every instruction. 
```rust
    opXXXXX()?.map() // where map turns result<(),err>.unwrap() into status.running
```

LuisPH3 (changes requested):
I see the need for this change and it is overall pretty good. Nevertheless, there is a point where I think we can improve.

LorenzSchueler:
passed CT (at least before the rebase)

LuisPH3 (approved):

PR #868, opened 2024-10-14
LFVM: Refactor interpreter Configuration, test registration of experimental and default features.

LuisPH3:
part of #684 

This PR refactors lfvm.go to clean up the public interface of the lfvm package

There were three different configuration structs found, in the package, in the interpreter and in the step function. The three objects have been merged into 2 objects: 
- `configuration` is an opaque object that allows configuring the interpreter. Because the default value may not suffice, a function `DefaultConfiguration` is created with the features planned for the release. 
- `configurationOverrides` is a key-value map that can set a subset of configuration flags, final set of configs is TBD. 

The public interface of LFVM is: 
- type `Configuration`
     - Constructor `DefaultConfiguration` is the sanctioned features for the release.
     - type `ConfigurationOverrides` is created because the unused any type passed to the factory, this type can be rewritten in any way. 
- `Interpreter` is the old lfvm instance. It is public to allow tests to query configuration parameters from outside the package.
     - `NewInterpreter` is the only constructor
- additionally. The default configured lfvm instance is pushed into the global registry. 
- `RegisterExperimentalInterpreterConfigurations` can be called to register more configurations     
     
To prevent polluting the registry during testing, registry tests use the lfvm_test package.

### DIFF
--- a/go/interpreter/lfvm/lfvm.go
+++ b/go/interpreter/lfvm/lfvm.go
@@ -45,6 +45,9 @@ func init() {
 // function should not be called in production code, as the resulting VMs are
 // not officially supported.
 func RegisterExperimentalInterpreterConfigurations() error {
+
+	configs := map[string]config{}
+
 	for _, si := range []string{"", "-si"} {
 		for _, shaCache := range []string{"", "-no-sha-cache"} {
 			for _, mode := range []string{"", "-stats", "-logging"} {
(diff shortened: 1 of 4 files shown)


facuMH (commented):
nice refactor, I have some questions though

LuisPH3:
refactor will not be done, instead the file will be tested as it is now.

PR #870, opened 2024-10-15
LFVM: Rest register experimental features into global registry

LuisPH3:
part of #684 

This PR adds testing for lfvm.go, in particular to registration of the experimental configurations of the vm and error handling during execution or construction of the vm.

### DIFF
--- a/go/interpreter/lfvm/lfvm_interface_test.go
+++ b/go/interpreter/lfvm/lfvm_interface_test.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package lfvm_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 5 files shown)


HerbertJordan (approved):
Very nice, thanks!

go/interpreter/lfvm/lfvm_interface_test.go
```diff
@@ -0,0 +1,72 @@
...
+			params.Gas = 5
+			res, err := vm.Run(params)
+			if err != nil {
+				t.Fatalf("failed to run interpreter: %v", err)
+			}
+
+			if want, got := true, res.Success; want != got {
+				t.Fatalf("unexpected success result: want %v, got %v", want, got)
+			}
+			if want, got := tosca.Gas(0), res.GasLeft; want != got {
```
simonlechner, line 67: Nit: Since there are many cases in which the VM sets the remaining gas to 0, the gas could be increased to 6 and checked that 1 is still remaining.

simonlechner (approved):
Very nice!

go/interpreter/lfvm/lfvm_interface_test.go
LuisPH3: There should not be such cases anymore. The VM never needs to do anything with gas other than consume the required fees. I see the point of your concern, but such errors should be delivered now via the err return.

simonlechner (approved):

PR #872, opened 2024-10-15
evmrs: add feature "mimalloc"

LorenzSchueler:
This PR adds the feature "mimalloc". When enabled the default memory allocator is replaced by mimalloc.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -28,3 +28,10 @@ pub unsafe extern "C" fn evmrs_dump_coverage() {
 pub extern "C" fn evmrs_is_coverage_enabled() -> u8 {
     cfg!(feature = "dump-cov") as u8
 }
+
+#[cfg(feature = "mimalloc")]
+use mimalloc::MiMalloc;
+
+#[cfg(feature = "mimalloc")]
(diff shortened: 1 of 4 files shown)


rust/src/lib.rs
```diff
@@ -28,3 +28,10 @@ pub unsafe extern "C" fn evmrs_dump_coverage() {
 pub extern "C" fn evmrs_is_coverage_enabled() -> u8 {
     cfg!(feature = "dump-cov") as u8
 }
+
+#[cfg(feature = "mimalloc")]
+use mimalloc::MiMalloc;
+
+#[cfg(feature = "mimalloc")]
+#[global_allocator]
+static GLOBAL: MiMalloc = MiMalloc;
```
LuisPH3, line 37: is this it? is this everything you need to change the system allocator?

How does it work?  is global a "keyword" symbol?

LuisPH3 (approved):

rust/src/lib.rs
LorenzSchueler: `#[global_allocator]` is a compiler build-in which registers a static as a global allocator. The name of the static does not matter.

PR #873, opened 2024-10-15
evmrs: add feature "stack-array"

LorenzSchueler:
This PR adds the feature "stack-array". When this feature is enabled the Stack holds a [MaybeUninit<u256>; 1024] instead of Vec<u256>.

### DIFF
--- a/rust/src/types/stack.rs
+++ b/rust/src/types/stack.rs
@@ -51,15 +51,16 @@ impl StackRaw {
         self.len
     }
 
+    /// # Safety
+    /// self.len must be < 1024
     pub unsafe fn push(&mut self, value: u256) {
-        *self.data.get_unchecked_mut(self.len) = MaybeUninit::new(value);
+        self.data[self.len] = MaybeUninit::new(value);
         self.len += 1;
     }
 
-    pub fn pop_array<const N: usize>(&mut self) -> Option<[u256; N]> {
-        if self.len < N {
-            return None;
-        }
+    /// # Safety
+    /// self.len must be >= N
+    pub unsafe fn pop_array<const N: usize>(&mut self) -> [u256; N] {
         self.len -= N;
         let start = self.data.as_ptr() as *const u256;
         // SAFETY:
@@ -70,7 +71,7 @@ impl StackRaw {
         // `self.len` just got decremented by N, which means now that the first `self.len + N`
         // elements are initialized. Therefore, it is safe to read N elements starting at index
         // `self.len` as an array of length N and type u256.
-        unsafe { Some(std::ptr::read(array_start as *const [u256; N])) }
+        unsafe { std::ptr::read(array_start as *const [u256; N]) }
     }
 }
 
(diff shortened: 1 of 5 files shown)


rust/src/types/stack.rs
```diff
@@ -1,19 +1,112 @@
...
+                &mut self.data[..self.len],
+            )
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.len
+    }
+
+    pub unsafe fn push(&mut self, value: u256) {
```
LuisPH3, line 54: I think you can limit the use of unsafe to this object, and do the range check, like you do with pop.

rust/src/types/stack.rs
```diff
@@ -1,19 +1,112 @@
...
+        *self.data.get_unchecked_mut(self.len) = MaybeUninit::new(value);
+        self.len += 1;
+    }
+
+    pub fn pop_array<const N: usize>(&mut self) -> Option<[u256; N]> {
+        if self.len < N {
```
LuisPH3, line 60: dead code?  check overflow should prevent this.

rust/src/types/stack.rs
```diff
@@ -1,19 +1,112 @@
...
+    pub fn as_slice(&self) -> &[u256] {
+        self.0.as_slice()
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
     }
 
     pub fn push(&mut self, value: impl Into<u256>) -> Result<(), FailStatus> {
         self.check_overflow_on_push()?;
```
LuisPH3, line 102: if we move the overflow check inside of push (like pop does) 
we can do the overflow check only for the vector case.  and remove the unsafe block from this scope.

rust/src/types/stack.rs
```diff
@@ -22,23 +115,30 @@ impl Stack {
...
         self.check_underflow(N)?;
 
-        let new_len = self.0.len() - N;
-        let mut array = [u256::ZERO; N];
-        array.copy_from_slice(&self.0[new_len..]);
-        self.0.truncate(new_len);
-        Ok(array)
+        #[cfg(feature = "stack-array")]
+        {
+            Ok(self.0.pop_array().unwrap())
```
LuisPH3, line 127: As mentioned before, there is a duplicated bound check here.
Maybe is easier to write: `return self.0.pop_array()` return a result and do the bounds check in the alternative implementation. 
same for Push

LuisPH3 (changes requested):
This is a good example of low level mem management in rust. I would like to reduce a little more the scope of the unsafe regions, tough.

rust/src/types/stack.rs
LorenzSchueler: Yes but without feature "stack-array" I have to do the range check in Stack::pop. When the feature is enabled I could do it in StackRaw::pop but that would mean that it is done in two different places depending on the feature.

rust/src/types/stack.rs
LorenzSchueler: Yes. I think pop_array should be unsafe and assume that the check was already done and then return [u256; N] directly.

rust/src/types/stack.rs
LorenzSchueler: See earlier comment. I don't like the idea of doing the check in different places.

LorenzSchueler:
One other idea: what do you think about renaming StackRaw to Stack and using it directly instead of wrapped in a Stack. That would solve all of those ^ issues.

LuisPH3 (approved):

PR #874, opened 2024-10-16
evmrs: add feature "custom-evmc"

LorenzSchueler:
This PR adds the feature "custom-evmc". When this feature is enabled a different version of evmc-vm is used, in which multiple allocations are avoided. 
To simplify the imports, evmrs (lib) reexports evmc_vm.

Important changes:
- evmrs reexports evmc_vm
- because ExecutionMessage now holds slices instead of Vec it has a lifetime
- output is now Option<Box<[u8]>> instead of Option<Vec<u8>>
- Interpreter -> ExecutionResult conversion now need a &mut because it moves the output field out of it leaving None behind

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -35,7 +35,6 @@ version = "0.1.0"
 dependencies = [
  "criterion",
  "driver",
- "evmc-vm",
  "evmrs",
 ]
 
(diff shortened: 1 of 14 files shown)


LuisPH3:
I need a little more explanation about the changes to the references, where do `&mut` come from?  can you point the feature that requires this?

LuisPH3 (approved):

PR #876, opened 2024-10-16
evmrs: add feature "jumptable"

LorenzSchueler:
This PR adds the feature "jumptable". When enabled, opcode dispatch is done using a jumptable instead of a switch.

Changes:
- add type alias OpResult and OpFn
- remove match in Interpreter::run and call Interpreter::run_op instead
- add two Interpreter::run_op impls, one using switch and one using the jumptable

### DIFF
--- a/rust/src/interpreter.rs
+++ b/rust/src/interpreter.rs
@@ -14,6 +14,11 @@ use crate::{
     utils::{check_min_revision, check_not_read_only, word_size, Gas, SliceExt},
 };
 
+type OpResult = Result<(), FailStatus>;
+
+#[cfg(feature = "jumptable")]
+type OpFn<'a, E> = fn(&mut Interpreter<'a, E>) -> OpResult;
+
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


rust/src/interpreter.rs
```diff
@@ -278,75 +133,500 @@ where
...
+            Opcode::StaticCall => self.static_call(),
+            Opcode::Revert => self.revert(),
+            Opcode::Invalid => self.invalid(),
+            Opcode::SelfDestruct => self.self_destruct(),
+        }
+    }
+
+    #[cfg(feature = "jumptable")]
+    fn jumptable_placeholder(_self: &mut Self) -> OpResult {
+        panic!("jumptable placeholder was called. This should never happen. Either the jump analysis or the jump table itself is broken.");
```
LuisPH3, line 558: nop.
Please return InvalidInstruction,

LuisPH3 (changes requested):

PR #877, opened 2024-10-16
CT: Add CodeHash for CT runs

LuisPH3:
This was discovered by running coverage of lfvm Ct runs. CodeHash was never filled.

### DIFF
--- a/go/ct/utils/adapter.go
+++ b/go/ct/utils/adapter.go
@@ -26,8 +26,10 @@ import (
 func ToVmParameters(state *st.State) tosca.Parameters {
 
 	var code []byte
+	var codeHash tosca.Hash
 	if state.Code != nil {
 		code = state.Code.Copy()
+		codeHash = state.Code.Hash()
 	}
 
 	transactionContext := state.TransactionContext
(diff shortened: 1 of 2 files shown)


go/ct/utils/adapter.go
```diff
@@ -35,6 +36,8 @@ func ToVmParameters(state *st.State) tosca.Parameters {
 		transactionContext = &st.TransactionContext{}
 	}
 
+	codeHash := tosca.Hash(sha3.Sum256(code))
```
HerbertJordan, line 39: nit: the `state.Code` has a [Hash](https://github.com/Fantom-foundation/Tosca/blob/e20124038f600b26fcdfc91817ca2a2d31d0edc1/go/ct/st/code.go#L73) function that caches the result. Since the same code is used in some case thousands of times this would reduce the introduced runtime costs.

Also, technically the code hash is expected to be a `keccak256` hash, but I guess no interpreter is checking for that.

HerbertJordan (approved):
Thanks for fixing this.

simonlechner (approved):
LGTM

simonlechner (approved):

PR #878, opened 2024-10-16
Fix selfdestruct empty vs. exists bug

simonlechner:
The ethereum tests pointed out another problem regarding selfdestruct, the dynamic gas computation should not rely on whether the account exists but if its balance = 0, code = nil and nonce = 0.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -656,8 +656,10 @@ func opSelfdestruct(c *context) (status, error) {
 			cost += getAccessCost(accessStatus)
 		}
 	}
-	cost += selfDestructNewAccountCost(c.context.AccountExists(beneficiary),
-		c.context.GetBalance(c.params.Recipient))
+	empty := c.context.GetBalance(beneficiary).Cmp(tosca.Value{}) == 0 &&
+		c.context.GetCode(beneficiary) == nil &&
+		c.context.GetNonce(beneficiary) == 0
+	cost += selfDestructNewAccountCost(empty, c.context.GetBalance(c.params.Recipient))
 	// even death is not for free
 	if err := c.useGas(cost); err != nil {
 		return statusStopped, err
@@ -668,8 +670,8 @@ func opSelfdestruct(c *context) (status, error) {
 	return statusSelfDestructed, nil
 }
 
-func selfDestructNewAccountCost(accountExists bool, balance tosca.Value) tosca.Gas {
-	if !accountExists && balance != (tosca.Value{}) {
+func selfDestructNewAccountCost(accountEmpty bool, balance tosca.Value) tosca.Gas {
+	if accountEmpty && balance != (tosca.Value{}) {
 		// cost of creating an account defined in eip-150 (see https://eips.ethereum.org/EIPS/eip-150)
 		// CreateBySelfdestructGas is used when the refunded account is one that does
 		// not exist. This logic is similar to call.
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -656,10 +656,16 @@ func opSelfdestruct(c *context) (status, error) {
 			cost += getAccessCost(accessStatus)
 		}
 	}
-	empty := c.context.GetBalance(beneficiary).Cmp(tosca.Value{}) == 0 &&
+
+	empty := c.context.GetBalance(beneficiary) == (tosca.Value{}) &&
 		c.context.GetCode(beneficiary) == nil &&
 		c.context.GetNonce(beneficiary) == 0
-	cost += selfDestructNewAccountCost(empty, c.context.GetBalance(c.params.Recipient))
+	balance := c.context.GetBalance(c.params.Recipient)
+	if empty && balance != (tosca.Value{}) {
+		// cost of creating an account defined in eip-150 (see https://eips.ethereum.org/EIPS/eip-150)
+		cost += 25_000
+	}
+
 	// even death is not for free
 	if err := c.useGas(cost); err != nil {
 		return statusStopped, err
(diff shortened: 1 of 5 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -932,6 +934,80 @@ func TestSelfDestruct_ExistingAccountToNewBeneficiary(t *testing.T) {
 	}
 }
 
+func TestSelfDestruct_AccountExistsButIsNotEmpty(t *testing.T) {
```
LuisPH3, line 937: We have been implementing a new style lately, and I would like to adapt the name of this test to it. 
Basically, test check properties, and properties are described with a sentence that contains a verb. 
Some tests define the properties in the name of the function, and some do the same in the cases of the table. 
The sentence here does have a verb, but it does not describe an action done by opSelfDestruct, instead describes the input: 
```suggestion
func TestSelfDestruct_ReturnsSelfDestructedAndCosumesGas(t *testing.T) {
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -932,6 +934,80 @@ func TestSelfDestruct_ExistingAccountToNewBeneficiary(t *testing.T) {
...
+	tests := map[string]struct {
+		beneficiaryNonce   uint64
+		beneficiaryBalance tosca.Value
+		beneficiarCode     []byte
+		remainingGas       tosca.Gas
+	}{
+		"empty": {
```
LuisPH3, line 944:
```suggestion
		"with empty recipient account": {
```

go/interpreter/lfvm/instructions_test.go
```diff
@@ -932,6 +934,80 @@ func TestSelfDestruct_ExistingAccountToNewBeneficiary(t *testing.T) {
...
+			0,
+			tosca.Value{},
+			nil,
+			tosca.Gas(0),
+		},
+		"nonEmptyNonce": {
```
LuisPH3, line 950: "with recipient account with nonce": {
```

File: go/interpreter/lfvm/instructions_test.go [hunk 4]
```diff
@@ -932,6 +934,80 @@ func TestSelfDestruct_ExistingAccountToNewBeneficiary(t *testing.T) {
...
+			1,
+			tosca.Value{},
+			nil,
+			tosca.Gas(25_000),
+		},
+		"nonEmptyBalance": {
```
LuisPH3, line 956: "with recipient account with non-empty balance": {

go/interpreter/lfvm/instructions_test.go
```diff
@@ -932,6 +934,80 @@ func TestSelfDestruct_ExistingAccountToNewBeneficiary(t *testing.T) {
...
+			0,
+			tosca.Value{1},
+			nil,
+			tosca.Gas(25_000),
+		},
+		"nonEmptyCode": {
```
LuisPH3, line 962: "with recipient account with non-empty code": {

go/interpreter/lfvm/instructions.go
```diff
@@ -668,8 +670,8 @@ func opSelfdestruct(c *context) (status, error) {
 	return statusSelfDestructed, nil
 }
 
-func selfDestructNewAccountCost(accountExists bool, balance tosca.Value) tosca.Gas {
-	if !accountExists && balance != (tosca.Value{}) {
+func selfDestructNewAccountCost(accountEmpty bool, balance tosca.Value) tosca.Gas {
+	if accountEmpty && balance != (tosca.Value{}) {
 		// cost of creating an account defined in eip-150 (see https://eips.ethereum.org/EIPS/eip-150)
 		// CreateBySelfdestructGas is used when the refunded account is one that does
 		// not exist. This logic is similar to call.
```
LuisPH3, line 677: This comment looks suspiciously long, does it still hold?

go/interpreter/lfvm/instructions.go
```diff
@@ -668,8 +670,8 @@ func opSelfdestruct(c *context) (status, error) {
 	return statusSelfDestructed, nil
 }
 
-func selfDestructNewAccountCost(accountExists bool, balance tosca.Value) tosca.Gas {
-	if !accountExists && balance != (tosca.Value{}) {
+func selfDestructNewAccountCost(accountEmpty bool, balance tosca.Value) tosca.Gas {
```
LuisPH3, line 673: Nit: now that we are cleaning this up. This function is just an if. We either inline it or we move the new balance&code&nonce inside.

LuisPH3:
> The ethereum tests pointed out another problem regarding selfdestruct, the dynamic gas computation should not rely on whether the account exists but if its balance = 0, code = nil and nonce = 0.

Can you reference some literature about this issue? Is this experimental or is it there a documented description about this behavior?

LuisPH3 (dismissed):

HerbertJordan (dismissed):

go/interpreter/lfvm/instructions_test.go
LuisPH3: And if using a table test, `ExpectedBehavior` can be the name in the map?

simonlechner:
> > The ethereum tests pointed out another problem regarding selfdestruct, the dynamic gas computation should not rely on whether the account exists but if its balance = 0, code = nil and nonce = 0.
> 
> Can you reference some literature about this issue? Is this experimental or is it there a documented description about this behavior?

Here is the documentation inside the [geth code](https://github.com/ethereum/go-ethereum/blob/978ca5fc5e890cbbe22e1bb49d8811ad56033ba1/core/vm/interface.go#L70) and the related [eip 161](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md)

go/interpreter/lfvm/instructions_test.go
LuisPH3: Test target is, the file?  the package?  the function? 
we tend to write something like: `TestMemory_ExpandMemory_DoesExpand`

simonlechner: @LuisPH3 I will try to adopt this style in the future, nice that there is a common naming scheme now.

@HerbertJordan I think it is quick to see whether the invocation is supposed to be successful as we check for `err != nil`.

LuisPH3: I see that this discussion has gone beyond the scope of htis PR, for the time being feel free to carry on, but let's try not to lose this discussion

go/interpreter/lfvm/instructions.go
simonlechner: I shortened it to just mention the eip.

go/interpreter/lfvm/instructions_test.go
LuisPH3: I took this discussion here: https://github.com/Fantom-foundation/Tosca/issues/883

go/interpreter/lfvm/instructions.go
simonlechner: I think someone pulled it out for testing but I inlined it again and updated the tests.

LuisPH3 (approved):
Thanks for test refactor. :+1:

PR #884, opened 2024-10-17
CT: refactor domain for offsets.

LuisPH3:
This PR changes the definition for offsets, adding the difference between Data and Memory:
- Memory is artificially capped, because the VM will always allocate the memory if there is enough gas and this has a hardware limit. 
- Data Offsets are not capped, this is because input, code, and return data are not resizable. They handle overflows via other mechanisms (zero padding for reads, errors for writes) and therefore the offset values can be much larger. 
- Additionally, Size has been unified because it is used indistinctly. Sometimes combined. 

Some other refactors for the CT:
- remove gas=0 on error, this is not something we need to check.
- reactivate Geth in TestCt_ExplicitCases (playground test).

- [ ]  rerun ct coverage

### DIFF
--- a/go/ct/evm_test.go
+++ b/go/ct/evm_test.go
@@ -22,13 +22,14 @@ import (
 	"github.com/Fantom-foundation/Tosca/go/ct/spc"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
 	"github.com/Fantom-foundation/Tosca/go/interpreter/evmzero"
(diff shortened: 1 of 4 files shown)


facuMH (approved):
LGTM

HerbertJordan (approved):
Nice improvement, thanks!

simonlechner (approved):

PR #886, opened 2024-10-17
LFVM: Refactor data copy routines.

LuisPH3:
This PR started tracking a Fixme inside the instructions_test, related to memory expansion errors during operations manipulating memory. 
Deleting the test exposed the untested code, and the repeating pattern was obvious. 

There are Read Only buffers (RO-buffer) data can be copied from and into vm memory:
- input
- code
- external account code 

They follow the same rules: 
- Every RO-buffer read charges 3*word count of the copy, plus an expansion of the destination memory, then copy is done.
- If the expansion succeeds, the read operation from the RO-buffer cannot fail (excess is padded). 

Note: Return buffer is slightly different because write must fit in, and therefore it can overflow.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -666,83 +666,46 @@ func TestTransientStorageOperations(t *testing.T) {
 	}
 }
 
-func TestExpansionCostOverflow(t *testing.T) {
-	memTestValues := []uint64{
-		maxMemoryExpansionSize,
-		maxMemoryExpansionSize + 1,
-		math.MaxUint64,
-	}
+func TestGenericDataCopy_ReturnsErrorOn(t *testing.T) {
+	testBuffer := make([]byte, 1024)
 
 	tests := map[string]struct {
-		op         func(*context) error
-		stackSize  int
-		memIndexes []int
-		setup      func(*tosca.MockRunContext)
+		offset        uint256.Int
+		size          uint256.Int
+		gas           tosca.Gas
+		expectedError error
 	}{
-		"mcopy": {
-			op:         opMcopy,
-			stackSize:  3,
-			memIndexes: []int{0, 1, 2},
-			setup:      func(runContext *tosca.MockRunContext) {},
+		"not enough gas": {
+			offset:        *uint256.NewInt(10),
+			size:          *uint256.NewInt(10),
+			gas:           1,
+			expectedError: errOutOfGas,
 		},
-		"calldatacopy": {
-			op:         opCallDataCopy,
-			stackSize:  3,
-			memIndexes: []int{0, 2},
-			setup:      func(runContext *tosca.MockRunContext) {},
-		},
-		"codecopy": {
-			op:         opCodeCopy,
-			stackSize:  3,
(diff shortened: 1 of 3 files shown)


go/interpreter/lfvm/instructions_test.go
```diff
@@ -666,83 +666,46 @@ func TestTransientStorageOperations(t *testing.T) {
...
+		"not enough gas": {
+			offset:        *uint256.NewInt(10),
+			size:          *uint256.NewInt(10),
+			gas:           1,
+			expectedError: errOutOfGas,
+		},
+
+		"expansionFailure": {
+			offset:        *uint256.NewInt(math.MaxUint64),
+			size:          *uint256.NewInt(10),
```
facuMH, line 687: size is 10 for both cases. doesn't need to be a test case parameter.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -666,83 +666,46 @@ func TestTransientStorageOperations(t *testing.T) {
...
-					}
-					ctxt.stack.stackPointer = test.stackSize
-					ctxt.stack.data[memIndex].Set(uint256.NewInt(memValue))
-					for i := range test.memIndexes {
-						if i != memIndex {
-							ctxt.stack.data[i].Set(uint256.NewInt(1))
-						}
-					}
+			ctxt := getEmptyContext()
+			ctxt.gas = test.gas
```
facuMH, line 697:
```suggestion
ctxt := context{gas: test.gas}
```
only gas and stack are used. nothing else is needed, so it can be changed for this single line.

go/interpreter/lfvm/instructions_test.go
```diff
@@ -666,83 +666,46 @@ func TestTransientStorageOperations(t *testing.T) {
...
-				runContext.EXPECT().GetCode(gomock.Any()).AnyTimes().Return([]byte{0x01, 0x02, 0x03, 0x04})
-			},
+		"not enough gas": {
+			offset:        *uint256.NewInt(10),
+			size:          *uint256.NewInt(10),
+			gas:           1,
+			expectedError: errOutOfGas,
+		},
+
+		"expansionFailure": {
```
facuMH, line 685:
```suggestion
		"expansion failure": {
```

facuMH (commented):
great unification !!! just a couple of comments on the tests

go/interpreter/lfvm/instructions_test.go
```diff
@@ -666,83 +666,46 @@ func TestTransientStorageOperations(t *testing.T) {
 	}
 }
 
-func TestExpansionCostOverflow(t *testing.T) {
-	memTestValues := []uint64{
-		maxMemoryExpansionSize,
-		maxMemoryExpansionSize + 1,
-		math.MaxUint64,
-	}
+func TestGenericDataCopy_ReturnsErrorOn(t *testing.T) {
```
HerbertJordan, line 669: Is there a positive test making sure that data is copied correctly if all conditions are satisfied?

HerbertJordan (approved):
Nice!

facuMH (approved):

PR #888, opened 2024-10-17
CT: Support constraining emptiness of accounts

HerbertJordan:
This PR updates the `Accounts` model in the CT to support the differentiation between empty and non-existing accounts. This new capability is used to refine the specification of the self-destruct rules which reveal issues in the LFVM implementation.

This PR covers the following changes (in suggested reviewing order):
- an updated representation of the chain `Accounts` in the CT state; the new model retains a map of accounts which can be used to differentiate between non-existing accounts (accounts not in the map) and empty accounts (accounts not in the map or with a zero-Account -- thus no balance and no code)
- added support for constraints `AccountIsEmpty` and `AccountIsNotEmpty`
- added support for resolving those constraints in the `AccountsGenerator`
- refactored the specification of the `SelfDestruct` rule in the CT specification to (a) reduce complexity and (b) consider the emptiness of the beneficiary account

### DIFF
--- a/go/ct/evm_test.go
+++ b/go/ct/evm_test.go
@@ -21,6 +21,7 @@ import (
 	. "github.com/Fantom-foundation/Tosca/go/ct/rlz"
 	"github.com/Fantom-foundation/Tosca/go/ct/spc"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
+	"github.com/Fantom-foundation/Tosca/go/interpreter/evmrs"
 	"github.com/Fantom-foundation/Tosca/go/interpreter/evmzero"
 	"github.com/Fantom-foundation/Tosca/go/interpreter/geth"
 	"github.com/Fantom-foundation/Tosca/go/interpreter/lfvm"
@@ -29,9 +30,10 @@ import (
 )
 
 var evms = map[string]ct.Evm{
-	"geth":    geth.NewConformanceTestingTarget(),
 	"lfvm":    lfvm.NewConformanceTestingTarget(),
+	"geth":    geth.NewConformanceTestingTarget(),
 	"evmzero": evmzero.NewConformanceTestingTarget(),
+	"evmrs":   evmrs.NewConformanceTestingTarget(),
 }
 
 func TestCt_ExplicitCases(t *testing.T) {
--- a/go/ct/evm_test.go
+++ b/go/ct/evm_test.go
@@ -21,7 +21,6 @@ import (
 	. "github.com/Fantom-foundation/Tosca/go/ct/rlz"
 	"github.com/Fantom-foundation/Tosca/go/ct/spc"
 	"github.com/Fantom-foundation/Tosca/go/ct/st"
-	"github.com/Fantom-foundation/Tosca/go/interpreter/evmrs"
 	"github.com/Fantom-foundation/Tosca/go/interpreter/evmzero"
 	"github.com/Fantom-foundation/Tosca/go/interpreter/geth"
 	"github.com/Fantom-foundation/Tosca/go/interpreter/lfvm"
@@ -33,7 +32,6 @@ var evms = map[string]ct.Evm{
 	"geth":    geth.NewConformanceTestingTarget(),
 	"lfvm":    lfvm.NewConformanceTestingTarget(),
 	"evmzero": evmzero.NewConformanceTestingTarget(),
-	"evmrs":   evmrs.NewConformanceTestingTarget(),
 }
 
 func TestCt_ExplicitCases(t *testing.T) {
--- a/go/ct/evm_test.go
+++ b/go/ct/evm_test.go
@@ -30,8 +30,8 @@ import (
 )
 
 var evms = map[string]ct.Evm{
-	"lfvm":    lfvm.NewConformanceTestingTarget(),
 	"geth":    geth.NewConformanceTestingTarget(),
+	"lfvm":    lfvm.NewConformanceTestingTarget(),
 	"evmzero": evmzero.NewConformanceTestingTarget(),
 	"evmrs":   evmrs.NewConformanceTestingTarget(),
 }
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -23,8 +23,8 @@ import (
 )
 
 type AccountsGenerator struct {
-	existingAccounts []existenceConstraint
-	warmCold         []warmColdConstraint
+	emptyAccounts []emptyConstraint
+	warmCold      []warmColdConstraint
 }
 
 func NewAccountGenerator() *AccountsGenerator {
@@ -33,8 +33,8 @@ func NewAccountGenerator() *AccountsGenerator {
 
 func (g *AccountsGenerator) Clone() *AccountsGenerator {
 	return &AccountsGenerator{
-		existingAccounts: slices.Clone(g.existingAccounts),
-		warmCold:         slices.Clone(g.warmCold),
+		emptyAccounts: slices.Clone(g.emptyAccounts),
+		warmCold:      slices.Clone(g.warmCold),
 	}
 }
 
@@ -42,21 +42,21 @@ func (g *AccountsGenerator) Restore(other *AccountsGenerator) {
 	if g == other {
 		return
 	}
-	g.existingAccounts = slices.Clone(other.existingAccounts)
+	g.emptyAccounts = slices.Clone(other.emptyAccounts)
 	g.warmCold = slices.Clone(other.warmCold)
 }
 
-func (g *AccountsGenerator) BindToAddressOfExistingAccount(address Variable) {
-	c := existenceConstraint{address, true}
-	if !slices.Contains(g.existingAccounts, c) {
-		g.existingAccounts = append(g.existingAccounts, c)
+func (g *AccountsGenerator) BindToAddressOfEmptyAccount(address Variable) {
+	c := emptyConstraint{address, true}
+	if !slices.Contains(g.emptyAccounts, c) {
+		g.emptyAccounts = append(g.emptyAccounts, c)
 	}
 }
 
-func (g *AccountsGenerator) BindToAddressOfNonExistingAccount(address Variable) {
-	c := existenceConstraint{address, false}
-	if !slices.Contains(g.existingAccounts, c) {
-		g.existingAccounts = append(g.existingAccounts, c)
+func (g *AccountsGenerator) BindToAddressOfNoneEmptyAccount(address Variable) {
+	c := emptyConstraint{address, false}
+	if !slices.Contains(g.emptyAccounts, c) {
+		g.emptyAccounts = append(g.emptyAccounts, c)
 	}
 }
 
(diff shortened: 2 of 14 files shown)


go/ct/spc/specification.go
```diff
@@ -2256,48 +2262,72 @@ func nonStaticSelfDestructRules(revision tosca.Revision, warm bool, destinationC
...
 			IsRevision(revision),
 			hasSelfDestructedCondition,
-			targetWarm,
+			beneficiaryExists,
+			beneficiaryWarm,
 		},
 		parameters: []Parameter{AddressParameter{}},
 		effect: func(s *st.State) {
-			selfDestructEffect(s, destinationColdCost, accountCreationFee, refundGas)
+			selfDestructEffect(s)
```
LuisPH3, line 2270: effect:  selfDestructEffect,

go/ct/spc/specification.go
```diff
@@ -2256,48 +2262,72 @@ func nonStaticSelfDestructRules(revision tosca.Revision, warm bool, destinationC
...
 	// transaction. The transfer of the current balance to the given account cannot fail. In particular,
 	// the destination account code (if any) is not executed, or, if the account does not exist, the
 	// balance is still added to the given address.
 
-	// account to send the current balance to
-	destinationAccount := s.Stack.Pop().Bytes20be()
-	currentAccount := s.CallContext.AccountAddress
-	CurrentBalance := s.Accounts.GetBalance(currentAccount)
+	beneficiaryAccount := s.Stack.Pop().Bytes20be()
+	originatorAccount := s.CallContext.AccountAddress
```
LuisPH3, line 2284: Is this the recipient? I am a little confused with naming. 
evm-codes uses the term  `current account`.

go/ct/spc/specification.go
```diff
@@ -2256,48 +2262,72 @@ func nonStaticSelfDestructRules(revision tosca.Revision, warm bool, destinationC
...
 	dynamicCost := tosca.Gas(0)
-	if !CurrentBalance.IsZero() && !s.Accounts.Exist(destinationAccount) {
-		dynamicCost += accountCreationFee
-	}
 
-	dynamicCost += destinationColdCost
+	// Add costs for transfering the remaining balance.
+	if !originatorBalance.IsZero() {
+		// If the target account does not exist, the account creation fee is added.
+		if !s.Accounts.Exists(beneficiaryAccount) {
```
LuisPH3, line 2292: evm codes says: "State: is the beneficiary account empty?"  +25k if yes
" If a positive balance is sent to an empty account, the dynamic gas is 25000" 

The way I understand it, all non-existing accounts are empty, but not otherwise.

go/ct/spc/specification.go
```diff
@@ -2256,48 +2262,72 @@ func nonStaticSelfDestructRules(revision tosca.Revision, warm bool, destinationC
...
+			dynamicCost += 25000
+		}
+
+		// Add warm-up costs if the beneficiary account is cold.
+		if s.Revision > tosca.R07_Istanbul && !s.Accounts.IsWarm(beneficiaryAccount) {
```
LuisPH3, line 2297: I know that this was written already like this, but I would prefer to have
```suggestion
		if s.Revision >= tosca.R08_Berlin && !s.Accounts.IsWarm(beneficiaryAccount) {
```
To facilitate recognition of the EIP-2929 semantics introduced in Berlin.

go/ct/spc/specification.go
LuisPH3: python reference: 
```python
  if (
        not is_account_alive(evm.env.state, beneficiary)
        and get_account(evm.env.state, evm.message.current_target).balance != 0
    ):
        gas_cost += GAS_SELF_DESTRUCT_NEW_ACCOUNT
```

where is_account_alive is
```python
   account = get_account_optional(state, address)
    if account is None:
        return False
    else:
        return not (
            account.nonce == Uint(0)
            and account.code == b""
            and account.balance == 0
        )
```

```

**Review by LuisPH3 — CHANGES_REQUESTED** [review]
I think the exist logic does not model the right semantics: should be isEmpty instead of exists

File: go/ct/rlz/conditions_test.go [hunk 5]
```diff
@@ -199,6 +199,58 @@ func TestCondition_CheckStorageConfiguration(t *testing.T) {
 	}
 }
 
+func TestCondition_CheckForExistingAccount(t *testing.T) {
+	state := st.NewState(st.NewCode([]byte{}))
+	state.Pc = 42
+	condition := AccountExists(Pc())
```
simonlechner, line 205: I am very surprised to see the program counter being used as an address.

go/ct/spc/specification.go
simonlechner, line 2292: I thought this was the bug found in the ethereum tests and should be `s.Accounts.Empty`. This is how the dynamic gas computation is performed in [geth](https://github.com/ethereum/go-ethereum/blob/3ff73d46b3592ca37175021b9aa3cb5187698070/core/vm/gas_table.go#L492).

simonlechner (commented):
Looks good, but should we not use `account.Empty` in the selfdestruct effect?

go/ct/spc/specification.go
LuisPH3: I really think this should be "empty" and not just "exists"

HerbertJordan: Right, thanks for checking. Also geth uses the `Empty` check [here](https://github.com/ethereum/go-ethereum/blob/3ff73d46b3592ca37175021b9aa3cb5187698070/core/vm/gas_table.go#L492).

go/ct/rlz/conditions_test.go
HerbertJordan: Anything CT expression that returns a `U256` is ok in this place. I copied this from the `IsWarmChecks` but you are right, semantically this odd; updated to consume a stack parameter.

facuMH (commented):
just out of curiosity, how do these changes affect runtime/test case generation of a non full mode CT run ? also the generator info full count?

go/ct/spc/specification.go
HerbertJordan: It is now checking for emptiness instead of existence.

go/ct/evm_test.go
```diff
@@ -32,6 +33,7 @@ var evms = map[string]ct.Evm{
 	"geth":    geth.NewConformanceTestingTarget(),
 	"lfvm":    lfvm.NewConformanceTestingTarget(),
 	"evmzero": evmzero.NewConformanceTestingTarget(),
+	"evmrs":   evmrs.NewConformanceTestingTarget(),
```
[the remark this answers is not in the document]
HerbertJordan: Removed

go/ct/gen/accounts.go
```diff
@@ -115,17 +154,35 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
...
+			}
+			nonEmptyAccounts[address] = struct{}{}
+		}
+	}
+	for address := range emptyAccounts {
+		accountsBuilder.SetBalance(address, NewU256(0))
+		accountsBuilder.SetCode(address, Bytes{})
+	}
+	for address := range nonEmptyAccounts {
+		accountsBuilder.SetBalance(address, NewU256(0))
```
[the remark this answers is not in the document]
HerbertJordan: For an account to be non-empty, having either the balance or the code not being zero is enough to satisfy the constraint. It does not really matter which property is non-zero.

I updated the code to randomly chose which property should be set to non-zero.

go/ct/gen/accounts.go
```diff
@@ -147,3 +206,15 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
 
 	return accountsBuilder.Build(), nil
 }
+
+type emptyConstraint struct {
+	address Variable
+	empty   bool
+}
```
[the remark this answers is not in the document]
HerbertJordan: I have lots of ideas on how to improve this ... but as you say, future work.

go/ct/gen/accounts_test.go
```diff
@@ -93,6 +96,151 @@ func TestAccountsGenerator_WarmColdConstraintsNoAssignment(t *testing.T) {
...
+	want := []string{
+		"empty($v1)",
+		"!empty($v2)",
+		"empty($v3)",
+		"!empty($v3)",
+	}
+
+	for _, w := range want {
+		if !strings.Contains(print, w) {
+			t.Errorf("Expected to find %v in %v", w, print)
```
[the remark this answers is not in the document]
HerbertJordan: There is actually the verb `%q` for this.

go/ct/gen/accounts_test.go
```diff
@@ -93,6 +96,151 @@ func TestAccountsGenerator_WarmColdConstraintsNoAssignment(t *testing.T) {
...
+		}
+	}
+}
+
+func TestAccountsGenerator_EmptinessConstraintsAreSatisfied(t *testing.T) {
+	v1 := Variable("v1")
+	v2 := Variable("v2")
+	assignment := Assignment{}
+	rnd := rand.New(0)
```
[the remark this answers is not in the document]
HerbertJordan: Removed.

go/ct/gen/accounts_test.go
```diff
@@ -93,6 +96,151 @@ func TestAccountsGenerator_WarmColdConstraintsNoAssignment(t *testing.T) {
...
+
+	generator.BindToAddressOfEmptyAccount(v1)
+	generator.BindToAddressOfNoneEmptyAccount(v2)
+
+	assignment := Assignment{}
+	assignment[v1] = NewU256(42)
+	assignment[v2] = NewU256(42)
+
+	_, err := generator.Generate(assignment, rnd, tosca.Address{})
+	if !errors.Is(err, ErrUnsatisfiable) {
```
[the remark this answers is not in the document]
HerbertJordan: The error is only created due to the combination of of the following constraints:
```
empty(%v1) && !empty(%v2) && %v1=42 && %v2=42
```
So everything that is in this test is needed.

go/ct/st/accounts.go
```diff
@@ -11,212 +11,179 @@
...
-	return !a.IsWarm(key)
+// -- Warm / Cold Accounts --
+
+func (a *Accounts) IsWarm(key tosca.Address) bool {
+	_, contains := a.warm[key]
+	return contains
 }
 
-func (a *Accounts) MarkWarm(key tosca.Address) {
+func (a *Accounts) MarkWarm(address tosca.Address) {
```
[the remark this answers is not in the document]
HerbertJordan: In the effects of instructions we only ever need to update a cold account to a warm account. There is no instruction that would ever do the reverse.

go/ct/gen/accounts.go
```diff
@@ -32,17 +33,33 @@ func NewAccountGenerator() *AccountsGenerator {
...
+
+func (g *AccountsGenerator) BindToAddressOfNoneEmptyAccount(address Variable) {
+	c := emptyConstraint{address, false}
+	if !slices.Contains(g.emptyAccounts, c) {
+		g.emptyAccounts = append(g.emptyAccounts, c)
+	}
+}
+
 func (g *AccountsGenerator) BindWarm(address Variable) {
```
LuisPH3, line 63: To keep naming, BindToWarmAddress

go/ct/gen/state.go
```diff
@@ -240,6 +240,18 @@ func (g *StateGenerator) BindTransientStorageToZero(key Variable) {
...
 
+// BindToAddressOfEmptyAccount constraints the given variable to be mapped to
+// an address of an account that is empty.
+func (g *StateGenerator) BindToAddressOfEmptyAccount(address Variable) {
+	g.accountsGen.BindToAddressOfEmptyAccount(address)
+}
+
+// BindToAddressOfNoneEmptyAccount constraints the given variable to be mapped
+// to an address of an account that is not empty.
+func (g *StateGenerator) BindToAddressOfNoneEmptyAccount(address Variable) {
```
LuisPH3, line 251:
```suggestion
func (g *StateGenerator) BindToAddressOfNonEmptyAccount(address Variable) {
```

LuisPH3 (approved):

go/ct/gen/accounts.go
HerbertJordan: There is a long list of badly named functions here ... they should all be updated in one go. I think it is beyond the scope of this PR to fix those naming issues.

go/ct/gen/accounts_test.go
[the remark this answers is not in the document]
facuMH: ah, my bad, I thought it was testing isEmtpy for both.

facuMH (dismissed):

simonlechner (approved):

LuisPH3 (approved):

PR #894, opened 2024-10-18
CT: add support for balance constraints

HerbertJordan:
This PR introduces support for a new constraint expression: `Balance(X)`

The following additions are included (in recommended review order):
- an extension of the `AccountsGenerator` to constrain the balance of individual accounts (lower and upper bounds)
- an extension of the top-level `StateGenerator` to facilitate the definition of constraints on the `AccountAddress`, which is the address of the contract that is currently executed; it also enables variables to be bound to the account address
- new constraint expressions `Balance`, `Self` and `ToAddress` to compose constraints like `Eq(Balance(Self()), NewU256(32))` or `Gt(Balance(ToAddress(Param(0))), NewU256(40))`
- an update of the specification of the `Selfdestruct` ruleset to include conditions on the balance of the self-destructed contract; while doing so, a bug in the specification of the self-destruct effect was identified; cold-account-access costs on the beneficiary need to be always payed, even if no actual value transfer happens

### DIFF
--- a/go/ct/gen/state.go
+++ b/go/ct/gen/state.go
@@ -11,6 +11,7 @@
 package gen
 
 import (
+	"bytes"
 	"fmt"
 	"slices"
 	"sort"
@@ -325,6 +326,10 @@ func (g *StateGenerator) IsAbsentBlobHashIndex(variable Variable) {
 // generators are invoked automatically.
 func (g *StateGenerator) Generate(rnd *rand.Rand) (*st.State, error) {
 	assignment := Assignment{}
+	return g.generateWith(rnd, assignment)
+}
+
+func (g *StateGenerator) generateWith(rnd *rand.Rand, assignment Assignment) (*st.State, error) {
 
 	// Enforce variable assignments.
 	for _, binding := range g.variableBindings {
@@ -632,6 +637,18 @@ func (g *StateGenerator) String() string {
 	sort.Slice(g.pcVariableConstraints, func(i, j int) bool { return g.pcVariableConstraints[i] < g.pcVariableConstraints[j] })
 	for _, pc := range g.pcVariableConstraints {
 		parts = append(parts, fmt.Sprintf("pc=%v", pc))
+	}
+
+	sort.Slice(g.selfAddressBindings, func(i, j int) bool { return g.selfAddressBindings[i] < g.selfAddressBindings[j] })
+	for _, v := range g.selfAddressBindings {
+		parts = append(parts, fmt.Sprintf("selfAddress=%v", v))
+	}
+
+	sort.Slice(g.selfAddressConstraints, func(i, j int) bool {
+		return bytes.Compare(g.selfAddressConstraints[i][:], g.selfAddressConstraints[j][:]) < 0
+	})
+	for _, address := range g.selfAddressConstraints {
+		parts = append(parts, fmt.Sprintf("selfAddress=%v", address))
 	}
 
 	parts = append(parts, g.gasConstraints.Print("gas"))
--- a/go/ct/gen/state.go
+++ b/go/ct/gen/state.go
@@ -42,13 +42,15 @@ import (
 // The same applies to subsequent generators.
 type StateGenerator struct {
 	// Constraints
-	statusConstraints     []st.StatusCode
-	readOnlyConstraints   []bool
-	pcConstantConstraints []uint16
-	pcVariableConstraints []Variable
-	gasConstraints        *RangeSolver[tosca.Gas]
-	gasRefundConstraints  *RangeSolver[tosca.Gas]
-	variableBindings      []variableBinding
+	statusConstraints      []st.StatusCode
+	readOnlyConstraints    []bool
+	pcConstantConstraints  []uint16
+	pcVariableConstraints  []Variable
+	gasConstraints         *RangeSolver[tosca.Gas]
+	gasRefundConstraints   *RangeSolver[tosca.Gas]
+	variableBindings       []variableBinding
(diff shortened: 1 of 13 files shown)


go/ct/rlz/expressions.go
```diff
@@ -160,6 +160,42 @@ func (gas) String() string {
 	return "Gas"
 }
 
+////////////////////////////////////////////////////////////
+// Self - the address of the called contract
+
+type self struct{}
```
facuMH, line 166: I think just `self` is not enough to convey to which part of the whole `self` is this expression referring to, self-balance, self-address, self-nonce, etc.

go/ct/spc/specification.go
```diff
@@ -1659,15 +1659,18 @@ func getAllRules() []Rule {
...
-					rules = append(rules, makeSelfDestructRules(
-						revision,
-						hasSelfDestructed,
-						beneficiaryAccountEmpty,
-						beneficiaryAccountIsWarm,
-					)...)
+		for _, originatorHasFunds := range []bool{true, false} {
+			for _, beneficiaryAccountEmpty := range []bool{true, false} {
+				for _, beneficiaryAccountIsWarm := range []bool{true, false} {
+					for _, hasSelfDestructed := range []bool{true, false} {
```
LuisPH3, line 1665: Has self-destructed is a term I am not familiar with: I think we may have introduced it in our side, and may induce confusion. The way I understand existence/emptiness is that empty accounts will be pruned from the database when a transaction is committed (block processed) 
Therefore, any self-destructed account is empty for the duration of the block execution, although exists would return true.

Therefore, this test is checking twice for empty, but not for exists.

go/ct/gen/state.go
```diff
@@ -372,7 +402,52 @@ func (g *StateGenerator) Generate(rnd *rand.Rand) (*st.State, error) {
 		return nil, fmt.Errorf("failed to resolve gas refund constraints: %w", err)
 	}
 
-	accountAddress := RandomAddress(rnd)
+	// --- Self Address ---
```
facuMH, line 405: nit: this could be move into its own function to reduce the already large size of `Generate`

go/ct/rlz/expressions.go
LuisPH3: I agree, self is not a good name for any type.

go/ct/spc/specification.go
LuisPH3: I have been reading on this. Yes, self-destructed is neither exists nor empty. it is its own semantic.

go/ct/rlz/expressions.go
HerbertJordan: Please propose an alternative.

go/ct/rlz/domains.go
```diff
@@ -138,6 +138,35 @@ func (d u256Domain) SamplesForAll(as []U256) []U256 {
...
 
+////////////////////////////////////////////////////////////
+// Address
+
+type addressDomain struct{}
+
+func (addressDomain) Equal(a tosca.Address, b tosca.Address) bool { return a == b }
+func (addressDomain) Less(a tosca.Address, b tosca.Address) bool  { panic("not useful") }
+func (addressDomain) Predecessor(a tosca.Address) tosca.Address   { panic("not useful") }
+func (addressDomain) Successor(a tosca.Address) tosca.Address     { panic("not useful") }
```
simonlechner, line 149: `SamplesForAll` calls this function, so it should be implemented. It could just increment the address, this would also be a simple implementation for `Predecessor`.

go/ct/spc/specification.go
HerbertJordan: It is possible that due to nested calls the same account is self-destructed multiple times during a single transaction. Something like:
- A calls B
- B self-destructs, the account of B is scheduled to be removed at the end of the transaction, and control goes back to A
- A calls B again, and B self-destructs again

This is all Ok, and the interpreters would not care to inform the DB multiple times about the intent of B to self-destruct itself. The one exception, though, are gas refunds.

To incentivise contracts to clean up unused accounts, you get a gas refund in such cases. But: you should only get this refund once. Calling it multiple times should only give you the reward once. And thus, we need to track in the state whether it is the first time the self-destruct is called or a second or third time. The effect is [here](), either releasing a refund or not.

You do not see this in the python spec for Cancun, since this refund was removed in London. In the Berlin specs you can see it [here](https://github.com/ethereum/execution-specs/blob/3282ee5bb88bdd56aabf2461360277ab8eac4032/src/ethereum/berlin/vm/instructions/system.py#L493-L494).

go/ct/rlz/domains.go
HerbertJordan: Actually, `SamplesForAll` is not needed either. I removed its implementation.

I would not want to implement a predecessor and successor for addresses just for this now, where it is not needed.

go/ct/gen/state.go
HerbertJordan: I do not want to break the style for one function here now, mixing styles. This function should be refactored as a whole eventually.

go/ct/gen/state.go
HerbertJordan, line 1: I forgot to add the self-address constraints to the print

LuisPH3 (dismissed):

go/ct/rlz/expressions.go
HerbertJordan: Sorry, I forgot to fix this, will add this fix and update PR after lunch.

LuisPH3 (approved):

facuMH (approved):

PR #895, opened 2024-10-18
LFVM: extCodeHash use isEmpty semantics instead of exists.

LuisPH3:
As @HerbertJordan pointed out, [python reference](https://github.com/ethereum/execution-specs/blob/3282ee5bb88bdd56aabf2461360277ab8eac4032/src/ethereum/cancun/vm/instructions/environment.py#L458-L487) does not call `exists`.

[evm.codes](https://www.evm.codes/) states: "0 if the account does not exist or has been destroyed." 
My interpretation of this sentence is: 
- If the account does not exist, any query to the state will yield the zero value (code, balance, and nonce). Exists would be False. 
- If the account has been destroyed, it may still have an entry in the state with zero values. Exists would be true, although contents are empty.

### DIFF
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -742,14 +742,23 @@ func opExtCodeSize(c *context) error {
 }
 
 func opExtCodeHash(c *context) error {
-	slot := c.stack.peek()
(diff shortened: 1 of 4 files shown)


LuisPH3:
- [] Depends on #888, CT may not pass until then.

PR #896, opened 2024-10-18
LFVM: genericCall precompiled contract recipient

facuMH:
While executing the geth tests, lfvm failed to process a precompiled contract. There we realized this special case for delegate call and call code were not handled properly. 
This PR:
- moves `isPrecompiledContract` to tosca pkg to avoid unnecesary dependeencies.
- adds check for precompiled contract in generciCall, and handled nested call recipient properly. 
- [ ] add a unit test for these cases.


facuMH:
this work will be done by another PR.

PR #897, opened 2024-10-18
evmrs: add feature "jumptable-tail-call"

LorenzSchueler:
This PR adds the feature "jumptable-tail-call". When enabled this enables the feature "jumptable" and additionally implements the function dispatch using tail calls.

The idea here is that instead of calling the function for an opcode, executing it, returning to the dispatch function and then calling the function of the next opcode we instead call the next opcode function from the first one.

In order to implement this each function has to advance the pc itself, because the (inlined) dispatcher function does no longer knows what the last opcode was.

### DIFF
--- a/rust/src/interpreter.rs
+++ b/rust/src/interpreter.rs
@@ -98,6 +98,7 @@ where
         }
     }
 
+    #[cfg(not(feature = "jumptable-tail-call"))]
     pub fn run(&mut self) -> Result<(), FailStatus> {
         loop {
             if self.exec_status != ExecStatus::Running {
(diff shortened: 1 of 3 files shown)


rust/src/interpreter.rs
```diff
@@ -120,17 +121,30 @@ where
...
             };
             self.run_op(op)?;
-
-            if !(Opcode::Push1 as u8..=Opcode::Push32 as u8).contains(&(op as u8))
-                && op != Opcode::Jump
-                && op != Opcode::JumpI
-            {
-                self.code_reader.next();
-            }
         }
```
LorenzSchueler, line 124: Currently it would be possible to fetch the current op again from the code reader and then call a function which holds this block. However that would add additional overhead and besides I always disliked this block. For readability I much prefer having it in the op impl itself.

Once I have the opcode to function pointer conversion feature ready, this would no longer be possible because at that point we don't even have the opcodes around anylonger.

LuisPH3 (approved):

PR #898, opened 2024-10-19
Replace all Exist checks with Empty checks in interpreters

HerbertJordan:
This PR replaces all Account-Exist checks in interpreters with Account-IsEmpty checks.

As was discovered by running the Ethereum tests and studying the new Python based EVM specification, interpreter implementations should have no dependency to the existence of an account. Their only concern should be the emptiness of an account in contexts where gas prices for the implicit creation of accounts need to be computed or the hash of an external account should be determined.

This PR introduces the following changes:
- fixes a bug in the geth reference-interpreter integration where an empty account was incorrectly identified by checking for its non-existence; while this is true for any finalized state on the chain (non-empty accounts must not exist), empty accounts may temporary exist during the execution of a contract
- updates the EVMC adapter to check for emptiness when being asked for the existence of an account. It appears, whenever an EVMC implementation asks for the existence of an account, it is actually asking for its emptiness
- update of numerous unit tests to match those changes

Remaining tasks:
- [x] run historic substate tests on these changes
- [x] run ethereum JSON tests on these changes
- [x] run CT testing on these changes
   - [x] geth
   - [x] lfvm
   - [x] evmzero

### DIFF
--- a/go/integration_test/interpreter/dynamic_gas.go
+++ b/go/integration_test/interpreter/dynamic_gas.go
@@ -413,7 +413,8 @@ func getOutOfDynamicGasTests(revision Revision) []*FailGasTest {
 		mock.EXPECT().IsSlotInAccessList(gomock.Any(), gomock.Any()).AnyTimes().Return(false, false)
 		mock.EXPECT().IsAddressInAccessList(gomock.Any()).AnyTimes().Return(false)
 		mock.EXPECT().AccessAccount(gomock.Any()).AnyTimes().Return(tosca.ColdAccess)
-		mock.EXPECT().AccountExists(gomock.Any()).AnyTimes().Return(false)
+		mock.EXPECT().GetNonce(gomock.Any()).AnyTimes()
+		mock.EXPECT().GetCodeSize(gomock.Any()).AnyTimes()
 		mock.EXPECT().AccessStorage(gomock.Any(), gomock.Any()).AnyTimes().Return(tosca.ColdAccess)
 		mock.EXPECT().GetBalance(gomock.Any()).AnyTimes().Return(tosca.Value{})
 		mock.EXPECT().HasSelfDestructed(gomock.Any()).AnyTimes().Return(true)
@@ -577,20 +578,20 @@ func gasDynamicCallCommon(revision Revision, useCallValue bool, addressCreationG
 
 	type callTest struct {
 		testName         string
-		addrExist        bool
+		addrEmpty        bool
 		addrInAccessList bool
 		callValue        *big.Int
 	}
 
 	tests := []callTest{
-		{"Address exist and in access list", true, true, big.NewInt(0)},
-		{"Address exist and not in access list", true, false, big.NewInt(0)},
+		{"Address not empty and in access list", false, true, big.NewInt(0)},
+		{"Address not empty and not in access list", false, false, big.NewInt(0)},
 	}
 
 	if useCallValue {
 		tests = append(tests, []callTest{
-			{"Call value is > 0", true, true, big.NewInt(1)},
-			{"Call value is > 0 & address not exist", false, false, big.NewInt(1)},
+			{"Call value is > 0", false, true, big.NewInt(1)},
+			//{"Call value is > 0 & address empty", true, false, big.NewInt(1)},
 		}...)
 	}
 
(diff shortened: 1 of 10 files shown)


go/integration_test/interpreter/dynamic_gas.go
```diff
@@ -839,7 +845,12 @@ func gasDynamicSelfDestruct(revision Revision) []*DynGasTest {
 				return tosca.Value{}
 			})
 
-			mock.EXPECT().AccountExists(targetAddress).AnyTimes().Return(!empty)
+			nonce := uint64(0)
+			if !empty {
+				nonce = 1
```
facuMH, line 850: same here, should it not set codeSize as well ?

go/interpreter/evmc/evmc_interpreter.go
```diff
@@ -166,7 +166,13 @@ type hostContext struct {
 }
 
 func (ctx *hostContext) AccountExists(addr evmc.Address) bool {
-	return ctx.context.AccountExists(tosca.Address(addr))
+	// Although the EVMC function name asks for the existence of an account,
+	// it is actually referring to the emptiness of an account. The concept
+	// of an existing or non-existing account is a DB concept that is not
+	// exposed to any interpreter implementation.
```
facuMH, line 172: is this confusion something we could fix in the names of the interface's functions as well ?

facuMH (approved):

go/interpreter/evmc/evmc_interpreter.go
```diff
@@ -166,7 +166,13 @@ type hostContext struct {
 }
 
 func (ctx *hostContext) AccountExists(addr evmc.Address) bool {
```
simonlechner, line 168: Aren't there any functions which use this function as `AccountExists` and not as `AccountEmpty`?

simonlechner (approved):

go/interpreter/lfvm/instructions.go
```diff
@@ -1131,3 +1135,11 @@ func opLog(c *context, n int) error {
 	})
 	return nil
 }
+
+// isEmpty is a utility function that checks if an account is empty. An account
+// is considered empty if it has no nonce, no balance, and no code.
+func isEmpty(c tosca.RunContext, addr tosca.Address) bool {
```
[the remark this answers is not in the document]
HerbertJordan: Right, but `c.context` is of type [`WorldState`](https://github.com/Fantom-foundation/Tosca/blob/94eb947e823931bddfae5a12a83b43787ca14b82/go/tosca/world_state.go#L21) which is an interface. Adding it there would require all implementations of that interface to re-implement the exact same function.

go/interpreter/evmc/evmc_interpreter.go
```diff
@@ -216,6 +222,9 @@ func (ctx *hostContext) GetCodeSize(addr evmc.Address) int {
 }
 
 func (ctx *hostContext) GetCodeHash(addr evmc.Address) evmc.Hash {
+	if !ctx.AccountExists(addr) { // < Note: in the EVMC, this actually checks for emptiness
```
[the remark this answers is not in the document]
HerbertJordan: We need to implement the interface defined by EVMC, and [there](https://github.com/ethereum/evmc/blob/5a66578af363b28698835aa45eb920f20ae79f56/bindings/go/evmc/host.go#L89) this function is named `AccountExists`.

But I can introduce implement an actual `isEmpty` helper function to clarify this.

go/integration_test/interpreter/dynamic_gas.go
```diff
@@ -577,38 +578,45 @@ func gasDynamicCallCommon(revision Revision, useCallValue bool, addressCreationG
...
+			}
+			mock.EXPECT().AccountExists(address).AnyTimes().Return(true)
 			mock.EXPECT().GetCodeHash(address).AnyTimes().Return(hash)
 			mock.EXPECT().GetCode(address).AnyTimes().Return(calledCode)
-			mock.EXPECT().AccountExists(address).AnyTimes().Return(exist)
+			mock.EXPECT().GetNonce(address).AnyTimes().Return(nonce)
+			mock.EXPECT().GetBalance(address).AnyTimes().Return(tosca.Value{})
+			mock.EXPECT().GetCodeSize(address).AnyTimes()
```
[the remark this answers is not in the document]
HerbertJordan: I am not sure what you mean with this. Can you clarify?

go/integration_test/interpreter/dynamic_gas.go
HerbertJordan: It is enough if one of the three properties (nonce, balance, code) is non-zero to make an account empty. And for simplicity, the code here just sets to nonce to 1 to make the account non-empty.

go/interpreter/evmc/evmc_interpreter.go
HerbertJordan: This interface is defined by an external dependency -- the [here](https://github.com/ethereum/evmc/blob/5a66578af363b28698835aa45eb920f20ae79f56/bindings/go/evmc/host.go#L89), in the EVMC library.

go/integration_test/interpreter/dynamic_gas.go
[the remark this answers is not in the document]
facuMH: `GetCodeSize` does not have any `.Return()` specified, should this not be set in the if of line 611 as well ?

go/integration_test/interpreter/dynamic_gas.go
facuMH: ahhh right.

go/interpreter/evmc/evmc_interpreter.go
HerbertJordan: As we learned lately, nowhere within an interpreter information about the existence of an account is needed. So nobody on the other side of the EVMC interface should need account existence information. 

Account existence information seems to be only relevant on the processor level.

facuMH (approved):

simonlechner (approved):

LuisPH3 (approved):

PR #903, opened 2024-10-21
CT/LFVM: call provided gas over int64

facuMH:
This bug was first found through the Ethereum json tests.
When a `CALL` command is executed and the top of the stack is a value bigger than `maxInt64` but smaller than `maxUint64` this value was still considered accepted and overflowed when converted to uint64. 
This PR:
- expands CT `gasParameter` to include values over max int64
- add fix in the lfvm interpreter
- adds a test that fails without the fix in lfvm

### DIFF
--- a/go/ct/rlz/parameters.go
+++ b/go/ct/rlz/parameters.go
@@ -152,11 +152,8 @@ type GasParameter struct{}
 var gasParameterSamples = []U256{
 	NewU256(0),
 	NewU256(1),
-	NewU256(1 << 10),
-	NewU256(st.MaxGasUsedByCt),
 	NewU256(math.MaxInt64),
 	NewU256(math.MaxInt64 + 1),
-	NewU256(math.MaxUint64),
 }
 
 func (GasParameter) Samples() []U256 {
--- a/go/ct/rlz/parameters.go
+++ b/go/ct/rlz/parameters.go
@@ -154,6 +154,9 @@ var gasParameterSamples = []U256{
 	NewU256(1),
 	NewU256(1 << 10),
 	NewU256(st.MaxGasUsedByCt),
+	NewU256(math.MaxInt64),
+	NewU256(math.MaxInt64 + 1),
+	NewU256(math.MaxUint64),
 }
 
 func (GasParameter) Samples() []U256 {
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -1543,7 +1543,7 @@ func TestGenericCall_CallKindPropagatesStaticMode(t *testing.T) {
 	}
 }
 
-func TestGeneralCall_ResultIsWrittenToStack(t *testing.T) {
+func TestGenericCall_ResultIsWrittenToStack(t *testing.T) {
 	zero := *uint256.NewInt(0)
 	one := *uint256.NewInt(1)
 	for _, success := range []bool{true, false} {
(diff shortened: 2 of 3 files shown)


go/ct/rlz/parameters.go
```diff
@@ -154,6 +154,9 @@ var gasParameterSamples = []U256{
 	NewU256(1),
 	NewU256(1 << 10),
 	NewU256(st.MaxGasUsedByCt),
+	NewU256(math.MaxInt64),
+	NewU256(math.MaxInt64 + 1),
+	NewU256(math.MaxUint64),
```
HerbertJordan, line 159: Did this change detect the issue fixed in this PR?

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1563,6 +1563,38 @@ func TestGeneralCall_ResultIsWrittenToStack(t *testing.T) {
...
+		"maxUint64":  *uint256.NewInt(math.MaxUint64),
+	}
+
+	for name, providedGas := range tests {
+		t.Run(name, func(t *testing.T) {
+			nestedGas := tosca.Gas(gas - gas/64)
+			providedGasU64 := providedGas.Uint64()
+			if providedGas.IsUint64() && providedGasU64 <= math.MaxInt64 && (nestedGas >= tosca.Gas(providedGasU64)) {
+				nestedGas = tosca.Gas(providedGasU64)
+			}
```
HerbertJordan, line 1582: This condition is never true and it can be removed.
```suggestion
```

HerbertJordan (changes requested):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -1563,6 +1563,38 @@ func TestGeneralCall_ResultIsWrittenToStack(t *testing.T) {
...
 }
 
+func TestGenericCall_HandlesBigProvidedGasValues(t *testing.T) {
+	zero := *uint256.NewInt(0)
+	gas := tosca.Gas(50_000) // value big enough to cover all gas costs
+	tests := map[string]uint256.Int{
+		"maxInt64-1": *uint256.NewInt(math.MaxInt64 - 1),
+		"maxInt64":   *uint256.NewInt(math.MaxInt64),
+		"maxInt64+1": *uint256.NewInt(math.MaxInt64 + 1),
+		"maxUint64":  *uint256.NewInt(math.MaxUint64),
```
simonlechner, line 1573: The names could be improved by adding whether the nestedGas is updated in the if check or not.

go/ct/rlz/parameters.go
facuMH: yes, I first added this to the PR, then it failed for lfvm, then I fixed it the lfvm interpreter and the CT calls worked again.

go/interpreter/lfvm/instructions_test.go
facuMH: I thought this names would be better in case of a failure then you would see the name of the test and the case that fails. Not sure naming based on the effect it should have would help readability that much. I will try and let you know.

go/ct/rlz/parameters.go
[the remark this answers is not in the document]
facuMH: it is a parameter type that is only used in the call rules. Which is where we  want to stress it, not sure we can avoid using them.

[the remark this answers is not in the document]
LuisPH3: The last value can be removed (maxuint64) since maxint+1 already checks numbers larger than boundary. 
- 1<<10 is an arbitrary value, could be removed
- MaxGasUsedByCt is no longer the max gas used by the CT, we need to reconsider this constant, where does it come from and what does it mean now

go/interpreter/lfvm/instructions_test.go
simonlechner: Since the condition is removed this comment is outdated.

go/ct/rlz/parameters.go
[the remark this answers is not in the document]
facuMH: discussed offline, resolved: delete `1<<10` and `st.MaxGasUsedByCt` since it is from another domain. I will leave however `NewU256(math.MaxInt64)` and `NewU256(math.MaxInt64 + 1)`.

LuisPH3 (approved):

HerbertJordan (approved):
Thanks!

simonlechner (approved):

PR #904, opened 2024-10-21
Geth adapter: Fix error conversion

simonlechner:
The execution of the ethereum tests surfaced a bug in the error conversion between geth and tosca. If the maximum call depth is reached, the call was not successful but all its gas is returned. The same behavior is expected in an nonce overflow. 
Further, errors exceeding the max init code size were not handled in the conversion.

### DIFF
--- a/go/geth_adapter/adapter.go
+++ b/go/geth_adapter/adapter.go
@@ -352,8 +352,11 @@ func decodeReadOnlyFromGas(depth int, readOnly bool, gas uint64) (bool, uint64) 
 
 func gethToVMErrors(err error, gas tosca.Gas) (tosca.CallResult, error) {
 	switch err {
-	case geth.ErrInsufficientBalance:
-		// In this case, the caller get its gas back.
+	case
+		geth.ErrInsufficientBalance,
+		geth.ErrDepth,
+		geth.ErrNonceUintOverflow:
+		// In these cases, the caller get its gas back.
 		// TODO: this seems to be a geth implementation quirk that got
 		// transferred into the LFVM implementation; this should be fixed.
 		return tosca.CallResult{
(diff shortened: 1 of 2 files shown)


go/geth_adapter/adapter.go
```diff
@@ -347,9 +350,9 @@ func gethToVMErrors(err error, gas tosca.Gas) (tosca.CallResult, error) {
 	case
 		geth.ErrOutOfGas,
 		geth.ErrCodeStoreOutOfGas,
-		geth.ErrDepth,
 		geth.ErrContractAddressCollision,
 		geth.ErrExecutionReverted,
+		geth.ErrMaxInitCodeSizeExceeded,
```
HerbertJordan, line 355: Is there a way to figure out for which of those we should consume all the gas and for which not?

HerbertJordan (approved):
Looks good, but it would be great if we could get some way to verify that our mapping between errors and gas-handling policy is correct.

go/geth_adapter/adapter.go
simonlechner: We could check the code coverage while running the ethereum tests, if all errors are touched and all tests pass, they are handled according to the ethereum specification.

facuMH: can that be unit tested after the ethereum tests are verified ?

facuMH (approved):

go/geth_adapter/adapter.go
```diff
@@ -336,8 +336,11 @@ func decodeReadOnlyFromGas(depth int, readOnly bool, gas uint64) (bool, uint64)
 
 func gethToVMErrors(err error, gas tosca.Gas) (tosca.CallResult, error) {
 	switch err {
-	case geth.ErrInsufficientBalance:
-		// In this case, the caller get its gas back.
+	case
+		geth.ErrInsufficientBalance,
+		geth.ErrDepth,
+		geth.ErrNonceUintOverflow:
```
LuisPH3, line 342: I had no idea this was possible. Very interesting.

go/geth_adapter/adapter.go
simonlechner: I will continue to look into this but I think this PR is ready to be merged.

LuisPH3 (approved):

PR #905, opened 2024-10-21
evmrs: make context a trait object

LorenzSchueler:
This PR changes Interpreter such that it is no longer generic over ExecutionContextTrait, but instead holds a trait object (dynamic dispatch). This is needed for an upcoming PR which stores function pointers in statics. However, statics can not be generic so this is needed. 
The potential overhead of dynamic dispatch should not matter, because ExecutionContextTrait provides the db callbacks, which are much more expensive. A quick bench run also showed no significant changes.


LuisPH3 (approved):
Hopefully compiler de-virtualizes this. Did you see any performance change?

LorenzSchueler:
No, I did not see any performance change.

PR #906, opened 2024-10-22
Adds a cleanup stage in Jenkinsfile

wsodsong:
Tosca PR may run on a dedicated PR testing agent. Each PR is built as a separated job and has its own workspace on the agent, which quickly fills the storage space. This PR adds a cleanup stage to reduce required disk space.


LuisPH3:
Hi @wsodsong, I would recommend using some cleanup-plugin from Jenkins: https://www.jenkins.io/doc/pipeline/steps/ws-cleanup/#workspace-cleanup-plugin
The reason being shifting responsibilities: make clean could be ill-written and not cleaning the environment, projects could have not well tested make clean mechanisms (specially projects like Aida, fetching different versions of Tosca, Carmen...).

PR #907, opened 2024-10-22
Calls* forward code address

facuMH:
Thanks to the Ethereum json tests we found that all call kinds are expected to forward `CodeAddress`, We did not interpret this in CT nor the LFVM, so they are both fixed to reflect this new understanding. 
- To reflect this in the CT representation of the state, `CodeAddress` is added to `PastCalls` of the `CallJournal`.
- For the specification, in the `callEffect` used for the different call kinds, `CodeAddress` is initialized to the second parameter for all call kinds. 
- For geth, in `createGethInterpreterContext` there is yet no concept of the stack value, so it has to be modified forwarded in `geth/ct.go` where each call forwards the corresponding parameters into the corresponding call. 
- A test is added in LFVM to ensure correct call parameter usage depending on call kind.

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -2597,20 +2597,18 @@ func callEffect(s *st.State, addrAccessCost tosca.Gas, op vm.OpCode) {
 
 	sender := s.CallContext.AccountAddress
 	recipient := target.Bytes20be()
-	codeAddress := tosca.Address{}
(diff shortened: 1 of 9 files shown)


LuisPH3 (approved):
The new test checks arguments passed to Call are filled.

HerbertJordan (approved):
Looks good, thanks.

PR #909, opened 2024-10-22
evmrs: refactor code analysis

LorenzSchueler:
- Instead of wrapping the inner type of CodeAnalysis in a Box/Rc/Arc, now the CodeAnalysis itself gets wrapped. This has the advantage that when in the next PR CodeAnalysis gets more fields, they don't have to be wrapped each individually.
- some renaming of jump_cache to code_analysis_cache
- rename compute_code_byte_types to analyze_code and make it a method of CodeAnalysis
- simplify match in opcode.rs and return length of data and not how much to increment pc


rust/src/types/opcode.rs
```diff
@@ -323,40 +323,12 @@ pub fn code_byte_type(code_byte: u8) -> (CodeByteType, usize) {
...
-        PUSH27 => (CodeByteType::Opcode, 28),
-        PUSH28 => (CodeByteType::Opcode, 29),
-        PUSH29 => (CodeByteType::Opcode, 30),
-        PUSH30 => (CodeByteType::Opcode, 31),
-        PUSH31 => (CodeByteType::Opcode, 32),
-        PUSH32 => (CodeByteType::Opcode, 33),
-        JUMPDEST => (CodeByteType::JumpDest, 1),
-        _ => (CodeByteType::DataOrInvalid, 1),
+        | REVERT | INVALID | SELFDESTRUCT => (CodeByteType::Opcode, 0),
+        PUSH1..=PUSH32 => (
```
LuisPH3, line 327: nice

LuisPH3 (approved):
What is the reason of jump analysis being changed to code analysis? 
are you planning other analyses?

LorenzSchueler:
Yes, the conversion to functions pointers will also happen there.

PR #912, opened 2024-10-23
CT: add missing rules for extcodehash

facuMH:
This PR addresses the following issues:
- Thanks to the new interpretation of "empty" vs "exists" a missing case emerged for extcodehash, where an account was tested to be cold or warm, but not "not empty". 

the rules introduced by the cross product were tested both against geth as well as the lfvm.

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -1402,6 +1402,26 @@ func getAllRules() []Rule {
 		},
 	})...)
 
+	// notEmpty
+	rules = append(rules, rulesFor(instruction{
+		op:        vm.EXTCODEHASH,
+		name:      "_not_empty",
+		staticGas: 2600, // highest possible cost.
+		pops:      1,
+		pushes:    1,
+		conditions: []Condition{
+			RevisionBounds(tosca.R09_Berlin, NewestSupportedRevision),
+			AccountIsNotEmpty(Param(0)),
+		},
+		parameters: []Parameter{
+			AddressParameter{},
+		},
+		effect: func(s *st.State) {
+			address := extCodeHashEffect(s)
+			s.Accounts.MarkWarm(address)
+		},
+	})...)
+
 	// --- CHAINID ---
 
 	rules = append(rules, rulesFor(instruction{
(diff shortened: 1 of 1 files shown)


go/ct/spc/specification.go
```diff
@@ -1652,10 +1672,36 @@ func getAllRules() []Rule {
...
+			AddressParameter{},
+			ValueParameter{},
+			MemoryOffsetParameter{},
+			SizeParameter{},
+			MemoryOffsetParameter{},
+			SizeParameter{},
+		},
+		Condition: And(
+			Eq(Status(), st.Running),
+			Eq(Op(Pc()), vm.CALL),
```
LuisPH3, line 1692: Because comment says `// --- CALL, CALLCODE, STATICCALL and DELEGATECALL ---` I would have expected this rule to be applied to the 4 calls. I can see that this a little of a problem because of semantics. 

The CT does not know that they are all using the same code, therefore we cannot assume that calling it for call makes it correct for the other 3 calls. 
How difficult would be to generate 4 rules from this code?

facuMH: this particular rules is only for `call` not all types of calls. This is one of the reasons why I decided to make it its own rule instead of trying to add it to `getRulesForAllCallTypes`. It is under `// --- CALL, CALLCODE, STATICCALL and DELEGATECALL ---` because it is a specialized rule of the more generic `getRulesForAllCallTypes` even if it applies to a single type of call

LuisPH3 (approved):

go/ct/spc/specification.go
```diff
@@ -1402,6 +1402,26 @@ func getAllRules() []Rule {
 		},
 	})...)
 
+	// notEmpty
+	rules = append(rules, rulesFor(instruction{
```
HerbertJordan, line 1406: It seems to me that the main intention of introducing this rule is to up code coverage, but it does not add any new specification coverage. This is against the idea of writing up rules.

It would be better if the rules for EXTCODEHASH would follow the pattern of creating a Cartesian product of influencing parameters, to get the full combinations of empty / warm / revision-ranges covered;

go/ct/spc/specification.go
```diff
@@ -1402,6 +1402,26 @@ func getAllRules() []Rule {
...
+		op:        vm.EXTCODEHASH,
+		name:      "_not_empty",
+		staticGas: 2600, // highest possible cost.
```
HerbertJordan, line 1409: This static price assumes a cold address, right? But the condition is not specifying this.

It is very unlikely that a random address is indeed warm, so tests may pass. But conceptually, I think this rule is contradicting other rules.

go/ct/spc/specification.go
```diff
@@ -1652,10 +1672,36 @@ func getAllRules() []Rule {
 		},
 	})...)
 
-	// --- CALL, STATICCALL and DELEGATECALL ---
+	// --- CALL, CALLCODE, STATICCALL and DELEGATECALL ---
 
 	rules = append(rules, getRulesForAllCallTypes()...)
 
+	rules = append(rules, Rule{
```
HerbertJordan, line 1679: This rule defines a test case and is not a specification rule. It should not be part of the specification but of a set of test examples. Rule of thumb: if you fix values like `Eq(Gas(), 1234)` it is likely not a general rule but a test case.

Currently, we have a few condition-derived tests in [`evm_test.go`](https://github.com/Fantom-foundation/Tosca/blob/7f15c9ec604d53f7448289dd41767c86650db4d5/go/ct/evm_test.go#L37). We could extend them or even add a rule-derived section. This would get those tests covered.

However, this will not solve the problem that these type of cases are not tested by CT runs. The only solution for this is to refine the specification conditions by partitioning the state space using Cartesian products of conditions.

HerbertJordan (changes requested):

go/ct/spc/specification.go
facuMH: cartesian product of relevant variables has been added

go/ct/spc/specification.go
facuMH: solved by adding cross product of warm/cold, non/empty and revisions.

go/ct/spc/specification.go
facuMH: discussed offline, we concluded it is best to remove this rule.

LuisPH3 (changes requested):

go/ct/spc/specification.go
```diff
@@ -1345,62 +1345,54 @@ func getAllRules() []Rule {
...
+					op:         vm.EXTCODEHASH,
+					name:       name,
+					staticGas:  staticGas,
+					pops:       1,
+					pushes:     1,
+					conditions: conditions,
+					parameters: []Parameter{
+						AddressParameter{},
+					},
+					effect: func(s *st.State) {
```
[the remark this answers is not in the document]
facuMH: done

LuisPH3 (approved):

simonlechner (approved):

PR #913, opened 2024-10-24
Geth adapter: Revision depending on geth rules

simonlechner:
The ethereum substate started to fail starting with the first Paris revision block. This was caused by the MergeNetSplitBlock not being set. In order to not run into this problem in the future, the revision selection inside the geth adapter is now leveraged to the geth's ChainConfig rules. 

Check compatibility
- [x] Unit tests
- [x] Ethereum tests
- [ ] Ethereum substate

### DIFF
--- a/go/geth_adapter/adapter.go
+++ b/go/geth_adapter/adapter.go
@@ -88,7 +88,7 @@ func (a *gethInterpreterAdapter) Run(contract *geth.Contract, input []byte, read
 	rules := a.evm.ChainConfig().Rules(a.evm.Context.BlockNumber, a.evm.Context.Random != nil, a.evm.Context.Time)
 	revision, err := convertRevision(rules)
 	if err != nil {
-		return nil, fmt.Errorf("unsupported revision: %v", err)
+		return nil, fmt.Errorf("unsupported revision: %w", err)
 	}
 	if adapterDebug {
 		fmt.Printf("Running revision %v\n", revision)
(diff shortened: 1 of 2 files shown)


facuMH (approved):

LuisPH3 (approved):

go/geth_adapter/adapter.go
```diff
@@ -85,7 +85,11 @@ func (a *gethInterpreterAdapter) Run(contract *geth.Contract, input []byte, read
 	a.evm.SetDepth(a.evm.GetDepth() + 1)
 	defer func() { a.evm.SetDepth(a.evm.GetDepth() - 1) }()
 
-	revision := convertRevision(a.evm.ChainConfig(), a.evm.Context.BlockNumber, a.evm.Context.Time)
+	rules := a.evm.ChainConfig().Rules(a.evm.Context.BlockNumber, a.evm.Context.Random != nil, a.evm.Context.Time)
+	revision, err := convertRevision(rules)
+	if err != nil {
+		return nil, fmt.Errorf("unsupported revision: %v", err)
```
facuMH, line 91: is this error covered by any test ?

simonlechner: There is a test which checks that the `convertRevision` function returns an error which is the important part. All the errors in the run function should be checked in unit tests but this will be a separate PR.

HerbertJordan (approved):
Looks good, just some minor comments.

facuMH (approved):

PR #915, opened 2024-10-25
LFVM: Fix flaky Blob unit test

simonlechner:
The stack output should only be checked in case there was no error. Same as #910.

### DIFF
--- a/go/interpreter/lfvm/instructions_test.go
+++ b/go/interpreter/lfvm/instructions_test.go
@@ -356,7 +356,7 @@ func TestBlobHash(t *testing.T) {
 	}
 }
 
-func TestBlobBaseFee(t *testing.T) {
+func TestBlobBaseFee_ReturnsErrorWhenCalledWithUnsupportedRevision(t *testing.T) {
 
 	blobBaseFeeValue := tosca.Value{1}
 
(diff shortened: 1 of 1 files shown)


LuisPH3 (approved):

simonlechner:
After a comment from @facuMH, both blob tests got refactored and overlapping parts where removed.

LuisPH3 (dismissed):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -282,15 +282,13 @@ func TestCreateChecksBalance(t *testing.T) {
 	}
 }
 
-func TestBlobHash(t *testing.T) {
-
+func TestBlobHash_PushesCorrectValueOnStack(t *testing.T) {
```
facuMH, line 285: gas and revision can be removed from the test case parameters. 
`stack.push(uint256.NewInt(0))` is also used in all cases so it can be done as part of the test instead of the test cases.

simonlechner: I removed gas and revision but not all cases push 0, the last case is pushing 1.

facuMH (approved):

go/interpreter/lfvm/instructions_test.go
```diff
@@ -337,77 +320,36 @@ func TestBlobHash(t *testing.T) {
...
-			want:     tosca.Value{},
+func TestBlobBaseFee_ReturnsErrorWhenCalledWithUnsupportedRevision(t *testing.T) {
+	ctxt := context{
+		params: tosca.Parameters{
+			Recipient: tosca.Address{1},
 		},
+		stack:  NewStack(),
+		memory: NewMemory(),
 	}
+	ctxt.gas = 2
```
HerbertJordan, line 347: Likely all those parameters except the Revision are not relevant for this test.

HerbertJordan (approved):

go/interpreter/lfvm/instructions_test.go
simonlechner: Yes, I was able to remove several parameters.

LuisPH3 (approved):

PR #921, opened 2024-10-28
evmrs: optimize u256 operations

LorenzSchueler:
This PR optimizes some operations on u256. This improves performance by 10%.

In particular `bytes.iter().take(24).any(|b| *b > 0)` is way slower than a comparison with `&[0; 24]`.


LuisPH3 (changes requested):
This PR could have used a sentence explaining what was the problem fixed.
Please address the function naming.

LuisPH3 (approved):

PR #923, opened 2024-10-28
Fuzzing VMs: Limit memory usage

LuisPH3:
This PR adds a limit to the memory expansions generated by fuzzing tests. Nightly fuzzing has crashed a few times and the problem is attributed to large memory allocations. 

The CT has a [limit](https://github.com/Fantom-foundation/Tosca/blob/7b33705a4675b39f52859e2aadcbb9b69d3e9805/go/ct/st/state.go#L44) to the available gas for the execution of an instruction, large enough for a 450MB allocation. This limit may be too high for `FuzzDifferentialLfvmVsGeth`. During fuzzing, two resulting states are created by executing both geth and lfvm. This doubles the memory expansion.

The offending corpus: 
```
go test fuzz v1
[]byte("\xa0") 
int64(499_999_999_947)   
byte('\x04')
[]byte("\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x1e\x00\xe90\x00")
```
Arguments are:
-  op: LOG0
- gas: almost maximum allowed
- revision
- stack values: Last value is parameter size: 503_376_128 (480 MB)

The new code will not allow this execution by limiting gas to a 1/10 of the current limit.

### DIFF
--- a/go/ct/evm_fuzz_test.go
+++ b/go/ct/evm_fuzz_test.go
@@ -138,9 +138,9 @@ func fuzzVm(testee ct.Evm, f *testing.F) {
 }
 
 const (
-	fuzzIdealStackSize     int = 7             // < max pops in a single instruction
-	fuzzMaximumCodeSegment int = 33            // < 1 instruction with 32 data bytes
-	fuzzMaxGas             int = 5_000_000_000 // < gas limits memory usage
+	fuzzIdealStackSize     = 7             // < max pops in a single instruction
+	fuzzMaximumCodeSegment = 33            // < 1 instruction with 32 data bytes
+	fuzzMaxGas             = 5_000_000_000 // < gas limits memory usage
 )
 
 // prepareFuzzingSeeds is a helper function to be used by similar fuzzing tests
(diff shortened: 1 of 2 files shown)


go/ct/evm_fuzz_test.go
```diff
@@ -138,8 +138,9 @@ func fuzzVm(testee ct.Evm, f *testing.F) {
 }
 
 const (
-	fuzzIdealStackSize     int = 7  // < max pops in a single instruction
-	fuzzMaximumCodeSegment int = 33 // < 1 instruction with 32 data bytes
+	fuzzIdealStackSize     int = 7             // < max pops in a single instruction
+	fuzzMaximumCodeSegment int = 33            // < 1 instruction with 32 data bytes
+	fuzzMaxGas             int = 5_000_000_000 // < gas limits memory usage
```
facuMH, line 143: the 2 uses for this are `int64`, why not declare it as such ?

simonlechner (approved):

simonlechner (approved):

facuMH (approved):

PR #925, opened 2024-10-29
evmrs: add feature opcode-fn-ptr-conversion

LorenzSchueler:
This PR translates each opcode to a struct holding an optional function pointer and an u256 as data. This translation is caches as part of the code analysis. This improves performance by 20%.

### DIFF
--- a/rust/src/types/op_fn_data.rs
+++ b/rust/src/types/op_fn_data.rs
@@ -1,0 +1,72 @@
+use std::fmt::Debug;
+
+use crate::{
+    interpreter::{Interpreter, OpFn},
+    types::CodeByteType,
+    u256,
+};
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct OpFnData {
+    func: Option<OpFn>,
+    data: u256,
+}
+
+impl OpFnData {
+    pub fn data(data: u256) -> Self {
+        OpFnData { func: None, data }
+    }
+
+    pub fn skip_no_ops_iter(count: usize) -> impl Iterator<Item = Self> {
+        std::iter::once(OpFnData {
+            func: Some(Interpreter::SKIP_NO_OPS_FN),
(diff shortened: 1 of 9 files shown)


rust/src/types/op_fn_data.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::{
+    interpreter::{Interpreter, OpFn},
+    types::CodeByteType,
+    u256,
+};
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct OpFnData {
+    func: Option<OpFn>,
+    data: u256,
```
LuisPH3, line 12: Idea: because only push32 will use all this data, we can assume that most of the time it will be dead weight.
At the same time, there will be 32 noops after such instruction to preserve PC stability. 

can we use here a "much" smaller type and use multiple instructions to encode larger data? 

How big is the pointer size?  64 bits?  I hear you mentioning 32bits sometime before. Why is this happening?

LorenzSchueler: Currently OpFnData has size 40 and alignment 8.
Before I also tried it with the data field being `[u8; 8]`, which means larger push operations need multiple of those to store all the associated data. In this case OpFnData had size 16 and alignment 8. However that did not give any speedup while the current version gives 20%.

LorenzSchueler: I also tried to have a usize. Because pointers only need 6 bytes the upper byte can be used to idicate if it is a normal fn ptr, a jumpdest fn ptr or only data. In case of data the lower 4 bytes would be used for that. However that is also not faster.

LorenzSchueler: I will probably also create PRs for those versions later on because it is interesting for my thesis...

LuisPH3 (approved):

Issue #694, opened 2024-08-26
Develop stress tests for the LFVM interpreter


HerbertJordan:
In order to understand LFVM's behavior under high-load situations stress tests should be developed and carried out. In particular, the following scenarios are of interest:

- [x] a high number of concurrent EVM executions for small transactions (10.000 parallel runs)
- [x] a high number of (parallel) calls targeting distinct codes (ineffective cashing of codes)
- [x] a high number of (parallel) calls with large codes
- [ ] a high number of parallel calls stressing the hash cache

simonlechner:
Solved with #926.

PR #926, opened 2024-10-30
Add interpreter stress tests

simonlechner:
The lfvm is using multiple caches to improve performance. To ensure that these caches are not a bottleneck when a high number of concurrent calls are performed, stress tests are developed. 

Solves #694.

### DIFF
--- a/go/integration_test/interpreter/stress_test.go
+++ b/go/integration_test/interpreter/stress_test.go
@@ -1,0 +1,163 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package interpreter_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/Fantom-foundation/Tosca/go/interpreter/lfvm"
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
+	"golang.org/x/sync/errgroup"
+)
+
+func BenchmarkParallelStressTests(b *testing.B) {
+	codes := map[string][]byte{
+		"smallTransaction": getStressBenchmark_SmallTransaction(),
+		"largeCodes":       getStressBenchmark_LargeCodes(),
+		"stressHashCache":  getStressBenchmark_StressHashCache(),
+		"noCodeReUsage":    getStressBenchmark_NoCodeReUsage(),
+	}
+
+	for _, interpreterName := range getAllInterpreterVariantsForTests() {
+		interpreter, err := tosca.NewInterpreter(interpreterName)
+		if err != nil {
+			b.Fatalf("failed to load %s with error: %v", interpreterName, err)
+		}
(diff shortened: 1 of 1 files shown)


simonlechner:
LFVM stress tests
The LFVM shows the expected behavior for a very high number of calls. The time per call is not increasing more than linear in the number of calls once the caches reached their limit.

LuisPH3 (approved):

facuMH (approved):
nice test !!

go/integration_test/interpreter/stress_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"testing"
+	"time"
+
+	"github.com/Fantom-foundation/Tosca/go/interpreter/lfvm"
+	"github.com/Fantom-foundation/Tosca/go/tosca"
+	"github.com/Fantom-foundation/Tosca/go/tosca/vm"
+	"golang.org/x/sync/errgroup"
+)
+
+func BenchmarkParallelStressTests(b *testing.B) {
```
HerbertJordan, line 25: This code does not really use the features of a benchmark, and the reported results when running this as a benchmark using `go test -bench` are likely not well defined. Maybe instead of writing this like a benchmark it should actually be a test.
```suggestion
func TestParallel_StressTests(b *testing.T) {
```

Not, when defining a benchmark, the parameter `b.N` defines a number of times the code under test should be processed and it has build-in parallel execution support (see [doc](https://pkg.go.dev/testing#hdr-Benchmarks)). However, if the benchmark is not making use of these facilities, the reported results of the benchmark tool are not accurate.

go/integration_test/interpreter/stress_test.go
```diff
@@ -0,0 +1,163 @@
...
+					start := time.Now()
+					for i := 0; i < numCalls; i++ {
+						errs.Go(func() error {
+
+							if codeName == "noCodeReUsage" {
+								// code has to be different for each call
+								code = append(code, byte(vm.PUSH4))
+								code = append(code, uint32ToBytes(uint32(i))...)
+								code = append(code, byte(vm.STOP)) // stop
+								hash = lfvm.Keccak256(code)
```
HerbertJordan, line 54: There should be a race condition on this benchmark ... all parallel go-routines are using the same code and hash instance.

Also, is the intention that the code is gradually extended or should just the value to be pushed be replaced?

HerbertJordan (changes requested):
This PR is not making proper use of Go's benchmark infrastructure. This should either be fixed, or the the reference to Go's benchmark infrastructure should be removed. If the runtime is reasonable, it may be fine to have this as a regular test.

Also, I think this code may contain data races.

HerbertJordan:
nit: The figure shown above would need a label on the vertical axis.

go/integration_test/interpreter/stress_test.go
simonlechner: I originally implemented it as a unit test, that's why I am using go's time function to measure runtime, but due to its extremely long runtime when running it for all interpreters I switched it to a benchmark. I rewrote it to make better use of go's benchmark infrastructure.

go/integration_test/interpreter/stress_test.go
simonlechner: Yes, I didn't notice the race condition when I refactored the different tests into a single one, fixed now.
I planed to simply extend it, as it should not matter what the code does it just needs to be different.

simonlechner:
LFVM stress tests(1)
The updated time measurements using go's benchmark infrastructure.

HerbertJordan (approved):

PR #927, opened 2024-10-30
evmrs: test tail call elimination

LorenzSchueler:
This PR adds a test which makes sure tail calls are eliminated in release mode. This is necessary because otherwise, with sufficiently operations executed, the stack will overflow if feature tail-call is enabled.


LuisPH3 (commented):
I dont think this is the real cause of the issue. 
The crash happens because nested instances of the interpreter allocate the vm stack in the program stack. evmzero must have the same issue, how do they solve it? 

It would be interesting to address the issue directly instead of minimizing side-effects

LuisPH3 (approved):

PR #928, opened 2024-11-04
evmrs: safe transmute

LorenzSchueler:
This PR removes some unsafe code by relying on the crate [zerocopy](https://crates.io/crates/zerocopy).
The basic idea is to derive marker traits. In the implementation of the derive macro, the type gets analyzed at compile time. zerocopy then provides safe wrappers around `mem::transmute` for types which implement the necessary marker traits.


LuisPH3 (approved):
As I understand this change does not modify semantics, the unsafe casting blocks are now leveraged by the included crate. 
I think that I like it.

PR #932, opened 2024-11-05
evmrs: use weak linkage for profiling functions

LorenzSchueler:
This PR uses weak linkage for all coverage functions. This way linking will always success and the dump-cov feature is no longer needed. Furthermore, Go can now set the filename which makes it possible to remove the hard coded value from the Jenkins file. Because weak linkage is only possible in nightly rust, this is done in C functions which get called from Rust.


LuisPH3:
> This PR uses weak linkage for all coverage functions. This way linking will always success and the dump-cov feature is no longer needed. Furthermore, Go can now set the filename which makes it possible to remove the hard coded value from the Jenkins file. Because weak linkage is only possible in nightly rust, this is done in C functions which get called from Rust.

I need some more explanation about the weak linkage, and how it does affect the go code. can you link references or develop a little more how does this mechanism executes a dump?  describe a trace of events ?

LorenzSchueler:
The problem is that because evmrs and evmzero are shared libs and not executables, the profiling data is not dumped automatically. Therefore, Go has to call a function which does this. Both evmzero and evmrs expose such a function.
Before this PR evmrs had a feature dump-cov which if enabled exposed this function (which internally calls llvm injected functions). It was assumed that dump-cov gets only enabled if the shared lib was compiled with instrumentation. if this was not the case linking would fail.
With this PR rust always exposes this function. This Rust function then calls a C function which calls a the llvm function, which now is only linked weakly. If the function is available it is called otherwise it is basically a no op.

LuisPH3 (approved):

PR #935, opened 2024-11-11
evmrs: refactor

LorenzSchueler:
- only build lib in Makefile (and update BUILD.md accordingly)
- document how to use Perf + Firefox Profiler
- build evmrs without coverage instrumentation in run_ct.sh now that we have #932

### DIFF
--- a/rust/BUILD.md
+++ b/rust/BUILD.md
@@ -144,8 +144,8 @@ Also see [compare-features.sh](./scripts/compare-features.sh) which when provide
 
 ```sh
 cargo install --locked samply
-# same steps as for Perf + Flamegraph but instead of `perf script | ...` run 
-samply import perf.data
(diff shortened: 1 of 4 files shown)


rust/BUILD.md
```diff
@@ -142,7 +142,11 @@ Also see [compare-features.sh](./scripts/compare-features.sh) which when provide
 
 ### Perf + Firefox Profiler
 
-TODO
+```sh
+cargo install --locked samply
+# same steps as for Perf + Flamegraph but instead of `perf script | ...` run
```
LuisPH3, line 147: This is unfortunately not enough, I do not know what should I use instead of ...
It is not clear how to open the Firefox profiler, neither. (and in the next paragraph as well)

LuisPH3 (changes requested):
I need some help understanding the instructions

LuisPH3 (approved):

PR #936, opened 2024-11-12
CT: Add no code no effect rule

simonlechner:
A nightly CT full mode [run](https://scala.fantom.network/job/Tosca/job/Conformance%20Tests%20Geth/13/) failed with a state without no code. 
Geth is returning immediately if the code is empty while the Tosca interpreters/CT treat it like a STOP instruction. 
This PR adds the failing input as a regression and fixes the CT do treat it like geth. The code generation, lfvm and evmzero has not been updated yet.

### DIFF
--- a/go/ct/spc/specification.go
+++ b/go/ct/spc/specification.go
@@ -131,6 +131,18 @@ func getAllRules() []Rule {
 		}),
 	})
 
+	// --- No Code ---
+
+	rules = append(rules, Rule{
+		Name: "no_code_no_effect",
+		Condition: And(
+			AnyKnownRevision(),
+			Eq(Status(), st.Running),
+			Eq(CodeLength(), 0),
+		),
+		Effect: NoEffect(),
+	})
+
 	// --- Arithmetic ---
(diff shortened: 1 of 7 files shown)


go/ct/spc/specification.go
```diff
@@ -131,6 +131,18 @@ func getAllRules() []Rule {
 		}),
 	})
 
+	// --- No Code ---
+
+	rules = append(rules, Rule{
```
HerbertJordan, line 136: Adding this rule is likely introducing a contradiction between rules.

The way the problem this PR is fixing was discovered is through the CT. This means a state with the empty code was generated, passed through the specification, a rule was found, and this rule states that the effect should be that the execution stopped. By adding this rule you provide a contradicting definition to the rule that previously handled this case, which is invalid.

You should create a state with the empty code and ask the specification which rules are responsible for this state. Then this conflict must be resolved between the existing rules an this new rule.

HerbertJordan (changes requested):

PR #937, opened 2024-11-13
Lorenz/rust benches

LorenzSchueler:
This PR ports the go vm_test.go benchmarks to Rust. This hopefully makes profiling easier and the benchmark times more stable.
In addition, MockExecutionMessage was changed to hold static references. While it is safe to use non static references, convert them to pointers and pass them via FFI, this leads to problems then the data that the references refer to is on the stack, and therefore maybe no longer valid when used by the other side (Go).


rust/benchmarks/src/main.rs
```diff
@@ -1,9 +1,18 @@
...
+    let (mut args, expected) = RunArgs::ffi_overhead(1);
+    //let (mut args, expected) = RunArgs::inc(1);
+    //let (mut args, expected) = RunArgs::fib(20);
+    //let (mut args, expected) = RunArgs::sha3(1000);
+    //let (mut args, expected) = RunArgs::arithmetic(280);
+    //let (mut args, expected) = RunArgs::memory(10000);
+    //let (mut args, expected) = RunArgs::jumpdest_analysis(0);
+    //let (mut args, expected) = RunArgs::stop_analysis(0);
+    //let (mut args, expected) = RunArgs::push1_analysis(0);
+    //let (mut args, expected) = RunArgs::push32_analysis(0);
```
LuisPH3, line 13: dead ?

LuisPH3 (approved):

rust/benchmarks/src/main.rs
LorenzSchueler: No, just commented out to focus on a single benchmark. This will be adjusted as needed when profiling.

PR #938, opened 2024-11-13
evmrs: add logging interpreter

LorenzSchueler:
This PR adds a logging version of evmrs. It works similar to how it is done in evmzero. The run method is now generic over an Observer. If the NoOpOberver is used, the calls to pre_op and post_op are no ops and should be eliminated by the compiler. The LoggingProfiler prints the state of the interpreter.

### DIFF
--- a/rust/src/types/observer.rs
+++ b/rust/src/types/observer.rs
@@ -1,0 +1,52 @@
+use std::borrow::Cow;
+
+use crate::interpreter::Interpreter;
+
+pub trait Observer {
+    fn pre_op(&mut self, interpreter: &Interpreter);
+
+    fn post_op(&mut self, interpreter: &Interpreter);
+
+    fn log(&mut self, message: Cow<str>);
+}
+
+pub struct NoOpObserver();
+
+impl Observer for NoOpObserver {
+    fn pre_op(&mut self, _interpreter: &Interpreter) {}
+
+    fn post_op(&mut self, _interpreter: &Interpreter) {}
+
+    fn log(&mut self, _message: Cow<str>) {}
+}
+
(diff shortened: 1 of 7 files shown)


rust/src/types/observer.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    fn log(&mut self, _message: Cow<str>) {}
+}
+
+pub struct LoggingObserver();
+
+impl Observer for LoggingObserver {
+    fn pre_op(&mut self, interpreter: &Interpreter) {
+        println!(
+            "pre opcode={:?} gas={} stack-size={}",
```
LuisPH3, line 28: Is this compatible with the printed logs of lfvm? is the format equal?

rust/src/types/observer.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn post_op(&mut self, _interpreter: &Interpreter) {}
+
+    fn log(&mut self, _message: Cow<str>) {}
+}
+
+pub struct LoggingObserver();
+
+impl Observer for LoggingObserver {
+    fn pre_op(&mut self, interpreter: &Interpreter) {
+        println!(
```
LuisPH3, line 27: Please, use a file descriptor, which we can set to stdout or a file sync. 
These are the standard options implemented in lfvm.

LuisPH3 (changes requested):

rust/src/types/observer.rs
LorenzSchueler: it was not, but I adjusted it now.

LuisPH3 (approved):

PR #939, opened 2024-11-15
evmrs: add plot_benchstat.py script

LorenzSchueler:
This PR adds a python script to generate plots from benchstat outputs.
It reads the benchstat output from stdin and crates a plot for each benchmark and exports them as a png.

example:
Sha3_1000_

### DIFF
--- a/rust/scripts/plot_benchstat.py
+++ b/rust/scripts/plot_benchstat.py
@@ -1,0 +1,103 @@
+# USAGE: benchstat ... | python3 plot_benchstat.py
+
+import sys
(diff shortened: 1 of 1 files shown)


LuisPH3 (approved):

facuMH:
can you export them in another format ? like svg or pdf ? also if there are any, a list of python dependencies ? for documentation. otherwise, great work !!

LorenzSchueler:
I added documentation for both

facuMH (approved):

PR #944, opened 2024-11-22
evmrs: optimize u256 <-> U256 conversion

LorenzSchueler:
This PR optimizes the conversions u256 <-> U256.
The conversion u256 -> U256 before had a call (because for what ever reason U256::from_be_slice was not inlined) and a jump (for the unwrap). Both of those are now gone.
Additionally the conversions u256 <-> U256 are now also used for the conversions u256 <-> I256 and u256 <-> U512.

Unfortunately this does not improve performance, however I think it is still worth changing.


LuisPH3 (approved):
did this change have a performance impact?

PR #945, opened 2024-11-24
evmrs: fix performance regression introduced in #942

LorenzSchueler:
This PR fixes a performance regression introduced in #942. It mainly affected the static overhead and therefore all short benchmarks.

Instead of returning an error when the gas is negative, the gas is now just set to 0 and will then return out-of-gas error once the first opcode is executed. This has the advantage that the case for negative gas no longer needs to be handled explicitly.


LuisPH3 (commented):
this is rather surprising to me. I would have expected result type to be pretty good optimized.

LuisPH3 (approved):

PR #949, opened 2024-11-25
evmrs: consume Interpreter in run method and return Step- or ExecutionResult

LorenzSchueler:
Now that the evm stack is stored on the heap and the Interpreter type is small again, the api can be refactored somewhat.
This PR changes Interpreter::run such that it consumes the interpreter and returns either an ExecutionResult or a StepResult.

### DIFF
--- a/rust/src/interpreter.rs
+++ b/rust/src/interpreter.rs
@@ -2,7 +2,7 @@ use std::cmp::min;
 
 use evmc_vm::{
     AccessStatus, ExecutionMessage, ExecutionResult, MessageFlags, MessageKind, Revision,
-    StatusCode as EvmcStatusCode, StepResult, StorageStatus, Uint256,
+    StatusCode, StepResult, StorageStatus, Uint256,
 };
 
 use crate::{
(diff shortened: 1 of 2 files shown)


rust/src/interpreter.rs
```diff
@@ -408,29 +409,35 @@ impl<'a> Interpreter<'a> {
                     break;
                 }
                 Err(GetOpcodeError::Invalid) => {
-                    return Err(FailStatus::InvalidInstruction);
+                    return FailStatus::InvalidInstruction.into();
```
LuisPH3, line 412: I am not sure if this is more idiomatic than `Err(FailStatus::InvalidInstruction)` why do you prefer this?

rust/src/interpreter.rs
```diff
@@ -385,10 +385,11 @@ impl<'a> Interpreter<'a> {
     /// If the const generic S is false, the step check is skipped.
     /// If the const generic J is false jumpdests are skipped.
     #[cfg(not(feature = "tail-call"))]
-    pub fn run<O: Observer, const STEP_CHECK: bool, const JUMPDEST: bool>(
-        &mut self,
-        observer: &mut O,
-    ) -> Result<(), FailStatus> {
+    pub fn run<O, R, const STEP_CHECK: bool, const JUMPDEST: bool>(mut self, observer: &mut O) -> R
```
LuisPH3, line 388: I am not really sure if making the return type generic helps readability. After all the prototype of a function is the best documentation you can get. This prototype invites to get "creative"

rust/src/interpreter.rs
LorenzSchueler: Because the return type is no longer Result<(), FailStatus> but now either ExecutionResult or StepResult, Err(FailStatus::InvalidInstruction) is no longer valid. FailStatus::InvalidInstruction.into() converts the fail status into one of the two result structs.

rust/src/interpreter.rs
LorenzSchueler: We have to support returning ExecutionResult and StepResult so I see no way around the generic. Maybe a comment explaining this would help. I will add one.

LuisPH3 (approved):

PR #951, opened 2024-11-27
load evmc shared libs in factory

LorenzSchueler:


LorenzSchueler:
@LuisPH3 do you want to look into this or should I close the PR? I mean its nice to have but not really needed.

PR #953, opened 2024-12-03
CT: Skip states with PC on data

simonlechner:
As fixed in #918, states with the program counter pointing to data should be skipped. This PR adds a regression input and adds the skipping of states to the regression driver.

### DIFF
--- a/go/ct/driver/regressions.go
+++ b/go/ct/driver/regressions.go
@@ -117,6 +117,12 @@ func doRegressionTests(context *cli.Context) error {
 			expected := state.Clone()
 			rule.Effect.Apply(expected)
 
+			// TODO: do not only skip state but change 'pc_on_data_is_ignored' rule to anyEffect
+			// Pc on data is not supported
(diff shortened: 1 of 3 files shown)


go/ct/driver/regressions.go
```diff
@@ -117,6 +117,12 @@ func doRegressionTests(context *cli.Context) error {
 			expected := state.Clone()
 			rule.Effect.Apply(expected)
 
+			// TODO: do not only skip state but change 'pc_on_data_is_ignored' rule to anyEffect
```
HerbertJordan, line 120: Is there an issue number for this that we can add to this TODO?

HerbertJordan (approved):

LuisPH3 (dismissed):

go/ct/driver/regressions.go
simonlechner: I created an issue and added it to the comments.

LuisPH3 (approved):

PR #955, opened 2024-12-04
CT: Fix specification map

simonlechner:
Invalid instructions are only invalid for known revisions. This removes the need of having to check multiple sets of rules when selecting them are selected in the specification map.

### DIFF
--- a/go/ct/spc/specification_map.go
+++ b/go/ct/spc/specification_map.go
@@ -59,11 +59,10 @@ func (s *specificationMap) GetRulesFor(state *st.State) []Rule {
 	var opString string
 	if err != nil {
 		opString = "noOp"
+	} else if state.Revision == common.R99_UnknownNextRevision || state.Status != st.Running {
+		opString = "noOp"
 	} else {
 		opString = op.String()
-		if state.Revision == common.R99_UnknownNextRevision || state.Status != st.Running {
-			getRules("noOp")
-		}
 	}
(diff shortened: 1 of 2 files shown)


go/ct/spc/specification_map.go
```diff
@@ -59,11 +59,10 @@ func (s *specificationMap) GetRulesFor(state *st.State) []Rule {
...
+	} else if state.Revision == common.R99_UnknownNextRevision || state.Status != st.Running {
+		opString = "noOp"
 	} else {
 		opString = op.String()
-		if state.Revision == common.R99_UnknownNextRevision || state.Status != st.Running {
-			getRules("noOp")
-		}
 	}
 
 	getRules(opString)
```
HerbertJordan, line 68: Since `getRules` is now only used once, please also inline the code for this function here. It is very unusual to have a function like this having the "side-effect" of building up a result.

HerbertJordan (changes requested):
Is this change covered by unit tests already or should should tests be added?

go/ct/spc/specification_map.go
simonlechner: Right, I inlined the lambda function. 
There is the `TestSpecificationMap_SameRulesPerOperation` unit test which checks that for random states the map returns all applicable rules.

HerbertJordan (approved):

PR #957, opened 2024-12-04
evmrs: do not acquire ownership of interpreter if not needed

LorenzSchueler:
In this PR the ffi code is adjusted to no longer acquire ownership of the interpreter if not needed, but instead use a mutable reference. Before it was converted into a Box<_> and then the  box was release at the end.


LuisPH3 (approved):
I am not 100% what is the intended goal here, but the new code is much cleaner.

PR #965, opened 2024-12-09
evmrs: reduce unsafe code

LorenzSchueler:
This PR reduces the amount of unsafe code by:
- using unsafe extern blocks (see https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#safe-items-with-unsafe-extern). The idea here is that you only assert once (at the declaration site) that the extern functions have indeed the signature and afterwards it is then safe to call those functions.
- passing optional references instead of pointers
- consolidating the custom solution for generic statics over const bool in the trait GetGenericStatic

Note: these changes require Rust 1.82, so because we have to increase the MSRV, we might as well use 1.83.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -14,7 +14,7 @@ git submodule update --init --recursive
 - C/C++ toolchain, Clang >= 16 and Gcc >= 11.4
     - Ubuntu/Debian package: `clang`
     - Recommended: install `clang-format`, `clangd`, and `gdb` for development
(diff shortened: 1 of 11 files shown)


LuisPH3 (approved):

LorenzSchueler:
I had to revert some of the changes because they caused performance regressions. I'm still trying to figure out why, but I would like to merge the other changes anyway.

LuisPH3 (approved):

PR #968, opened 2024-12-11
evmrs: make dup, swap and push generic & inline return_from_op

LorenzSchueler:
return_from_op should always be inlined. Because the function is so short this is almost always the case, however while debugging something unrelated it turned out that this was not always the case.
Additionally Stack::swap_with_top now uses ptr::swap_nonoverlapping. This provides more guarantees to the compiler. The generated assembly is still the same and and not optimized better, however future versions now would have the chance to optimize it more.
Furthermore, dup, swap and push are now generic and no longer take any arguments.

Performance improvements/ regressions: None


LorenzSchueler:
This PR also needs a newer Rust version.

rust/src/interpreter.rs
```diff
@@ -175,22 +175,22 @@ const fn gen_jumptable<const STEPPABLE: bool>() -> [OpFn<STEPPABLE>; 256] {
...
-        |i| i.swap(8),
-        |i| i.swap(9),
-        |i| i.swap(10),
-        |i| i.swap(11),
-        |i| i.swap(12),
-        |i| i.swap(13),
-        |i| i.swap(14),
-        |i| i.swap(15),
-        |i| i.swap(16),
+        |i| i.swap(const { NonZeroUsize::new(1).unwrap() }),
```
LuisPH3, line 178: I still believe that a generic function will be less verbose and intuitive. 
this nonzero pattern is quite distracting.

LuisPH3 (commented):
Can you give it a try to see how bad is it with a generic ?

rust/src/interpreter.rs
LuisPH3: I understand the need for this type marker. But I do not like that it is pushed into the call to swap. 
I think we can declare it locally inside the swap function itself, to prevent polluting the code with unwraps.

LuisPH3 (changes requested):

rust/src/types/stack.rs
```diff
@@ -4,6 +4,12 @@ use std::sync::Mutex;
 
 use crate::types::{u256, FailStatus};
 
+struct NonZero<const N: usize>;
+
+impl<const N: usize> NonZero<N> {
+    const VALID: () = assert!(N > 0);
```
LuisPH3, line 10: is this the reason for the new toolchain support?

LuisPH3 (approved):
Have you noticed any change regarding inlining ?

rust/src/types/stack.rs
LorenzSchueler: No, the inline consts which have now been removed. But it is needed for #965 anyway.

LorenzSchueler:
No, because of inlining it seems to compile to pretty much the same assembly.

PR #970, opened 2024-12-16
evmrs: cross platform compilation

LorenzSchueler:
In order to make the c compilation work also on other platforms, this PR uses the cc crate which calls the platform c compiler instead of calling gcc directly.
It also uses a fixed up the version of evmc-vm (and evmc-sys) which now also builds on windows.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -52,8 +52,7 @@ fn-ptr-conversion-inline-dispatch = ["needs-fn-ptr-conversion"]
 [dependencies]
 bnum = "0.12.0"
 ethnum = "1.5.0"
-# evmc-vm-tosca = { package = "evmc-vm", git = "https://github.com/Fantom-foundation/evmc", branch = "tosca-extensions" }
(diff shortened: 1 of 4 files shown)


rust/Cargo.toml
```diff
@@ -52,7 +52,8 @@ fn-ptr-conversion-inline-dispatch = ["needs-fn-ptr-conversion"]
 [dependencies]
 bnum = "0.12.0"
 ethnum = "1.5.0"
-evmc-vm-tosca = { package = "evmc-vm", git = "https://github.com/Fantom-foundation/evmc", branch = "tosca-extensions" }
+# evmc-vm-tosca = { package = "evmc-vm", git = "https://github.com/Fantom-foundation/evmc", branch = "tosca-extensions" }
+evmc-vm-tosca = { package = "evmc-vm", git = "https://github.com/LorenzSchueler/evmc", branch = "lorenz/cross-platform" }                                # TODO
```
LuisPH3, line 56: fixme or todo would be nice to explain the commented branch line. otherwise: dead code

LorenzSchueler: That was only because the emvc PR was not merged yet. I will update the dependencies now.

PR #973, opened 2024-12-18
Floria: state contract fix nonce endian

simonlechner:
The nonce in the state contract has been set with the wrong endian.


LuisPH3 (approved):

facuMH (approved):
small fix, looks good. I'm wondering if this (or similar errors) can be caught in some way as to prevent similar endianess issues

PR #975, opened 2024-12-18
Floria: Fix access list

simonlechner:
For the access list introduced in the Berlin revision, the sender is always warm and the account which storage keys are warm is also marked as warm.


LuisPH3 (approved):
Transaction access list now include sender

facuMH (approved):
LGTM

